PR #5, opened 2015-08-27
Feature/edsc 714

mightynimble:


config/application.rb
```diff
@@ -92,6 +92,7 @@ def self.load_version
     config.services = YAML.load_file(Rails.root.join('config/services.yml'))
     config.gibs = JSON.parse(IO.read(Rails.root.join('config/gibs.json')))
     config.echo_env = 'ops'
+    config.cmr_client_id = ENV['cmr_client_id'] || config.services['cmr'][Rails.env.to_s]['cmr_client_id']
```
bilts: Could you change this and `config/services.yml` to just put the string in `config/environments/*.rb`? I think this is overly convoluted and non-standard.

I realize that this seems to directly contradict what I did I did earlier on line 97 and had you do. ECHO, URS, and CMR endpoints are set up differently because we wanted users to be able to dynamically point our SIT environment to different ECHO environments by setting a URL parameter. I can't think of a case where we'd want to dynamically set client id on a per-user basis, so it makes sense to go the easy/standard route and just use `config/environments/*.rb`

mightynimble: Sure. Will do

mightynimble:
Updated the pull request.

PR #6, opened 2015-08-27
Attempt to fix flaky delayed-job tests.

mightynimble:
Enqueue the job. Then manually work_off the job at a later time.


bilts:
I think this was a case where the 1s wait wasn't long enough. I have this fixed to use a more dynamic wait in #7

bilts:
... you win

mightynimble:
Don't say it too early, it is still building.

PR #13, opened 2015-09-14
EDSC-726 Add OpenSearch Descriptive Document link for dataset on data…

mightynimble:
…set details page.

### DIFF
--- a/app/presenters/dataset_details_presenter.rb
+++ b/app/presenters/dataset_details_presenter.rb
@@ -51,7 +51,7 @@ class DatasetDetailsPresenter < DetailsPresenter
     @dataset.dif_url = "#{metadata_url}.dif#{url_token}"
     @dataset.smap_iso_url = nil #"#{metadata_url}.smap_iso"
     opensearch_url = "#{Rails.configuration.services['earthdata'][env]['opensearch_root']}/granules/descriptor_document.xml"
-    @dataset.osdd_url = "#{opensearch_url}?utf8=%E2%9C%93&clientId=#{Rails.configuration.opensearch_client_id}&shortName=#{URI.escape(@dataset.short_name)}&versionId=#{URI.escape(@dataset.version_id)}&dataCenter=#{URI.escape(@dataset.archive_center)}&commit=Generate#{url_token}"
+    @dataset.osdd_url = "#{opensearch_url}?utf8=%E2%9C%93&clientId=#{Rails.configuration.client_id}&shortName=#{URI.escape(@dataset.short_name)}&versionId=#{@dataset.version_id}&dataCenter=#{URI.escape(@dataset.archive_center)}&commit=Generate#{url_token}"
 
     # Set description to URL if URLDescription doesn't exist
     @dataset.online_access_urls = [] if @dataset.online_access_urls.nil?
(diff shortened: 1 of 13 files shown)


config/services.yml
```diff
@@ -4,15 +4,17 @@ earthdata:
...
     urs_root: "https://uat.urs.earthdata.nasa.gov"
     echo_root: "https://api-test.echo.nasa.gov"
     cmr_root: "https://cmr.uat.earthdata.nasa.gov" # Does not exist yet
+    opensearch_root: "https://api-test.echo.nasa.gov/opensearch/"
   testbed:
     urs_root: "https://sit.urs.earthdata.nasa.gov"
     echo_root: "https://testbed.echo.nasa.gov"
-    cmr_root: "https://cmr.sit.earthdata.nasa.gov"
-    
+    cmr_root: "https://api.echo.nasa.gov/opensearch"
```
bilts: This is not correct

config/environments/uat.rb
```diff
@@ -82,4 +82,6 @@
   config.urs_client_id = services['urs'][Rails.env.to_s][services['earthdata'][config.echo_env]['urs_root']]
 
   config.cmr_client_id = ENV['cmr_client_id'] || 'edsc-prod'
+
+  config.opensearch_client_id = ENV['opensearch_client_id'] || 'edsc-prod'
```
bilts: Can you just reuse the CMR client id instead of configuring them separately? I can't think of a case where we'd want them to be different.

app/presenters/dataset_details_presenter.rb
```diff
@@ -50,6 +50,8 @@ def initialize(dataset, collection_id=nil, token=nil, env='ops')
     @dataset.iso19115_url = "#{metadata_url}.iso19115#{url_token}"
     @dataset.dif_url = "#{metadata_url}.dif#{url_token}"
     @dataset.smap_iso_url = nil #"#{metadata_url}.smap_iso"
+    opensearch_url = "#{Rails.configuration.services['earthdata'][env]['opensearch_root']}/granules/descriptor_document.xml"
+    @dataset.osdd_url = "#{opensearch_url}?utf8=%E2%9C%93&clientId=#{Rails.configuration.opensearch_client_id}&shortName=#{@dataset.short_name}&versionId=#{@dataset.version_id}&dataCenter=#{@dataset.archive_center}&commit=Generate#{url_token}"
```
bilts: Should URL escape the params.

config/services.yml
```diff
@@ -4,15 +4,17 @@ earthdata:
...
+    opensearch_root: "https://api.echo.nasa.gov/opensearch"
   partnertest:
     urs_root: "https://uat.urs.earthdata.nasa.gov"
     echo_root: "https://api-test.echo.nasa.gov"
     cmr_root: "https://cmr.uat.earthdata.nasa.gov" # Does not exist yet
+    opensearch_root: "https://api-test.echo.nasa.gov/opensearch/"
   testbed:
     urs_root: "https://sit.urs.earthdata.nasa.gov"
     echo_root: "https://testbed.echo.nasa.gov"
-    cmr_root: "https://cmr.sit.earthdata.nasa.gov"
```
bilts: I think this is still incorrect. You removed the cmr_root.

config/environments/development.rb
```diff
@@ -37,5 +37,6 @@
 
   config.url_limit = 200
 
-  config.cmr_client_id = ENV['cmr_client_id'] || 'edsc-dev'
+  # This is the client ID that Earthdata Search sends to other services, i.e. CMR and OpenSearch
+  config.client_id = ENV['client_id'] || 'edsc-dev'
```
bilts: Can we please change this back to just be cmr_client_id? The open search endpoint ultimately talks to the ECHO/CMR, which is why we specify it. I want to avoid confusion over the purpose of it and avoid having to change our operational environment variables twice as this change rolls out.

config/services.yml
mightynimble: Oh... I even double checked this line....

PR #15, opened 2015-09-15
Ignore vender/bundle to try to fix a 'failure' in travis.

mightynimble:


spec/features/dataset_facets/dataset_facets_spec.rb
```diff
@@ -329,6 +329,7 @@
 
     context "when selecting a term keyword" do
       before :all do
+        page.save_screenshot "1.png"
```
bilts: Please remove this

.gitignore
```diff
@@ -34,6 +34,7 @@ test_results
 /doc/ui/portfolio
 
 /vendor/cmep
+/vendor/bundle
```
bilts: Was this actually the problem?

mightynimble: Not sure. But the 'error' (in quotes, since it is not really an error - the task exited with 0 status) was gone after I ignored the files.

bilts: It looks like the build failed when you pushed this

mightynimble: It is failed due to another issue from a spec file. See commit 8061976.

bilts: My fault for putting the comment on the wrong side of the diff.

PR #18, opened 2015-09-16
EDSC-724 Hide deprecated order options.

mightynimble:


app/controllers/data_access_controller.rb
```diff
@@ -242,13 +242,19 @@ def get_order_access_methods(dataset_id, granules, hits)
...
+      else
+        config[:id] = option_id
+        config[:type] = 'order'
+        config[:form] = option_def['form']
+        config[:all] = config[:count] == granules.size
+        config[:count] = (hits.to_f * config[:count] / granules.size).round
+      end
       config
     end
+    defs = defs.select{|d| d != nil}
```
bilts: I think this is the same as `defs.compact`

fixtures/cassettes/orders_responses.yml
```diff
@@ -121,6 +121,68 @@ d0c2a1b2a0fd3078fcb7492bc2ea067cd82834e8:
       encoding: UTF-8
       string: " "
     http_version: 
+ae82d2a1dc8801714b10b5ddc4152532641556d8:
+  response:
+    status:
+      code: 404
```
bilts: These are 404 errors. Is that intentional?

mightynimble: They are not intentional but won't affect tests either. There are a couple of other existing appearances of 404 error with this order GUID (d0c2a1b2a0fd3078fcb7492bc2ea067cd82834e8 is one of them).

The tests are not affected because they don't assert on these responses.

app/controllers/data_access_controller.rb
mightynimble: Will switch to compact.

PR #20, opened 2015-09-21
EDSC-732 Introduce a browse scaler url for datasets to handle browse …

mightynimble:
…links in dataset metadata.

### DIFF
--- a/app/assets/javascripts/config.js.coffee.erb
+++ b/app/assets/javascripts/config.js.coffee.erb
@@ -37,7 +37,7 @@
     debug: true
     gibsSitUrl: 'https://sit.gibs.earthdata.nasa.gov/wmts-{lprojection}/wmts.cgi?{timeparam}&SERVICE=WMTS&REQUEST=GetTile&VERSION=1.0.0&LAYER={product}&STYLE=&TILEMATRIXSET={resolution}&TILEMATRIX={z}&TILEROW={y}&TILECOL={x}&FORMAT=image%2F{format}{extra}'
     gibsUrl: 'https://map1{s}.vis.earthdata.nasa.gov/wmts-{endpoint}/wmts.cgi?{timeparam}SERVICE=WMTS&REQUEST=GetTile&VERSION=1.0.0&LAYER={product}&STYLE=&TILEMATRIXSET={projection}_{resolution}&TILEMATRIX={z}&TILEROW={y}&TILECOL={x}&FORMAT=image%2F{format}{extra}'
-    browseScalerUrl: "https://api.echo.nasa.gov/browse-scaler/browse_images/granules"
+    browseScalerUrl: "https://api.echo.nasa.gov/browse-scaler/browse_images"
     present: -> new Date() - 0
     animateMap: true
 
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/config.js.coffee.erb
```diff
@@ -19,7 +19,8 @@
     csrfToken: csrfToken
     hoverTimeoutMs: 0
     gibsUrl: "/assets/gibs-example.jpeg?{timeparam}SERVICE=WMTS&REQUEST=GetTile&VERSION=1.0.0&LAYER={product}&STYLE=&TILEMATRIXSET={projection}_{resolution}&TILEMATRIX={z}&TILEROW={y}&TILECOL={x}&FORMAT=image%2F{format}{extra}"
-    browseScalerUrl: "/assets/gibs-example.jpeg#"
+    granuleBrowseScalerUrl: "/assets/gibs-example.jpeg#"
+    collectionBrowseScalerUrl: "/assets/gibs-example.jpeg#"
     present: -> Date.UTC(2014, 2, 1)
```
mightynimble: Patrick, do you think we should combine these two links into one?
And in coffeescripts we can say "#{scalerUrl}/datasets/C123-ABC?foo=bar" and "#{scalerUrl}/granules/G123-ABC?foo=bar"?

bilts: Yes

app/assets/javascripts/models/data/dataset.js.coffee
```diff
@@ -97,7 +97,13 @@ ns.Dataset = do (ko
 
     thumbnail: ->
       granule = @browseable_granule
-      "#{scalerUrl}/#{granule}?h=85&w=85" if granule?
+      collection_id = @browseable_collection
```
bilts: The way we have things now, collections show the browse image from their first granule. This logic won't work for that.

app/models/dataset_extra.rb
```diff
@@ -162,6 +162,7 @@ def decorate(dataset)
     decorate_modaps_layers(dataset)
 
     dataset[:links] = Array.wrap(dataset[:links]) # Ensure links attribute is present
+    dataset[:browseable_collection] = dataset[:id] unless dataset[:links].select{|link| link[:rel].include? 'browse'}.empty?
```
bilts: This shouldn't be on DatasetExtra. It can be easily gleaned from the metadata on the client side.

bilts: Style-wise, this line is a little confusing because it's a double-negative. I'd use positive logic:

    dataset[:browseable_collection] = dataset[:id] if dataset[:links].any? { |link| link[:rel].include?('browse') }

mightynimble: Moved this line out of rails and put it into coffeescript.

app/assets/javascripts/models/data/dataset.js.coffee
```diff
@@ -97,7 +97,13 @@ ns.Dataset = do (ko
 
     thumbnail: ->
       granule = @browseable_granule
-      "#{scalerUrl}/#{granule}?h=85&w=85" if granule?
+      collection_id = @id for link in @links when link['rel'].indexOf('browse#') > -1
+      if granule?
```
bilts: Can we flip this logic around and make the collection thumbnail take precedence?

mightynimble: Sure

PR #21, opened 2015-09-21
EDSC-698 Drop 'sensor' as a facet. Will fix failed specs in a later commit

mightynimble:


app/controllers/datasets_controller.rb
```diff
@@ -154,6 +154,9 @@ def facet_results(request, response)
 
     keyword_facets.sort_by! {|facet| facet['term']}
     results.insert(1, {name: "Keywords", param: nil, values: selected_keywords + keyword_facets})
+    # EDSC-698 Dropping sensor as a facet.
+    # FIXME: This may need retouch once CMR drops the sensor facet.
+    results.reject!{|facet| !facet.nil? && facet[:name] == 'Sensor'}
```
bilts: Do the compact before this and get rid of the `.nil?` check. This and the following line can be combined into just:

    results.compact.reject { |facet| facet[:name] == 'Sensor' }

spec/features/dataset_facets/dataset_facets_spec.rb
```diff
@@ -27,12 +27,6 @@
       find("h3.facet-title", text: 'Instrument').click
     end
 
-    it "shows the first Sensors facet" do
```
bilts: Since we have a requirement to hide sensor, I would add a new test that asserts that the sensor facet isn't shown.

PR #28, opened 2015-10-14
Feature/edsc 738

mightynimble:


lib/tasks/colormaps.rake
```diff
@@ -35,6 +35,19 @@ namespace :colormaps do
...
+  end
+
+  desc "Record the last run of task 'colormaps:load' by touching a file in ./tmp dir"
+  task :log_colormaps_load do
+    Dir.mkdir Rails.root.join('tmp') unless Dir.exist? Rails.root.join('tmp')
+    Dir.glob(Rails.root.join('tmp', "colormaps_load_*")).each do |f|
+      File.delete(f)
+    end
+
+    FileUtils.touch Rails.root.join('tmp', "colormaps_load_#{Time.now.to_i}")
```
bilts: Note that if you're using touch, you don't need to append the date, you can just use `File.mtime(Rails.root.join('tmp', 'colormaps_load')` to get the modification time. This could simplify logic.

lib/tasks/colormaps.rake
```diff
@@ -35,6 +35,19 @@ namespace :colormaps do
 
     exit 1 if error_count > 0
 
+    # explicitly invoke the task here to make sure task colormaps:load completes successfully.
+    Rake::Task['colormaps:log_colormaps_load'].invoke
```
bilts: Would it be useful to distinguish between the cases of running but failing and not running at all?

spec/controllers/health_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+      FileUtils.touch Rails.root.join('tmp', "colormaps_load_#{Time.now.to_i}")
+    end
+
+    after :each do
+      Dir.glob(Rails.root.join('tmp', "data_load_*")).each { |f| File.delete(f) }
+      Dir.glob(Rails.root.join('tmp', "colormaps_load_*")).each { |f| File.delete(f) }
+    end
+
+    context "when everything is up" do
+      it "should return a json response indicating the system is ok" do
```
bilts: Language for specs shouldn't use "should".  Just, "when everything is up, it returns a json response indicating the system is ok" (similar for specs below).

app/models/health.rb
```diff
@@ -0,0 +1,88 @@
...
+  end
+
+  private
+
+  def check_cron_job(job, interval)
+    Dir.glob(Rails.root.join('tmp', "#{job}_*")).each do |f|
+      last_run_time = Time.at(f.match(/[a-zA-Z_]+_(\d+)/)[1].to_i)
+      if last_run_time.nil? || last_run_time < Time.now - interval * 3
+        @ok = false
+        return {ok?: false, error: "Cron job '#{job.split('_').join(':')}' hasn't been run since #{last_run_time}."}
```
bilts: This error message is possibly misleading. If the cron job runs and fails, it will generate this error.

app/models/health.rb
```diff
@@ -0,0 +1,88 @@
...
+  end
+
+  def edsc_status
+    {ok?: @ok}
+  end
+
+  def delayed_job_status
+    job = DelayedJob.last
+    retrieval = Retrieval.last
+    if job.run_at > retrieval.created_at
```
bilts: I think this needs a grace period of a couple of minutes, or you run the risk of reporting a false positive the moment after a retrieval is queued, before the queue gets checked by the background job. Maybe select the first retrieval that's more than 5 minutes old?

app/models/health.rb
```diff
@@ -0,0 +1,98 @@
...
+    {ok?: @ok}
+  end
+
+  def delayed_job_status
+    # The job hasn't been run yet. Go on to check retrieval.created_at. If the time is more than 10 minutes (threshold)
+    # ago, consider the delayed job is down.
+    job = DelayedJob.last
+    retrieval = Retrieval.find_by_id(job.id)
+
+    if job.run_at.nil? && retrieval.created_at < 10.minutes.ago
```
bilts: First, I think you want retrieval.created_at > 10.minutes.ago.  But I think the way this whole condition is structured is going to cause problems. Consider that every time a new retrieval comes in, we could potentially falsely report that the jobs are running for the next 10 minutes. If we have a retrieval coming in every minute, we'll say we're processing them all, no matter what the state of the background job.

To do the grace period correctly, you need something like 
  
    DelayedJob.find('created_at > ? && run_at IS NULL', 10.minutes.ago)

If that comes back with a result, there is a stale job that's never been run.

bilts: Double check the way delayed job is configured. I believe by default it deletes old jobs, maybe even old failed jobs. Please make sure it's not doing that.

mightynimble: I am more inclined to your solution. But the condition needs slightly changed to avoid this false negative situation: if a retrieval is saved one minute ago and the delayed job on that one hasn't started yet, the query will return (at least) the latest entry. I haven't figured out how to make it better though. I will send out an update shortly.

mightynimble: I see that you are focusing on the status before 10 minutes ago whereas I am looking at what happened in the past 10 minutes.

We keep all failed jobs but DelayedJob always deletes the earlier successful ones.

PR #31, opened 2015-10-15
EDSC-754 WIP: a one-liner that triggers page load when no granules left in the list.

mightynimble:
Still in progress. "excludedGranulesList" gets cleared up after the new set of granules are loaded.


mightynimble:
There is a failed test in the pull request. Is there a specific reason that we need to assert the 'Undo' button is gone when granule filters are selected?

bilts:
I think it's because the undo button becomes ambiguous when you change the query and it gets less clear what's you're undoing.

mightynimble:
Will create a new branch off of the latest master since there are too many conflicts after dataset->collection renaming.

PR #42, opened 2015-10-28
Change /data/options AJAX call to use POST.

mightynimble:


app/assets/javascripts/models/data/project.js.coffee
```diff
@@ -79,6 +79,7 @@ ns.Project = do (ko,
         dataType: 'json'
         url: '/data/options'
         data: params
+#        method: 'post'
```
bilts: This is commented out, so unless I'm missing something, nothing has changed.

app/controllers/data_access_controller.rb
```diff
@@ -182,9 +182,9 @@ def options
         response.headers[key] = value if key.start_with?('cmr-')
       end
 
-      respond_with(result, status: catalog_response.status)
+      respond_with(result, status: catalog_response.status, location: nil)
```
bilts: Why did you have to do this? Why haven't we needed to do this with other tests?

mightynimble: It will otherwise throw an location related error because "result" is not an ActiveRecord. Can't remember the exact error though.

app/assets/javascripts/models/data/project.js.coffee
mightynimble: My bad. I must have commented out during the testing... I will correct it later today.

PR #49, opened 2015-11-05
EDSC-827 Edit condition to update browseable granule ID if it is older than a week.

mightynimble:


app/models/collection_extra.rb
```diff
@@ -26,7 +26,8 @@ def self.load
       extra.has_browseable_granules = false unless extra.has_granules
 
       if extra.has_granules
-        if extra.has_browseable_granules.nil? || (extra.has_browseable_granules && extra.browseable_granule.nil?)
+        # Edited the condition to update browseable_granule ID if it hasn't been updated for a week.
+        if extra.has_browseable_granules.nil? || (extra.has_browseable_granules && extra.updated_at < 1.week.ago)
```
bilts: If we are grabbing the oldest granule, this may break for NRT granules, which cycle out of existence. Can you make sure we're grabbing the newest granule?

spec/features/users/first_time_user_spec.rb
```diff
@@ -13,7 +13,7 @@
       visit '/'
     end
 
-    it 'saves the preferences and shows the landing page', intermittent: 1 do
+    it 'saves the preferences and shows the landing page', intermittent: 2 do
       expect(User.count).to eq(1)
```
bilts: Was it this line that failed? Are we not correctly clearing the User table somewhere?

mightynimble: Yes, the table is not cleaned up properly even though we have
```ruby
    ActiveRecord::Base.descendants.each do |model|
      model.destroy_all unless models_to_preserve.include?(model)
    end
```

bilts:
I approve this, but I'm waiting to merge so I can get in a fix to the zero downtime deployment that needs to get to UAT.

PR #53, opened 2015-11-09
[bugfix] Manual set isRelevant after sending user back to collecitons page.

mightynimble:


spec/features/granules/granule_filters_spec.rb
```diff
@@ -18,7 +19,7 @@
 
     before_granule_count = 0
     synchronize do
-      number_granules = expect(page.text).to match /\d+ Granules/
+      number_granules = expect(find('#project-collections-list').text).to match /\d+ Granules/
```
bilts: Use helpers defined in overlay_helpers.rb or define new ones as needed. I think `expect(project_overview).to ...` will do what you want here.

app/assets/javascripts/models/ui/state_manager.coffee
```diff
@@ -251,6 +251,8 @@
       if @_isValid(path, serialized)
         changed = urlUtil.saveState(path, serialized, !@historyChanged, @page.workspaceNameField.peek())
         @historyChanged = true if changed
+        if path == '/search/collection-details' || path == '/search/granules' || path == '/search/project'
```
bilts: I think this is the wrong logic. page.collections should not be relevant on any of these paths. It should only be relevant when the collections view is visible. It's also not necessarily relevant when transitioning *from* these paths. For example, from the project page, you can transition to collection details or granules, which should not make the collection search relevant.

app/assets/javascripts/models/ui/state_manager.coffee
```diff
@@ -247,10 +248,12 @@
         @_sendToCollectionList()
         path = @path()
         serialized = @serialize()
+        refreshColList = true
 
       if @_isValid(path, serialized)
         changed = urlUtil.saveState(path, serialized, !@historyChanged, @page.workspaceNameField.peek())
         @historyChanged = true if changed
+        @page.collections.isRelevant(refreshColList)
```
bilts: This should be:

    @page.collections.isRelevant(true) if refreshColList

It's possible that collections are relevant when you get to this point, and you don't want them to accidentally be set to false. Given that, I'm not actually sure how the specs all seem to be passing.

fixtures/cassettes/hand-edited_requests.yml
```diff
@@ -295,6 +295,19 @@ http_interactions:
   digest: invalid_but_load_page
 - request:
     method: get
+    uri: https://cmr.earthdata.nasa.gov/search/collections.json?include_facets=true&page_size=20&page_num=1&include_has_granules=true&include_granule_counts=true
```
bilts: How did this end up here?

mightynimble:
I saw in the test log that it tries to find this fixture in hand_edited in ```after :all```. So I put it there and the error is gone. Well, for most of the time, VCR just complains but the tests won't fail.

fixtures/cassettes/hand-edited_requests.yml
mightynimble: I saw in the test log that it tries to find this fixture in hand_edited in after :all. So I put it there and the error is gone. Well, for most of the time, VCR just complains but the tests won't fail.

bilts: Yes, but it's very strange, because this is pretty much a default collections search. Some logic must be broken.

PR #57, opened 2015-11-11
CWIC-specific OSDD endpoints

bilts:


app/assets/javascripts/models/data/collection.js.coffee
```diff
@@ -143,6 +155,15 @@ ns.Collection = do (ko
           success: (data) ->
             @featured = data
 
+    isExternal: ->
+      @openSearchEndpoint()?
+
+    openSearchEndpoint: ->
+      if @json.tags
+        for [k, v] in @json.tags
+          return openSearchKeyToEndpoint[v]?(this) if k == 'gov.nasa.earthdata.search.opensearch'
```
mightynimble: Should we move ```gov.nasa.earthdata.search.opensearch``` into config?

app/controllers/collections_controller.rb
```diff
@@ -287,6 +287,8 @@ def collection_params_for_request(request, hierarchical=true)
 
     params['hierarchical_facets'] = 'true' if params['include_facets'] == 'true' && hierarchical
 
+    params['include_tags'] = 'gov.nasa.earthdata.search.*'
```
mightynimble: Same question here. Should we get this value from the config?

PR #58, opened 2015-11-12
EDSC-831 and EDSC-863

mightynimble:
drop granules that don't have the specified access option and the dropped are listed on the order retrieval page.


bilts:
General comment: I'd like to tidy up our commits and pull requests a bit. All commits pertaining to a JIRA issue should start with the JIRA issue ID. I'd encourage squashing commits or rebasing in order to remove "Revert" or "WIP" commits. Ideally, all the commits in master would add valuable information about what a change did. When a change is made and then reverted before getting to master, it clouds things up.

You don't have to fix that here, but something to keep in mind moving forward.

lib/echo/echo_client.rb
```diff
@@ -59,7 +59,7 @@ def get_availability_events
     end
 
     def get_order_information(item_ids, token)
-      get('/echo-rest/order_information.json', {catalog_item_id: item_ids}, token_header(token))
+      post('/echo-rest/order_information.json', options_to_granule_query({catalog_item_id: item_ids}).to_query, token_header(token).merge('Content-Type' => 'application/x-www-form-urlencoded'))
```
bilts: Why are you calling options_to_granule_query? I don't think it alters anything, and this is not a granule query, it's an order information query.

mightynimble: should have called ```options_to_item_query```

app/views/data_access/retrieval.html.erb
```diff
@@ -78,7 +78,22 @@
                  data-method="post"
                  data-confirm="Are you sure you want to remove this order?  This action cannot be undone."
                  rel="nofollow">Cancel</a>
-              <!-- /ko -->
+            <!-- /ko -->
+            <!-- ko if: dropped_granules -->
+              <p>The following granules are dropped from this order because the data access method <strong data-bind="text: method_name"></strong> is not supported:</p>
```
bilts: Please change to:

    The following granules will not be processed because they do not support the <strong data-bind="text: method_name"></strong> access method:

Note that despite what line 79 says, we try not to use the word "order" anywhere user-visible.

app/views/data_access/retrieval.html.erb
```diff
@@ -78,7 +78,22 @@
...
+              <div class="granule-list">
+                <div>
+                  <ul>
+                  <!--ko foreach: dropped_granules-->
+                    <li>
+                      <h5 data-bind="text: producer_granule_id"></h5>
```
bilts: This isn't the correct logic for displaying a granule. See granule.coffee. It should be producer_granule_id if present, otherwise title.

lib/echo/echo_client.rb
```diff
@@ -133,7 +147,7 @@ def create_order(granule_query, option_id, option_name, option_model, user_id, t
       Rails.logger.info "User info response: #{user_info_response.body.inspect}"
       Rails.logger.info "Submission response: #{submission_response.body.inspect}"
 
-      {order_id: id, response: submission_response, count: granules.size}
+      {order_id: id, response: submission_response, count: granules.size, dropped_granules: dropped_granules}
```
bilts: dropped_granules contains all of the granule metadata, which could be a lot of information, way more than we need. Just store enough to display and re-query. id and name (where name is producer_granule_id or title)

lib/echo/echo_client.rb
```diff
@@ -94,6 +94,20 @@ def create_order(granule_query, option_id, option_name, option_model, user_id, t
...
       catalog_response = cmr_client.get_granules(granule_query, token)
       granules = catalog_response.body['feed']['entry']
+      order_info = get_order_information(granules.map {|g| g['id']}, token).body
+
+      # drop granules from the order if the selected order option is not one of the supported options by this granule
+      dropped_granules = []
+      Array.wrap(order_info).each do |info|
+        info = info['order_information']
+        unless Array.wrap(info['option_definition_refs']).any? {|ref| ref['name'] == option_name}
+          granule = granules.select {|g| g['id'] == info['catalog_item_ref']['id']}
```
bilts: This and line 106 make the algorithm O(n^2). 4 million iterations for an order size of 2000.

lib/echo/echo_client.rb
```diff
@@ -94,6 +94,20 @@ def create_order(granule_query, option_id, option_name, option_model, user_id, t
       #return {order_id: 1234, count: 2000}
       catalog_response = cmr_client.get_granules(granule_query, token)
       granules = catalog_response.body['feed']['entry']
+      order_info = get_order_information(granules.map {|g| g['id']}, token).body
+
+      # drop granules from the order if the selected order option is not one of the supported options by this granule
+      dropped_granules = []
+      Array.wrap(order_info).each do |info|
```
bilts: Suggestions for cleaning up and making it faster (You don't necessarily need to do it this way):
    
    # Given order info JSON references, returns a Hash of granule ids to their options
    def order_info_to_granule_options(order_info)
      order_info.index_by { |info| info['catalog_item_ref']['id'] }
        .transform_values { |info| Array.wrap(info['option_definition_refs']) }
        .transform_values { |ref| ref['name'] }
    end

    def get_option_definition(id, token)
      ...
      granules_by_id = granules.index_by {|g| g['id']}
      granule_options = order_info_to_granule_options(order_info)
      included_granule_ids = granule_options.select {|k, v| v.include?(option_name)}
      excluded_granule_ids = granule_options.reject {|k, v| v.include?(option_name)}
      granules = granules_by_id.values_at(*included_granule_ids)
      dropped_granules = granules_by_id.values_at(*excluded_granule_ids)

fixtures/cassettes/orders_requests.yml
```diff
@@ -83,6 +83,51 @@ http_interactions:
...
+      - eed-edsc-dev
+      Echo-ClientId:
+      - eed-edsc-dev
+      Echo-Token:
+      - edsc
+  recorded_at: Thu, 12 Nov 2015 16:47:06 GMT
+  digest: 31af7ba74170cc2541b61a8c978886bc1d3c38cc
+- request:
+    method: post
+    uri: https://api.echo.nasa.gov/echo-rest/orders/B83E748F-BEEF-528A-0265-CA6DDA7A8A78/order_items/bulk_action
```
bilts: This appears to be a re-recording of the previous fixture

spec/features/data_access/background_ordering_spec.rb
```diff
@@ -4,8 +4,41 @@
...
   orderable_collection_id = 'C90762182-LAADS'
   orderable_collection_title = 'MODIS/Aqua Calibrated Radiances 5-Min L1B Swath 250m V005'
+  aster_collection_id = 'C14758250-LPDAAC_ECS'
 
-  before :all do
+  context "ordering ASTER data" do
+    before :all do
+      Delayed::Worker.delay_jobs = true
+
+      load_page :search, {project: [aster_collection_id], view: :project}
```
bilts: I think the fixtures will end up actually trying to place the order. You should prevent them from doing this by using hand edited fixtures.

fixtures/cassettes/orders_responses.yml
```diff
@@ -92,6 +92,37 @@ d0c2a1b2a0fd3078fcb7492bc2ea067cd82834e8:
       encoding: UTF-8
       string: "<errors><error>Could not find order with guid [B83E748F-BEEF-528A-0265-CA6DDA7A8A78].</error></errors>"
     http_version: 
+31af7ba74170cc2541b61a8c978886bc1d3c38cc:
+  response:
+    status:
+      code: 404
```
bilts: Is this supposed to be 404? And line 93, too?

app/views/data_access/retrieval.html.erb
```diff
@@ -78,7 +78,22 @@
                  data-method="post"
                  data-confirm="Are you sure you want to remove this order?  This action cannot be undone."
                  rel="nofollow">Cancel</a>
-              <!-- /ko -->
+            <!-- /ko -->
+            <!-- ko if: dropped_granules -->
+              <p>The following granules will not be processed because they do not support the <strong data-bind="text: method_name"></strong> access method:</p>
+              <div class="granule-list">
+                <div>
```
bilts: This div doesn't seem to serve any purpose. I'm not sure the div above it does either.

PR #62, opened 2015-11-18
EDSC-820 Hide flyouts onclick and on content/overlay changing.

mightynimble:


app/assets/javascripts/modules/master_overlay.coffee
```diff
@@ -24,12 +25,18 @@ do (document, window, $=jQuery, config=@edsc.config, plugin=@edsc.util.plugin, p
...
 
     maximize: ->
+      @_hideFlyouts()
       @_minimized = false
       @_updateMinMaxState()
       @contentHeightChanged()
 
     showParent: -> @toggleParent(true)
 
+    _hideFlyouts: ->
```
bilts: The CollectionsCollapsedModel has a hideFlyout method. You should use that. This code ends up making assumptions about the internals of a component.

The cleanest way to handle this would be to have some sort of callback/event mechanism that gets triggered when the current pane is hidden and one that gets triggered when it's minimized or maximized. That would decouple the overlay from the specific components.

As I think you're encountering below, hiding flyouts this way leaks memory.

edsc/collections-collapsed/src/js/edsc-collections-collapsed.coffee
```diff
@@ -44,7 +44,7 @@ class CollectionsCollapsedModel extends KnockoutComponentModel
 
   toggleFlyout: (context, e) =>
     target = @_getFlyoutTarget(e)
-    if domData.get(target, 'flyout') || dom.hasClass(target, 'flyout-visible')
+    if dom.hasClass(target, 'flyout-visible')
```
bilts: I'm curious why you removed the first part of the condition.

edsc/collections-collapsed/src/js/edsc-collections-collapsed.coffee
```diff
@@ -53,7 +53,7 @@ class CollectionsCollapsedModel extends KnockoutComponentModel
 
   showFlyout: (context, e) =>
     target = @_getFlyoutTarget(e)
-    if target && !domData.get(target, 'flyout')
+    if target
```
bilts: Same as above. Why did you remove this?

edsc/collections-collapsed/src/js/edsc-collections-collapsed.coffee
bilts: That's what I was saying above about the memory leak. If you close it the way you did above, the domData remains and keeps listening to events because it's never destroyed.

app/assets/javascripts/modules/master_overlay.coffee
```diff
@@ -24,6 +25,7 @@ do (document, window, $=jQuery, config=@edsc.config, plugin=@edsc.util.plugin, p
       @contentHeightChanged()
 
     maximize: ->
+      @root.trigger('edsc.overlaychange')
```
bilts: These work coincidentally. Also need to hide overlays whenever the level changes.

edsc/collections-collapsed/src/js/edsc-collections-collapsed.coffee
```diff
@@ -22,6 +22,9 @@ class CollectionsCollapsedModel extends KnockoutComponentModel
     super(params, componentInfo)
     @_scrollParent = dom.scrollParent(@_root)
     @_flyouts = []
+    setTimeout((=>
```
bilts: I don't understand why the timeout is needed.

edsc/collections-collapsed/src/js/edsc-collections-collapsed.coffee
```diff
@@ -22,6 +22,9 @@ class CollectionsCollapsedModel extends KnockoutComponentModel
...
+      $(window).on 'edsc.overlaychange', @_hideFlyouts), 0)
```
bilts: This is the only dependency on jQuery in the plugin, so it's not desirable. It also operates globally, when only the control's own flyouts should be hidden

edsc/collections-collapsed/src/js/edsc-collections-collapsed.coffee
```diff
@@ -106,8 +109,11 @@ class CollectionsCollapsedModel extends KnockoutComponentModel
       domData.clear(flyout)
       dom.remove(flyout)
 
-  hideFlyout: (context, e) =>
-    target = @_getFlyoutTarget(e)
+  _hideFlyouts: (e) => @_destroyAndRemoveFlyout(target) for target in document.getElementsByClassName('flyout-visible')
```
bilts: This should just hide all if the component's own flyouts, which are maintained in a property.

bilts: I'd prefer using 2-3 lines here instead of one for readability

mightynimble:
Don't merge it yet - there are several map_zooming_spec tests randomly failing in the CI and locally...

PR #63, opened 2015-11-19
Pluggable backends and vis for CMR and CWIC

bilts:


mightynimble:
Travis somehow won't start the build automatically after pushes. Tried close and re-open the pr that worked...

PR #66, opened 2015-11-24
EDSC 807 Grab user info from URS.

mightynimble:


app/assets/javascripts/models/data/account.js.coffee
```diff
@@ -2,82 +2,20 @@ ns = @edsc.models.data
 
 ns.Account = do (ko, ajax=@edsc.util.xhr.ajax) ->
 
-  class Address
```
mightynimble: Removed phone and address class since they are no longer used.

app/assets/javascripts/models/data/account.js.coffee
```diff
@@ -2,82 +2,20 @@ ns = @edsc.models.data
...
-      @userTypeError = ko.observable(false)
-      @primaryStudyArea = ko.observable("")
-      @primaryStudyAreaError = ko.observable(false)
-      @address = new Address()
-      @region = ko.computed =>
-        if @address.country() == "United States" then "USA" else "INTERNATIONAL"
-
-      @phone = new Phone("BUSINESS")
-      @fax = new Phone("BUSINESS_FAX")
+      @email = ko.observable("Loading ...")
```
mightynimble: Initialize the text on contact-info page with loading... until the results come back from URS. Not sure if this is the right way.

app/assets/javascripts/models/data/account.js.coffee
```diff
@@ -100,36 +38,24 @@ ns.Account = do (ko, ajax=@edsc.util.xhr.ajax) ->
           else
             @errors(["Contact information could not be loaded, please try again later"])
 
-    updateContactInformation: (callback) =>
+    updateNotificationPreference: (callback) =>
```
mightynimble: This line and quite a few lines below are refactoring - users can only update the notification preferences now.

app/controllers/application_controller.rb
```diff
@@ -111,6 +111,7 @@ def use_collection(id)
   def clear_session
     store_oauth_token()
     session[:user_id] = nil
+    session[:user_name] = nil
```
mightynimble: In addition to :user_id, also store/clear user_name for easy interaction with URS.

app/controllers/users_controller.rb
```diff
@@ -21,15 +21,22 @@ def contact_info
   end
 
   def get_preferences
-    response = echo_client.get_preferences(get_user_id, token)
-    render json: response.body, status: response.status
+    preferences_response = echo_client.get_preferences(get_user_id, token, echo_client, session[:access_token])
+
+    urs_response = echo_client.get_urs_user(session[:user_name], session[:access_token])
```
mightynimble: An extra call to URS to get user profile in case the user has never visited Reverb before.

lib/echo/echo_client.rb
```diff
@@ -34,10 +34,39 @@ def get_phones(user_id, token)
...
+      return false if contact.blank? || contact['phones'].blank? || contact['address'].blank?
+      true
+    end
+
+    def get_preferences(user_id, token, client, access_token)
       response = get("/echo-rest/users/#{user_id}/preferences.json", {}, token_header(token))
       if response.status == 404
         response = update_preferences(user_id, {preferences: {}}, token)
+      elsif response.status == 200 && !general_contact_valid?(response.body['preferences']['general_contact'])
+        user_response = get_echo_user(user_id, token)
```
mightynimble: One call to get ECHO user, if it is a first-time EDSC user, get ECHO empty user obj with GUID. The other one get the content of the user from URS. Then merge the URS info to the ECHO user.

lib/echo/echo_client.rb
```diff
@@ -34,10 +34,39 @@ def get_phones(user_id, token)
...
+    def general_contact_valid?(contact)
+      return false if contact.blank? || contact['phones'].blank? || contact['address'].blank?
+      true
+    end
+
+    def get_preferences(user_id, token, client, access_token)
       response = get("/echo-rest/users/#{user_id}/preferences.json", {}, token_header(token))
       if response.status == 404
         response = update_preferences(user_id, {preferences: {}}, token)
+      elsif response.status == 200 && !general_contact_valid?(response.body['preferences']['general_contact'])
```
mightynimble: There are three "types" of users:
1. Users who have full profiles in ECHO (and in URS). These users exist before URS (with address, phone, and fax etc.)
2. Users who have partial profiles in ECHO - no address and/or phones. These users won't be able to successfully submit orders without a place holder in address and phone.
3. Users who have no profiles in ECHO. These users can't submit orders either.

bilts:
I get ECHO errors trying to submit orders from this branch, which I don't see in SIT. This happens when ECHO knows my user as well as when it doesn't:

POST https://api.echo.nasa.gov/echo-rest/order_information.json (200) [2.089903s]
POST https://api.echo.nasa.gov/echo-rest/orders.json
POST https://api.echo.nasa.gov/echo-rest/orders.json (201) [0.508838s]
Response: {"order"=>{"id"=>"2BE828BF-C0E2-490F-F072-457BAA76EA5B", "provider_orders"=>[]}}
POST https://api.echo.nasa.gov/echo-rest/orders/2BE828BF-C0E2-490F-F072-457BAA76EA5B/order_items/bulk_action
POST https://api.echo.nasa.gov/echo-rest/orders/2BE828BF-C0E2-490F-F072-457BAA76EA5B/order_items/bulk_action (400) [0.303115s]
GET https://api.echo.nasa.gov/echo-rest/users/891C369D-AB06-F479-B40B-2BB197C1E13F/preferences.json
GET https://api.echo.nasa.gov/echo-rest/users/891C369D-AB06-F479-B40B-2BB197C1E13F/preferences.json (200) [0.926554s]
GET https://api.echo.nasa.gov/echo-rest/users/891C369D-AB06-F479-B40B-2BB197C1E13F
GET https://api.echo.nasa.gov/echo-rest/users/891C369D-AB06-F479-B40B-2BB197C1E13F (200) [0.286355s]
PUT https://api.echo.nasa.gov/echo-rest/orders/2BE828BF-C0E2-490F-F072-457BAA76EA5B/user_information
PUT https://api.echo.nasa.gov/echo-rest/orders/2BE828BF-C0E2-490F-F072-457BAA76EA5B/user_information (200) [0.727486s]
POST https://api.echo.nasa.gov/echo-rest/orders/2BE828BF-C0E2-490F-F072-457BAA76EA5B/submit
POST https://api.echo.nasa.gov/echo-rest/orders/2BE828BF-C0E2-490F-F072-457BAA76EA5B/submit (422) [0.827698s]
Order response: {"order"=>{"id"=>"2BE828BF-C0E2-490F-F072-457BAA76EA5B", "provider_orders"=>[]}}
Items response: {"errors"=>{"error"=>"OrderItem is required and must have at least 1 entries"}}

app/controllers/users_controller.rb
```diff
@@ -21,15 +21,22 @@ def contact_info
...
   end
 
-  def update_contact_info
+  def update_notification_pref
     preferences = {preferences: params.delete('preferences')}
-    user_id = get_user_id
-
-    response = echo_client.update_preferences(user_id, preferences, token)
+    preferences[:preferences]['general_contact']['role'] = "Order Contact"
+    response = echo_client.update_preferences(get_user_id, preferences, token)
```
bilts: If I'm reading this correctly, you're relying on getting the whole preference object from the client but only updating the notification preference. Can you use get_preferences to avoid having to get all that information?

app/controllers/users_controller.rb
bilts: This also gets called if the user has visited Reverb, and will overwrite whatever preferences are stored in ECHO. I think that may be what users would expect, though, so I wouldn't change it.

app/views/users/_contact_info_form.html.erb
```diff
@@ -1,116 +1,90 @@
...
-    <input data-bind="value: lastName" id="lastName" size="30" type="text">
-  </p>
-  <p class="required">
-    <label for="email">Email</label>
-    <input data-bind="value: email" id="email" size="30" type="text">
-  </p>
   <p>
-    <label for="organizationName">Organization name</label>
-    <input data-bind="value: organizationName" id="organizationName" size="30" type="text">
+    <a href="<%="#{Rails.configuration.services['earthdata'][Rails.configuration.echo_env]['urs_root']}/profile/edit" %>" target="_blank" class="button text-button">Edit Profile in Earthdata Login <i class="fa fa-arrow-circle-right"></i></a> <span>(in a new page)</span>
```
bilts: Could you remove the "In a new page" text?

app/views/users/_country_options.html.erb
```diff
@@ -1,196 +0,0 @@
...
-<option value="United Kingdom">United Kingdom</option>
-<option value="Uruguay">Uruguay</option>
-<option value="Uzbekistan">Uzbekistan</option>
-<option value="Vanuatu">Vanuatu</option>
-<option value="Vatican City">Vatican City</option>
-<option value="Venezuela">Venezuela</option>
-<option value="Vietnam">Vietnam</option>
-<option value="Yemen">Yemen</option>
-<option value="Zambia">Zambia</option>
-<option value="Zimbabwe">Zimbabwe</option>
```
bilts: So glad to see this go.

fixtures/cassettes/collections_requests.yml
```diff
@@ -2,6 +2,19 @@
...
     method: get
+    uri: https://cmr.earthdata.nasa.gov/search/collections.json?echo_collection_id%5B%5D=C179003030-ORNL_DAAC&include_has_granules=true&include_granule_counts=true
+    body:
+      encoding: US-ASCII
+      string: ''
+    headers:
+      User-Agent:
+      - Faraday v0.8.8
+      Client-Id:
+      - eed-edsc-dev
```
bilts: It's strange that this request ended up getting recorded. It should have been recorded ages ago. Is it possible the Client-Id is causing us to re-record things?

lib/echo/echo_client.rb
bilts: There's a slight problem here. On the configuration page, we're showing contact info coming from URS even if the user has ECHO contact info. If we then use the ECHO contact info (which could be different than the URS contact info) the user may get emails sent to an unexpected address or something of that sort. I'd say always use the URS contact info and override any ECHO contact info.

fixtures/cassettes/hand-edited_responses.yml
```diff
@@ -1259,6 +1103,42 @@ second_service_request:
...
+      status:
+      - 2000 OK
+      x-frame-options:
+      - SAMEORIGIN
+      x-xss-protection:
+      - 1; mode=block
+      x-content-type-options:
+      - nosniff
+      www-authenticate:
+      - Bearer realm="URS",error="invalid_token"
```
bilts: This doesn't seem right

fixtures/cassettes/hand-edited_requests.yml
```diff
@@ -351,4 +274,94 @@ http_interactions:
...
+    method: get
+    uri: https://urs.earthdata.nasa.gov/api/users/edsc?client_id=Gsk6IBbEfa98sbE5UavsZQ
+    body:
+      encoding: US-ASCII
+      string: ''
+    headers:
+      User-Agent:
+      - Faraday v0.8.8
+      Authorization:
+      - Bearer 47398e33eef3a1a2cd7ad86bc7a3dc182ea8f4b39b6ebb4992295acced920469
```
bilts: I think you've recorded a real token here

app/controllers/users_controller.rb
mightynimble: When updatingn the notification pref is done, ECHO will return a GUID for phone and address elements which is required when submitting the order. So a whole preference obj less GUID is constructed in the coffeescript and is sent to the controller who submits the content to ECHO-Rest.

app/views/users/_contact_info_form.html.erb
mightynimble: Removed.

fixtures/cassettes/hand-edited_responses.yml
mightynimble: Corrected.

fixtures/cassettes/collections_requests.yml
mightynimble: Not sure why it hasn't ever record this request. I checked the ```edsc_configurer.rb``` - we don't match_request_on client-id in this situation. I also changed the client_id to something else and ran the test, no new fixtures were added.

fixtures/cassettes/hand-edited_requests.yml
mightynimble: This is a dup recording. Removed.

lib/echo/echo_client.rb
mightynimble: Not quite sure what you mean...
Firstly, there are two routes that can hit this ```def get_preferences``` method:
 - From an AJAX (```account.coffee```) when user is on the contact_info page
   - It will go through ```users_controller#get_preferences``` which contacts URS and gets the user profile there.
 - When user tries to place an order
   - if it returns a 404, the ```general_contact``` will be constructed in ```def create_order```
   - if it returns a 200 but without a valid ```general_contact```, the contact will be constructed in this method and another PUT will be sent out.

app/controllers/application_controller.rb
bilts: For these sorts of generally useful comments, could you please put them in code comments instead of pull request comments so we can understand the code better in the future?

PR #67, opened 2015-11-25
EDSC-842 Keep browse image ratio in collection and granule list views.

mightynimble:


app/assets/stylesheets/modules/panel_list.css.scss
```diff
@@ -1,4 +1,4 @@
-$panel-list-item-height: 85px;
+$panel-list-item-height: 75px;
```
bilts: We're requesting a 65px image from the browse scaler, but it ends up being drawn at 75px on the screen. Please update the browse scaler URL to request the correct size.

app/assets/stylesheets/modules/panel_list.css.scss
```diff
@@ -80,12 +80,18 @@ $panel-list-thumbnail-size: $panel-list-item-height;
...
   position:relative;
   margin-right: .5em;
+  width: $panel-list-thumbnail-size + 10px;
+  height: $panel-list-thumbnail-size + 10px;
+  background-color: #222;
+  border: 5px solid #323232;
 }
 
 .panel-list-thumbnail {
   height: $panel-list-thumbnail-size;
```
bilts: What happens when images are wider than they are tall?

mightynimble: The ```max-width``` below will handle that.

app/assets/stylesheets/modules/panel_list.css.scss
mightynimble: I see we are requesting *85px* images in ```collection.coffee```:
```
      if collection_id?
        "#{scalerUrl}/datasets/#{collection_id}?h=85&w=85"
      else if granule?
        "#{scalerUrl}/granules/#{granule}?h=85&w=85"
```
Should I update this to 75px? Or you are referring somewhere else?

bilts: That's on the collection. I was looking at the granule thumbnail. Both of them probably need to be fixed.

app/assets/stylesheets/modules/panel_list.css.scss
bilts: Change line 17 of granule.coffee from:

    "#{scalerUrl}/granules/#{@id}?h=#{h}&w=#{w}"

to:

    "http://placehold.it/30x65"

Which is an image taller than it is wide. The code looks correct.  Now change it to:

    "http://placehold.it/65x30"

Which is an image wider than it is tall. The code stretches it into a square. It should be centered and maintain aspect ratio.

bilts:
The "No Image Available" image on collections is broken with this change

app/assets/javascripts/models/data/collection.js.coffee
```diff
@@ -125,9 +125,9 @@ ns.Collection = do (ko
       granule = @browseable_granule
       collection_id = @id for link in @links when link['rel'].indexOf('browse#') > -1
       if collection_id?
-        "#{scalerUrl}/datasets/#{collection_id}?h=85&w=85"
+        "#{scalerUrl}/datasets/#{collection_id}?h=75&w=75"
```
bilts: Can you define the dimension in one place and use that constant? I'm uncomfortable of sprinkling the magic number 75 everywhere. Also, it seems like we've failed to update it consistently in the past.

mightynimble: I will put it in config.js.coffee.erb.

bilts:
This still breaks the layout of "No Image Available" albeit less so (compare to ops, which has no light gray border). Also, the GIBS and NRT badges are no longer appearing.

bilts:
GIBS and NRT badges appear below thumbnails in the collections list

bilts:
Landscape-oriented images are not quite vertically centered, probably due to extra whitespace (Please solve for real, don't just subtract 3 from the line-height).

mightynimble:
Fixed "No Image Available" borders
Fixed GIBS and NRT badges.
About the vertical-align issue: it seems that it works as it should be. Take ASTER L1A Reconstructed Unprocessed Instrument Data V003 as an example here:
The browse is 70px tall which left 5px blank space. Vertical-align divides the 5 into 3 and 2 px. If you zoom in the page so that you can measure it, you will see that the top is 1px (2px if you zoom in 2X) taller than the bottom.

PR #70, opened 2015-12-02
EDSC-883 Backend changes to make facets show up on keyword searches.

mightynimble:


spec/features/collection_facets/collection_facets_spec.rb
```diff
@@ -246,6 +246,26 @@
...
+  context "when applied science keyword facets and search terms filter the collections list to no results" do
+    before(:all) do
+      find(".facets-item", text: "BIOSPHERE").click
+      fill_in :keywords, with: "somestringthatmatchesnocollections"
+      wait_for_xhr
+    end
+
+    after(:all) do
+      reset_search
+      find(".facets-item", text: 'BIOSPHERE').click
```
bilts: Doesn't reset_search reset the facets list, too? If so, performing this click could apply the biosphere facet and break subsequent tests.

bilts:
This doesn't work right if there are multiple levels of keywords selected. Choose "Atmosphere" then "Aerosols" then "Aerosol Extinction" then put in the keyword "blargh."  The facet display reverses order.

It *may* be as simple as just doing a .reverse on something you're returning, but I'd want an explanation of why that is the guaranteed to be the right answer, because it could just be a coincidence of URL order or alphabetical order or something like that. We need a solution that is always right.

Also, please add a test for this case.

mightynimble:
I noticed this issue and tried ```selected_keywords.insert(0, {...``` to address this. Apparently, something else needs to be looked at.

PR #71, opened 2015-12-02
EDSC-832 Config change to point SIT to testbed.

mightynimble:


config/environments/sit.rb
```diff
@@ -77,6 +77,9 @@
   config.feedback_url = 'https://fbm.uat.earthdata.nasa.gov/for/EdSearch_SIT/feedback.js'
 
   config.url_limit = 2000
+  config.echo_env = 'testbed'
+  services = config.services
+  config.urs_client_id = services['urs'][Rails.env.to_s][services['earthdata'][config.echo_env]['urs_root']]
```
mightynimble: Just followed the example in ```uat.rb```. Is this the only place to make the change?

PR #72, opened 2015-12-07
EDSC-807 Address an out-of-date session issue after pointing SIT to t…

mightynimble:
…estbed.


mightynimble:
The issue is that when a logged in user in SIT might see an error the first time he/she loads the contact_info page due to a nil:NilClass error. (```preferences_response.body['preferences']``` is nil)

You are right about not always returning 200 - I have changed the code to forward non-200 response back to the front end.

bilts:
I may be missing something, but it still appears to always return 200.

PR #77, opened 2016-01-04
EDSC-902 Add carriage returns to the descriptions.

mightynimble:


spec/features/collections/collection_details_spec.rb
```diff
@@ -123,4 +123,17 @@
...
+  context "when selecting a collection with multiple lines of description" do
+    before :all do
+      load_page :search
+      fill_in 'keywords', with: 'C1029-NSIDCV0'
+      wait_for_xhr
+      first_collection_result.click_link('View collection details')
+    end
+
+    it "displays carriage returns in the description" do
+      expect(page).to have_content('to April 1992.')
```
bilts: This doesn't actually test the presence / rendering of carriage returns.

app/views/search/_collection_details.html.erb
```diff
@@ -81,7 +81,14 @@
         </div>
 
         <div class="row">
-          <div class="col-sm-12 long-paragraph collapsed" data-bind="text: $data.description"></div>
+          <div class="col-sm-12 long-paragraph collapsed">
```
bilts: I think you can and should implement this in CSS rather than adding code to the backend and frontend:

    .long-paragraph {
        white-space: pre-line;
    }

PR #79, opened 2016-01-11
EDSC-898: Rename the button text and refactored tests accordingly.

mightynimble:


mightynimble:
The "Retrieve Listed Data" will be changed again.

PR #80, opened 2016-01-12
EDSC-842: Fix the url in config also addressed a browse-scaler error …

mightynimble:
…when loading the image. Added but commented out a test since we can't really test it in test env.

### DIFF
--- a/app/assets/javascripts/config.js.coffee.erb
+++ b/app/assets/javascripts/config.js.coffee.erb
@@ -8,11 +8,11 @@
   echo_env = if match then match[1] else '<%= Rails.configuration.echo_env %>'
 
   if echo_env == 'partnertest'
-    browse_scaler_url = "https://api-test.echo.nasa.gov/browse-scaler/browse_images"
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/config.js.coffee.erb
```diff
@@ -3,6 +3,17 @@
 
   csrfToken = document.querySelector?('meta[name=csrf-token]')?.content
 
+  query = window.location.search
+  match = /echo_env=(\w+)/.exec(query)
+  echo_env = if match then match[1] else '<%= Rails.configuration.echo_env %>'
+
+  if echo_env == 'partnertest'
+    browse_scaler_url = "https://api-test.echo.nasa.gov/browse-scaler/browse_images"
```
bilts: Can you still use Rails.configuration for this?  Something like:

    browse_scaler_url = "<%= Rails.configuration.services['earthdata']['partnertest']['echo_root'] %>/browse-scaler/browse_images"

That way we still maintain the ECHO roots in a single place rather than hard-coding some places and using rails config elsewhere.

PR #82, opened 2016-01-12
EDSC-911: Remove 'not provided' version_id from expanded and collapse…

mightynimble:
…d collection list if the collection doesn't have one.

### DIFF
--- a/app/views/search/_collection_list_item.html.erb
+++ b/app/views/search/_collection_list_item.html.erb
@@ -29,7 +29,7 @@
               <p>
                 <span class="panel-list-details" <% if is_project %>data-bind="style: {'background-color': $root.project.colorForCollection($data)}"<% end %>>
                   <span data-bind="text: short_name"></span>
-                  <!-- ko if: $data.version_id && $data.version_id !== 'Undefined' -->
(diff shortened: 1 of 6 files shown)


app/views/search/_collection_list_item.html.erb
```diff
@@ -29,7 +29,7 @@
               <p>
                 <span class="panel-list-details" <% if is_project %>data-bind="style: {'background-color': $root.project.colorForCollection($data)}"<% end %>>
                   <span data-bind="text: short_name"></span>
-                  <!-- ko if: $data.version_id && $data.version_id !== 'Undefined' -->
+                  <!-- ko if: $data.version_id && $data.version_id !== 'Undefined' && $data.version_id !== 'Not provided' -->
```
bilts: Sorry for not being more clear on this, but this is what I was saying we shouldn't do. If it's a metadata problem, we need to kick it back. People are hacking around a metadata requirement. I wonder if anything is 'Undefined' currently.

mightynimble: Oh, yeah, you are clear enough - I was just not listening; somehow didn't see the chat..
What should we ask CMR to provide to us if there is no version_id for a collection? I can write a ticket against CMR. 
So close it?

PR #106, opened 2016-02-04
EDSC-918: Point to CMR for OSDDs and fix dataCenter OSDD param

bilts:


app/presenters/collection_details_presenter.rb
```diff
@@ -51,7 +54,7 @@ def initialize(collection, collection_id=nil, token=nil, env='prod')
     @collection.dif_url = "#{metadata_url}.dif#{url_token}"
     @collection.smap_iso_url = nil #"#{metadata_url}.smap_iso"
     opensearch_url = "#{Rails.configuration.services['earthdata'][env]['opensearch_root']}/granules/descriptor_document.xml"
-    @collection.osdd_url = "#{opensearch_url}?utf8=%E2%9C%93&clientId=#{Rails.configuration.cmr_client_id}&shortName=#{URI.encode_www_form_component(@collection.short_name)}&versionId=#{@collection.version_id}&dataCenter=#{URI.encode_www_form_component(@collection.archive_center)}&commit=Generate#{url_token}"
+    @collection.osdd_url = "#{opensearch_url}?utf8=%E2%9C%93&clientId=#{Rails.configuration.cmr_client_id}&shortName=#{URI.encode_www_form_component(@collection.short_name)}&versionId=#{@collection.version_id}&dataCenter=#{URI.encode_www_form_component(data_center)}&commit=Generate#{url_token}"
```
mightynimble: Should we backport this file as well along with EDSC-907? (well, I guess it doesn't hurt to backport the whole commit since opensearch prod url is live already.)

bilts: They've set up a redirect, so I don't think it's a big enough deal to need a backport.

mightynimble: I mean the link will be incorrect after we backport the data center fix if we don't replace ```@collection.archive_center``` with ```data_center``` in the url above.

PR #110, opened 2016-02-09
EDSC-797: Hide all filters but temporal for CWIC granules

mightynimble:


mightynimble:
## Existing issues and questions of this pull request:

**ISSUES**

- The url doesn't have temporal param `pg[0][qt]`. So a refresh of the page will lose temporal constraint.
- CWIC endpoint doesn't accept encoded time parameters (a 400, see `FIXME` in `lib/echo/cwic_client.rb`.)

**QUESTIONS**
 - Should we support recurring temporal search? The CWIC endpoint doesn't support this feature.(http://cwic.wgiss.ceos.org/opensearch/datasets/gov.noaa.class.J2-XGDR/osdd.xml?clientId=edsc-dev)

bilts:
If CWIC doesn't support recurring temporal search, we cannot support it. It should be disabled on the granule details view, and we need a new story to deal with what to do if the user applies collection-level recurring temporal.

app/assets/javascripts/models/data/collection.js.coffee
```diff
@@ -259,7 +256,7 @@ ns.Collection = do (ko
       null
 
     canFocus: ->
-      @hasAtomData() && @granuleDatasourceName()
+      @hasAtomData() && @granuleDatasourceName()?
```
bilts: Was there an issue that required the `?` to be added?

app/assets/javascripts/modules/banner.coffee
```diff
@@ -27,7 +27,7 @@
       $banner.data('banner.persist', options.persist)
       $('body').after($banner)
       # Do this in a timeout so the element has time to be placed in the DOM and animations can happen
-      setTimeout((-> $banner.removeClass('banner-hidden')), 0)
+      setTimeout((-> $banner?.removeClass('banner-hidden')), 0)
```
bilts: See previous comment. The banner should be in the DOM. Is there a case where it's not?

app/assets/javascripts/modules/map/leaflet-plugins/interpolation.js.coffee
```diff
@@ -22,13 +22,7 @@ ns.interpolation = do (L, gcInterpolate = window.edsc.map.geoutil.gcInterpolate,
...
     return [] if latLngs.length == 0
 
-    # Clone path and set its last element to its first so we can interpolate the last segment
-    if Math.abs(latLngs[0].lng) == Math.abs(latLngs[latLngs.length - 1].lng) == 180
-      # In this case the last element is an artificial longitude crossing.  Avoid interpolating
-      # with the first to prevent drawing strokes along the dateline
-      latLngs = latLngs.concat(latLngs[latLngs.length - 1])
-    else
-      latLngs = latLngs.concat(latLngs[0])
+    latLngs = latLngs.concat()
```
bilts: Any idea what happened here? It looks like you're re-applying my polygon rendering fixes.

app/assets/javascripts/models/data/collection.js.coffee
mightynimble: Without the "?", canFocus will not always be eval to a boolean value. Therefore the granule filter button will not show up for cwic collection.

app/views/search/_granule_search.html.erb
```diff
@@ -1,97 +1,116 @@
...
-        or
-        <a href="#" class="search-granules-multiple button-small button-outline-light">
-          Search Multiple <i class="fa fa-arrow-circle-o-right"></i>
-        </a>
-      </section>
-      <section class="multiple-granule-id-search">
-        <p>
-          <a href="#" class="search-granules-one button-small button-outline-light">
-            <i class="fa fa-arrow-circle-o-left"></i> Search One
+    <!--ko ifnot: cwic()-->
```
bilts: I'd really like to find a better way to do this, long-term. The whole panel could/should probably be swapped.

config/services.yml
```diff
@@ -4,19 +4,19 @@ earthdata:
     urs_root: "https://urs.earthdata.nasa.gov"
     echo_root: "https://api.echo.nasa.gov"
     cmr_root: "https://cmr.earthdata.nasa.gov" # Does not exist yet
-    opensearch_root: "https://api.echo.nasa.gov/opensearch"
+    opensearch_root: "https://cmr.earthdata.nasa.gov/opensearch"
```
bilts: Again, shouldn't this already be in master?

app/assets/javascripts/modules/banner.coffee
mightynimble: I have seen an error in the console twice. So I just added it there but I didn't investigate why the error occurs...

spec/features/cwic/granules/cwic_granule_filtering_spec.rb
```diff
@@ -0,0 +1,50 @@
+require "spec_helper"
+
+describe "CWIC Granule list", reset: false do
```
bilts: All acceptance criteria should have a matching spec that is tagged "acceptance: true"

app/assets/javascripts/modules/map/leaflet-plugins/interpolation.js.coffee
mightynimble: I rebased the branch against master, which may pull in some commits.

spec/features/data_access/service_access_spec.rb
```diff
@@ -66,23 +66,21 @@
...
-        # this test and the one below are quite flaky.
-        # Cannot reliably display a progress bar using recordings
-        xit 'displays a progress bar while the service is processing' do
-          wait_for_xhr
-          if page.has_content?("AMSR-E/Aqua 5-Day L3 Global Snow Water Equivalent EASE-Grids V002 Processing")
-            expect(page).to have_content('of 12 items processed')
-            expect(page).to have_css('div.progress-bar')
-          else
-            expect(page).to have_content("AMSR-E/Aqua 5-Day L3 Global Snow Water Equivalent EASE-Grids V002 Complete")
+        end
```
bilts: Another case of something that I thought was in master.

app/views/search/_granule_search.html.erb
mightynimble: Totally!!

spec/features/cwic/granules/cwic_granule_filtering_spec.rb
mightynimble: I will add the tag. Thanks!

app/assets/javascripts/modules/banner.coffee
bilts: Please investigate or file a bug. This just ends up hiding a legitimate underlying problem.

PR #113, opened 2016-02-12
EDSC-818: Download CWIC granule metadata.

mightynimble:


app/assets/javascripts/models/data/collection.js.coffee
```diff
@@ -281,12 +281,18 @@ ns.Collection = do (ko
       @cwic(@_isCwic(jsonObj))
 
       @nrt = jsonObj.collection_data_type == "NEAR_REAL_TIME"
-      @granuleCount(jsonObj.granule_count)
 
       for own key, value of jsonObj
         this[key] = value unless ko.isObservable(this[key])
 
+      if @_isCwic(jsonObj)
```
mightynimble: CMR doesn't report ```granule_counts``` and ```has_granules``` if it doesn't hold any. So I hacked around this to provide ```granule_counts(1)``` so that the collection behaves like other CMR collections with granules.

bilts:
Two feature changes necessary:

1. Just remove the granule size message. We do that even with CMR granules when we can't determine their size.
2. Remove the link to the access script. It queries CMR, and there's almost no chance we can make it work, since most remote CWIC endpoints require auth.

app/models/opendap_configuration.rb
```diff
@@ -160,22 +160,30 @@ def info_urls_for(granule)
...
+        Array.wrap(links_for(granule['link'], 'enclosure', /enclosure/).first)
+      else
+        links_for(granule['links'], 'enclosure', /\/data/)
+      end
     end
   end
 
-  def browse_urls_for(granule)
-    links_for(granule, /\/browse/)
+  def browse_urls_for(granule, is_cwic=false)
```
bilts: This needs to be updated to match the logic in the browse URL branch.

Also, for whatever reason, I'm not seeing browse links show up on the download page for my test collection (http://edsc.dev/search?cmr_env=uat&q=USGS_EDC_EO1_ALI&ff=Int%27l+%2F+Interagency)

app/views/data_access/configure.html.erb
```diff
@@ -39,6 +39,7 @@
             <p class="summary" data-bind="if: granuleAccessOptions().size && granuleAccessOptions().size != 0">
               <span class="summary-data" data-bind="text: granuleAccessOptions().size"></span>
               <span class="summary-label" data-bind="text: granuleAccessOptions().sizeUnit"></span>
+              <span class="summary-caveat" data-bind="if: collection.cwic()">This is <span class="cwic-label" data-bind="text: collection.granuleDescription"></span> data. Granule size is not available. </span>
```
bilts: As mentioned in the overall comment, I think this should be removed. Just make sure `granuleAccessOptions().size` is null or 0 and the whole thing won't be displayed, which is what we do for CMR.

app/models/opendap_configuration.rb
mightynimble: I edited the ```db/seeds.rb``` file to set the browse_flag so that the "view browse link" button will show up. I guess it is not the way to implement?

db/seeds.rb
```diff
@@ -6,6 +6,7 @@
 load_extra && CollectionExtra.create!([
   {echo_id: "C179003030-ORNL_DAAC", has_browseable_granules: false, has_granules: true, browseable_granule: nil, granule: "G179109563-ORNL_DAAC", searchable_attributes: {"attributes"=>[]}, orbit: {}},
   {echo_id: "C179001887-SEDAC", has_browseable_granules: false, has_granules: false, browseable_granule: nil, granule: nil, searchable_attributes: {"attributes"=>[]}, orbit: {}},
+  {echo_id: "C1000003579-GCMDTEST", has_browseable_granules: true, has_granules: true, browseable_granule: 'http://cwic.wgiss.ceos.org/opensearch/granules.atom?uid=3DIMG_L2P_IRW:MOSDAC:3DIMG_L2P_IRW:18JUL2014:0400', granule: nil, searchable_attributes: {"attributes"=>[]}, orbit: {}},
```
bilts: This will be overridden by a background process in collection_extra which reloads granule data from the CMR. And I think that's a good thing, because otherwise, this risks errors in that background process, as the process expects that if something has granules and has browseable granules, then we can query the CMR for it.

So, please remove.

app/controllers/granules_controller.rb
```diff
@@ -42,7 +42,10 @@ def timeline
 
   class GranuleUrlStreamer
     def initialize(params, token, url_mapper, echo_client, url_type=:download)
-      params.reject!{|p| ['datasource', 'short_name'].include? p}
+      @is_cwic = params['datasource'].present? && params['datasource'] != 'cmr'
```
bilts: This should just be params['datasource'] == 'cwic'. As-is, this prevents us from having other data sources

db/seeds.rb
mightynimble: (See my question above.) CMR doesn't report ```has_browseable_granules``` for CWIC collections. How/where can we get the information if not from this seed file?

app/controllers/granules_controller.rb
```diff
@@ -42,7 +42,10 @@ def timeline
 
   class GranuleUrlStreamer
```
bilts: I'd prefer to see two completely separate URL streamers, one for CMR and one for CWIC (`CmrGranuleUrlStreamer`, `CwicGranuleUrlStreamer`). They probably deserve their own files. In general, we should limit the amount that we have to use if-logic to distinguish between CWIC and CMR and use methods or classes instead.

app/models/opendap_configuration.rb
```diff
@@ -160,22 +160,30 @@ def info_urls_for(granule)
     end
   end
 
-  def download_urls_for(granule)
+  def download_urls_for(granule, is_cwic=false)
```
bilts: There should be a separate `CwicUrlMapper` class instead of using if-logic to overwrite methods.

PR #114, opened 2016-02-18
EDSC-942: CSS change to force a newline for long non-breaking org name.

mightynimble:


app/assets/stylesheets/modules/master_overlay.css.scss
```diff
@@ -443,6 +443,7 @@ $parent-width: $parent-split * 0.21;
     font-weight: bold;
     color: #FFFFFF;
     margin-bottom: 10px;
+    word-wrap: break-word;
```
bilts: Hey, I learned something! I've never seen this attribute.

There's an additional problem here, though. If you look at the example collection in the ticket, it appears as though it actually has two archive centers, which is one reason it's so long. I think serializing them to a string joins them by a "," (no whitespace) causing the line breaking problem. Note that it still needs this CSS even after breaking properly.

My recommendation here is to grab the first archive center if there are multiple in the echo10 format, because it will match what the CMR provides as the archive_center field for its json response.

Line 14 of CollectionDetailsPresenter is the one you want. It should be something like:

    @collection.archive_center = Array.wrap(collection_xml['ArchiveCenter']).first

PR #120, opened 2016-02-23
Add synchronize to wait for zoom in to finish.

mightynimble:


spec/features/address_bar_spec.rb
```diff
@@ -321,8 +321,10 @@ def query_string
     before(:all) do
       visit '/search/map'
       wait_for_xhr
-      page.execute_script("$('#map').data('map').map.setView(L.latLng(12, -34), 5)")
-      wait_for_xhr
+      synchronize do
```
bilts: Were you able to reproduce the failure? This shouldn't fix the failure. synchronize just runs the block repeatedly until there's no exception, and there was no exception on master.

mightynimble: No I couldn't reproduce it. I saw some requests, if not all, failed from the test output. Then maybe I can add a wait for animation to complete? I saw that there was a very short period of time delay for the 'm' param to show in the url after the script is executed.

bilts: A wait_for_zoom_animation(5) followed by a wait_for_xhr would probably work.

PR #121, opened 2016-02-24
EDSC-946 EDSC site tour glitches

mightynimble:


bilts:
The tour doesn't work for me anymore. It doesn't advance past 'AQUA'. Are you testing against ops data, and are you not relying on "Recent" collections?

PR #126, opened 2016-03-01
EDSC-971 Sort facets case-insensitively.

mightynimble:


bilts:
You committed a screenshot. Please remove it.

spec/features/data_access/service_access_spec.rb
```diff
@@ -77,8 +77,8 @@
             is_esi_completed = true if page.has_content?("Complete")
             sleep 0.5
             retries += 1
+            Rails.logger.info "-- Retried: #{retries}"
```
bilts: This is almost certainly going to cause a merge conflict

bilts:
This needs a test

spec/features/data_access/service_access_spec.rb
mightynimble: I will revert this change since you are working on it in another pr already.

mightynimble:
I added a test in ```collection_facets_spec``` "shows facets in a case insensitive order". Is this what you are expecting?

PR #130, opened 2016-03-02
ECHO-592 Change border corder of browse image container.

mightynimble:
browse-scaler changes are on branch ```feature/echo-592``` in ```echo-browser-scaler``` which is behind firewall (not in stash), so no pull request for that :(

Well, the best way to review this is through a gotomeeting so that you don't have to set up the ECHO stack.

### DIFF
--- a/spec/features/collections/collection_results_spec.rb
+++ b/spec/features/collections/collection_results_spec.rb
@@ -55,6 +55,12 @@ describe "Collection results", reset: false do
     expect(page).to have_css('.badge-opendap')
   end
 
+  it "displays  for collections which have no thumbnail URLs" do
+    fill_in "keywords", with: 'C179003030-ORNL_DAAC'
+    wait_for_xhr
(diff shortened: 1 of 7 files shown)


bilts:
Can you explain why we're changing the border color? I think the design team set the current border color.

mightynimble:
I will, when we are on the goto meeting later.

mightynimble:
I will send you the browse-scaler controller changes in an email.

spec/features/collections/collection_results_spec.rb
```diff
@@ -55,13 +55,6 @@
     expect(page).to have_css('.badge-opendap')
   end
 
-  it "displays a placeholder for collections which have no thumbnail URLs" do
```
bilts: This spec should still be here, but just assert the image is image_unavailable.svg

bilts:
Small fix, but otherwise I approve of this pull request, including the scaler changes you sent via email.

Please merge/deploy the browse scaler changes before merging this.

PR #131, opened 2016-03-03
EDSC-968 Display all bounding boxes if there are more than one.

mightynimble:


mightynimble:
@bilts I'd say this is ok to merge since we have addressed the failed specs in another PR.

PR #134, opened 2016-03-07
EDSC-960 Show modal for orders with more than 2000 granules

mightynimble:


bilts:
If you choose two access methods, the first one being "Download" and the second being an order, the dialog appears. If you choose them in the opposite order, it doesn't.

PR #137, opened 2016-03-09
EDSC-812 Add a modal to display CWIC related info.

mightynimble:


mightynimble:
The failed test is the CWIC img one. The content of the modal needs review though.

app/assets/stylesheets/modules/buttons.css.scss
```diff
@@ -161,10 +161,18 @@ button:disabled {
   }
 }
 
-.modal-button {
-  position: absolute;
-  right: 20px;
-  bottom: 20px;
+.whats-cwic {
```
bilts: Style note (probably not worth fixing at the moment): This class name is heavily tied to the specific modal dialog. It'd be better to have it be something more generic/reusable in case we need other detailed help in the future.

app/views/search/_whats_cwic_modal.html.erb
```diff
@@ -0,0 +1,69 @@
+<div class="modal fade" id="whats-cwic-modal" tabindex="-1" role="dialog" aria-labelledby="whats-cwic-modal-title">
```
bilts: I'm going to move this into the renderer. The purpose of renderer plugins is to hold these sort of one-off things that would otherwise bloat the page.

app/views/layouts/application.html.erb
```diff
@@ -99,5 +99,7 @@
         })();
       </script>
     <% end %>
+
+    <%= render "search/whats_cwic_modal" %>
```
bilts: I'm moving this, but for future reference, this sort of thing doesn't belong in application.html.erb. Putting it here will cause it to be included in all the accounts pages, data access, etc, when it's only needed in the granules list for some collections

PR #138, opened 2016-03-14
EDSC-981: empty query param &pg=[] causes 500 error in rails.

mightynimble:


bilts:
Could you also fix the underlying problem on the controller-side? With this change, it's still possible that someone has an old bookmark / retrieval / saved project / with a URL that will trigger 500 errors.

mightynimble:
&pg=[] is a malformed param which will not reach our controller. In OPS, if you visit a link that has this param, a 400 error will be shown in the console and an blank page will be displayed.

PR #140, opened 2016-03-15
Fix issue where CWIC visualization plugin gets unloaded

bilts:
and doesn't clean itself up


mightynimble:
Found an issue that may not related to this:
1. Go to http://edsc.dev/search/granules?cmr_env=uat&p=C1204595275-GCMDTEST&m=0!0!2!1!0!&ff=Int%27l+%2F+Interagency&fpj=CWIC
2. Draw a polygon over Spain (polygon=1.125%2C43.03125%2C-8.71875%2C43.734375%2C-9%2C37.828125%2C-0.421875%2C36.984375%2C1.125%2C43.03125)
3. The granule list should be cleared out if there is no matching granules. However it shows "20 of 0 matching granules".

PR #146, opened 2016-03-21
EDSC-795: Display CWIC granule detail information (on information tab).

mightynimble:


app/controllers/cwic_controller.rb
```diff
@@ -15,6 +15,12 @@ def download
     render content_type: 'text/html', layout: false
   end
 
+  def granule
```
bilts: I'm not sure you needed a separate route for this. I think you could just use the `/cwic` route, but maybe I'm missing something. The separate route for `/cwic/edsc_download` allows the HTML rendering, but `/cwic` is just a pass-through.

If for some reason you *do* need a separate route, call it "edsc_granule" in the URL to avoid clashing with legitimate CWIC routes.

app/assets/stylesheets/modules/master_overlay.css.scss
```diff
@@ -394,6 +394,8 @@ $parent-width: $parent-split * 0.21;
 
 .details-content {
   white-space: pre;
+  font-family: monospace;
```
bilts: How does this affect the current rendering?

edsc/plugins/datasource/cwic/src/js/Granule.jsx
```diff
@@ -1,12 +1,39 @@
...
+import * as CwicUtils from './CwicUtils.jsx';
+
+let ajax = window.edsc.util.xhr.ajax;
 
 let CwicGranule = (function() {
   extend(CwicGranule, window.edsc.models.data.Granule);
 
   function CwicGranule(jsonData) {
     CwicGranule.__super__.constructor.apply(this, Array.prototype.slice.call(arguments));
+    this.details = this.asyncComputed({}, 100, this._computeGranuleDetails, this);
```
bilts: I don't think you need this. Rather, you shouldn't nee this, so please find a way get rid of it. Inheritance should cause the correct method to be used.

mightynimble: You are right. I don't need this line.

app/assets/stylesheets/modules/master_overlay.css.scss
mightynimble: This also affects CMR granule details but I think make it monospace the indentation will be more clear.

app/controllers/cwic_controller.rb
mightynimble: I will rename the route to ```/cwic/edsc_granule```. The path needs ```URL.decode``` so I guess it is better to put it in a separate action so that we are not accidentally 'decode' things that CWIC endpoints can't handle :)

PR #149, opened 2016-03-28
Include Babel Polyfill

8v060htwyc:


bilts:
Thanks for this. The merge conflicts are because the polyfill is in our package.json on master, but I committed the `require` statement to master as well, so this change is now incorporated.

PR #151, opened 2016-03-29
Fix a flaky test that forces user to log in on data retrieval.

mightynimble:


spec/features/data_access/data_access_workflow_spec.rb
```diff
@@ -39,10 +40,13 @@
     end
 
     it "forces the user to login before showing data access page", intermittent: true do
-      screenshot_path = "./tmp/screenshots/debug-#{Time.now.to_i}.png"
-      expect(page).to have_content('EOSDIS Earthdata Login'), lambda {
-        "Expect to see 'EOSDIS Earthdata Login' on the page. #{page.save_screenshot(screenshot_path)}"
-        Rails.logger.info Base64.encode64(File.open(screenshot_path, "rb").read)}
+      unless page.has_content?('EOSDIS Earthdata Login')
```
bilts: What is this doing? In what circumstances is it skipped?

mightynimble: To make sure that the user was logged in ('Manage user account' link will be there in this case) if the redirection finishes before CI looks at the page.

The removed part that base64 encode the screenshot indicates that the user has been redirected back to Earthdata Search already when the test expects the user being on Earthdata Login page.

bilts: This doesn't really make sense. The test suite is incapable of getting past a real Earthdata Login challenge, as far as I'm aware. It has no username or password. Something else must be going on.

Also, the purpose of the test is to assert that the user must log in before data access and that we put a barrier in the way to force just that. Bypassing it when it's not there negates the test.

PR #152, opened 2016-04-01
EDSC-996 store cron job history in db so that /health endpoint can gi…

mightynimble:
…ve consistent results from multiple hosts.

### DIFF
--- a/spec/controllers/health_spec.rb
+++ b/spec/controllers/health_spec.rb
@@ -13,21 +13,30 @@ describe HealthController, type: :controller do
       @retrieval.created_at = Time.now - 1.minute
     end
 
-    before :each do
-      Dir.glob(Rails.root.join('tmp', "data_load_*")).each { |f| File.delete(f) }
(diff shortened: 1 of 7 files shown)


bilts:
This work needs to account for the fact that the colormaps loading code currently needs to run as be run on both hosts, either by fixing our deployment to let it run on both hosts or by updating its data store to not be host-specific.

My preference (longer-term) would be to store colormap info in a tag association, but the CMR tagging API isn't quite there yet, so for now let's try to avoid doing things that are hard to migrate.

spec/controllers/health_spec.rb
```diff
@@ -13,21 +13,30 @@
       @retrieval.created_at = Time.now - 1.minute
     end
 
-    before :each do
-      Dir.glob(Rails.root.join('tmp', "data_load_*")).each { |f| File.delete(f) }
-      Dir.glob(Rails.root.join('tmp', "colormaps_load_*")).each { |f| File.delete(f) }
+    # before :each do
```
bilts: Should remove this entirely.

PR #156, opened 2016-04-07
EDSC-969: Add script to detect whether provider integrated URS and wh…

mightynimble:
…ether user has approved the app in URS using curl. Add capability to authenticate and download data using saved cookies for wget and curl. TODO: Add ASF special case.

### DIFF
--- a/app/views/granules/download.sh.erb
+++ b/app/views/granules/download.sh.erb
@@ -43,7 +43,7 @@ fi
 if command -v curl >/dev/null 2>&1
 then
     STATUS_CODE=`curl -s -z "$(date)" -w %{http_code} <%= raw @urls.first %> | tail -1`
-    if [ "$STATUS_CODE" -ne "200" ]; then
+    if [ "$STATUS_CODE" -ne "200" || "$STATUS_CODE" -ne "304" ]; then
         # Need URS authentication
         if [ ! -f ~/.netrc ]; then
             echo "\n.netrc file not found. Please configure your username and password for authentication using a .netrc file:\n    > echo 'default login <uid> password <password>' > .netrc\nwhere <uid> is your Earthdata Login username and <password> is your Earthdata Login password."
@@ -73,8 +73,10 @@ then
         echo "\n.netrc file not found. Please configure your username and password for authentication using a .netrc file:\n    > echo 'default login <uid> password <password>' > ~/.netrc\nwhere <uid> is your Earthdata Login username and <password> is your Earthdata Login password."
         exit
     fi
-
-    while read -r line; do wget --load-cookies ~/.urs_cookies --save-cookies ~/.urs_cookies --keep-session-cookies $line; done;
+    
+    fetch_urls() {
+        while read -r line; do wget --load-cookies ~/.urs_cookies --save-cookies ~/.urs_cookies --keep-session-cookies $line; done;
+    }
 else
     fail_with_error "Error: Could not find a command-line downloader.  Please install curl or wget"
 fi
--- a/app/views/granules/download.sh.erb
+++ b/app/views/granules/download.sh.erb
@@ -11,8 +11,6 @@ function finish {
 trap finish EXIT
 WGETRC="$wgetrc"
 
-write_credentials
-
 prompt_credentials() {
     echo "Enter your Earthdata Login credentials"
     read -p "Username (<%= @user %>): " username
@@ -36,7 +34,6 @@ exit_bad_login() {
 }
 
 prompt_credentials
-exit
 
 <% if @urls.first.start_with? 'https://datapool.asf.alaska.edu' %>
 <%= render partial: 'asf_download' %>
@@ -46,6 +43,7 @@ exit
 
 fetch_urls() {
     if command -v curl >/dev/null 2>&1; then
+        detect_user_approval
         setup_auth_curl
         while read -r line; do
             curl -b "$cookie_jar" -c "$cookie_jar" -L -n -Og -- $line
--- a/app/views/granules/download.sh.erb
+++ b/app/views/granules/download.sh.erb
@@ -1,16 +1,37 @@
 #!/bin/sh
 
+#TODO: ASF special case
+
 if command -v curl >/dev/null 2>&1
 then
-    fetch_urls() {
-        while read -r line; do curl -Og $line; done;
-    }
+    STATUS_CODE=`curl -s -z "$(date)" -w %{http_code} <%= raw @urls.first %> | tail -1`
+    if [ "$STATUS_CODE" -ne "200" ]; then
+    # Need URS authentication
+    if [ ! -f ~/.netrc ]; then
+        echo "\n.netrc file not found. Please configure your username and password for authentication using a .netrc file:\n    > cd ~\n    > touch .netrc\n    > echo \"machine urs.earthdata.nasa.gov login <uid> password <password>\" > .netrc\n    > chmod 0600 .netrc\nwhere <uid> is your Earthdata Login username and <password> is your Earthdata Login password."
+        exit
+    fi
 
+    APPROVED_STATUS=`curl -s -b ~/.urs_cookies -c ~/.urs_cookies -L --max-redirs 2 -n <%= raw @urls.first %> -w %{http_code} | tail  -1`
+    if [ "$APPROVED_STATUS" -ne "302" ]; then
+        # User didn't approve the app. Direct users to approve the app in URS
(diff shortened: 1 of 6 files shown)


app/views/granules/download.sh.erb
```diff
@@ -1,20 +1,88 @@
 #!/bin/sh
 
+<% if @urls.first.start_with? 'https://datapool.asf.alaska.edu' %>
```
mightynimble: ASF data. This is the special case.

app/views/granules/download.sh.erb
```diff
@@ -1,20 +1,88 @@
...
+set_cookie ()
+{
```
mightynimble: Normally, you can do ```curl -c cookie -b cookie``` with ```-d password&username``` to simulate the session. However, in this case, the generated cookie file is empty. I had to generate the cookie from response manually.

app/views/granules/download.sh.erb
```diff
@@ -1,20 +1,88 @@
...
     }
+elif command -v wget >/dev/null 2>&1
+then
+    if [ ! -f ~/.asf_cookies ]; then
+        echo "\n.asf_cookies file not found. Please configure your username and password for authentication using a .asf_cookies file:\n    > cd ~\n    > wget --keep-session-cookies --save-cookies ~/.asf_cookies --post-data='userid=<uid>&password=<password>' --no-check-certificate https://vertex.daac.asf.alaska.edu/services/authentication\n    > chmod 0600 .asf_cookies\nwhere <uid> is your Earthdata Login username and <password> is your Earthdata Login password."
+        exit
+    fi
 
+    fetch_urls() {
+        while read -r line; do wget --load-cookies ~/.asf_cookies --save-cookies ~/.asf_cookies --keep-session-cookies $line; done;
```
mightynimble: ```wget``` works fine for ASF data.

app/views/granules/download.sh.erb
```diff
@@ -1,20 +1,88 @@
...
+    }
+else
+    fail_with_error "Error: Could not find a command-line downloader.  Please install curl or wget"
+fi
+
+<% else %>
```
mightynimble: This is for the rest of the providers.

app/views/granules/download.sh.erb
```diff
@@ -1,20 +1,88 @@
 #!/bin/sh
```
mightynimble: M - E - S - S - Y !
I know...

app/views/granules/download.sh.erb
```diff
@@ -1,20 +1,88 @@
...
+    }
+else
+    fail_with_error "Error: Could not find a command-line downloader.  Please install curl or wget"
+fi
+
+<% else %>
+
+if command -v curl >/dev/null 2>&1
+then
+    STATUS_CODE=`curl -s -z "$(date)" -w %{http_code} <%= raw @urls.first %> | tail -1`
```
mightynimble: Instead of actually downloading the data, ```curl``` can test it using timestamp.

app/views/granules/download.sh.erb
```diff
@@ -1,20 +1,88 @@
...
+    if [ "$STATUS_CODE" -ne "200" || "$STATUS_CODE" -ne "304" ]; then
```
mightynimble: A poke to the destination of the file, if there is a redirect, it needs go through URS.

app/views/granules/download.sh.erb
```diff
@@ -1,20 +1,88 @@
...
+        # Need URS authentication
+        if [ ! -f ~/.netrc ]; then
+            echo "\n.netrc file not found. Please configure your username and password for authentication using a .netrc file:\n    > echo 'default login <uid> password <password>' > .netrc\nwhere <uid> is your Earthdata Login username and <password> is your Earthdata Login password."
+            exit
+        fi
+
+        APPROVED_STATUS=`curl -s -b ~/.urs_cookies -c ~/.urs_cookies -L --max-redirs 2 -n <%= raw @urls.first %> -w %{http_code} | tail  -1`
```
mightynimble: Furthermore, check whether the app/provider is approved by the user. (Note limit the redirects to 2 here)

app/views/granules/download.sh.erb
```diff
@@ -1,20 +1,88 @@
...
+        if [ "$APPROVED_STATUS" -ne "302" ]; then
```
mightynimble: Different provider returns different status code here. However, if it is approved, a ```302``` will be responded.

app/views/granules/download.sh.erb
```diff
@@ -1,20 +1,88 @@
...
+            # User didn't approve the app. Direct users to approve the app in URS
+            echo "\nERROR: this is your first time retrieving data from this provider using Earthdata Login credentials. You will need to authorize the provider's web service (application) in Earthdata Login GUI before running this script again."
+            exit
+        else
+            fetch_urls() {
+                while read -r line; do curl -b ~/.urs_cookies -c ~/.urs_cookies -L -n -Og $line; done;
+            }
+        fi
+    else
```
mightynimble: Some provider doesn't integrated URS, so the normal route.

app/views/granules/download.sh.erb
```diff
@@ -1,20 +1,88 @@
...
+        fetch_urls() {
+            while read -r line; do curl -Og $line; done;
+        }
+    fi
 elif command -v wget >/dev/null 2>&1
 then
+    # We can't use wget to poke provider server to get info whether or not URS was integrated without download at least one of the files.
+    echo "\nWARNING: Can't find curl, use wget instead, which may affect data retrieval that authenticates through Earthdata Login."
```
mightynimble: See comment in the script. ```wget``` has to download the file once to compare the size and timestamp of the local file against the remote file.

app/views/granules/download.sh.erb
```diff
@@ -1,20 +1,88 @@
...
+    if [ ! -f ~/.netrc ]; then
+        echo "\n.netrc file not found. Please configure your username and password for authentication using a .netrc file:\n    > echo 'default login <uid> password <password>' > ~/.netrc\nwhere <uid> is your Earthdata Login username and <password> is your Earthdata Login password."
+        exit
+    fi
+    
     fetch_urls() {
-        while read -r line; do wget $line; done;
+        while read -r line; do wget --load-cookies ~/.urs_cookies --save-cookies ~/.urs_cookies --keep-session-cookies $line; done;
```
mightynimble: However, if user approves the app in URS profile, the user can download the data without a problem.

app/views/granules/download.sh.erb
bilts: These should be comments in the source code rather than in the pull request.

bilts:
ASF silently fails / breaks if you haven't authorized the vertex app.

bilts:
FYI I've been working on cleaning this up to make it more automatic and not require setting default credentials in netrc (which would pass your URS credentials to every server you ever wget or curl).

bilts:
I just pushed out a big cleanup to the shell script. I had tested it as working against ASF, but I've refactored since then. It's about as safe as I can make it.

I don't think it'll work for FTP. I'm very hesitant to write a script that's going to send URS credentials in the clear.

mightynimble:
I took a look at the latest WIP commit. There are a couple of issues I wanna point out: (I know you are still refactoring the script, so I commented out ```write_credentials``` and the ```exit``` below ```prompt_credentials``` to make the script run.)

1. For ASF data, the cookie has to be manually generated see ```set_cookie``` function. Otherwise it will redirects to URS and URS will redirect it back to vertex; it causes an infinite loop and will finally error out.
2. For other DAACs data which requires URS auth but user didn't approve the app, there is a way to detect it by following redirections using ```--max-redir 2``` option (see https://github.com/nasa/earthdata-search/pull/156/commits/5f68bfc490fe6cebf2e7d4ecf77a02fe492dd776 line#15)

I can take over on Monday and continue working on the refactoring since you are out.

app/views/granules/download.sh.erb
```diff
@@ -1,19 +1,78 @@
...
+    echo
+}
+
+exit_bad_login() {
+    echo
+    echo "Unable to log in"
+    echo
+    echo $1
+    echo
+    echo "<%=raw @urls.first %>"
```
bilts: This looks like WIP code

mightynimble: ```$1``` is the passed-in exit reason. We show user two error messages on exiting. Is this what you are referring to?

PR #157, opened 2016-04-12
EDSC-1000 CSS change to prevent text on map in detail view.

mightynimble:


app/assets/stylesheets/modules/media_queries.css.scss
```diff
@@ -19,7 +19,7 @@
 	}
 }
 
-@media (max-width: 1349px) {
+@media (max-width: 1549px) {
```
bilts: This approach seems fragile and likely to break again. Can you find something more robust?

Maybe use flexbox with https://github.com/10up/flexibility ?

bilts:
Needs polyfill for IE < 11 as mentioned in my comment.

mightynimble:
Tested in IE9 and IE11. The page looks ok without the polyfill setting. With ```-js-display```, the spatial/temporal part hides behind the mini map when the width of the detail panel shrinks..

app/views/layouts/application.html.erb
```diff
@@ -30,6 +30,7 @@
     <% end %>
     <%= yield :javascript %>
     <%= javascript_include_tag "jquery.simulate" if Rails.env.test? %>
+    <!--[if lt IE 10]> <%= javascript_include_tag "flexibility" %> <![endif]-->
```
bilts: Does IE 10 work?

app/assets/stylesheets/modules/master_overlay.css.scss
```diff
@@ -418,13 +418,14 @@ $parent-width: $parent-split * 0.21;
 }
 
 .row {
+  -js-display: flex;
```
bilts: I thought you had said -js-display messed up IE 11. Is it only if flexibility is loaded?

app/views/layouts/application.html.erb
mightynimble: I don't have IE10 to test. But Flex is supported by IE10 by default. So I think it should behave like IE11.

app/assets/stylesheets/modules/master_overlay.css.scss
mightynimble: Besides -js-display, the fixed-width mini map also contributes to the issue in IE11. Therefore, I made changes to .flex-column class as well.

app/views/layouts/application.html.erb
bilts: I thought IE 10 required a vendor prefix, `-ms-`

app/assets/stylesheets/modules/master_overlay.css.scss
bilts: So, does this line still cause problems in IE 11?

app/views/layouts/application.html.erb
mightynimble: I am downloading an IE10 VM. I will test it when it finishes, with and without -ms- prefix.

app/assets/stylesheets/modules/master_overlay.css.scss
mightynimble: No. IE11 is fine now.

mightynimble:
So, another commit. This fixes layouts on REAL IE boxes. ( Although it looks alright in VM IE9-11, the two-column detail page displays as a single column on physical Windows machines:) )

app/assets/stylesheets/modules/media_queries.css.scss
bilts: Are these media changes still needed? Why?

mightynimble: I have no idea how they get in.. I will remove it.

PR #162, opened 2016-04-20
EDSC-994 Show additional attribute when the collection is tagged with…

mightynimble:
… .features.attribute_search


app/views/search/_granule_search.html.erb
```diff
@@ -78,7 +78,7 @@
           <span data-bind="visible: !cmrGranuleQuery.validateCloudCoverRange(cmrGranuleQuery.cloudCover.min(),cmrGranuleQuery.cloudCover.max())" class="error">Minimum must be less than Maximum</span>
         </p>
       </dd>
-      <!-- ko if: $root.labs() && searchable_attributes -->
+      <!-- ko if: $root.labs() && searchable_attributes || tags() && tags()['<%=Rails.configuration.cmr_tag_namespace.gsub(/\.$/, '')%>.features.attribute_search'] && tags()['<%=Rails.configuration.cmr_tag_namespace.gsub(/\.$/, '')%>.features.attribute_search'].data -->
```
bilts: cmr_tag_namespace doesn't have a trailing dot. You should be able to safely drop the .gsub.

bilts: Can you put a method on the collection along the lines of "has_feature('attribute_search')" rather than doing this in the view?

bilts: The &&'s and ||'s are messed up here. Should not show Collection-Specific attributes if there are no searchable_attributes no matter what is contained in labs or tags. Right now, if the collection is tagged, the header shows up regardless of whether there are actually any attributes to show.

mightynimble: I will edit the PR applying the later two comments. About the first one: Doug sees tags like
```            {"concept-id": "T1230649091-CMR",
            "revision-id": 1,
            "tag-key": "edsc..extra.browseable_granule",
            "description": null,
            "originator-id": "edsc.sys"}```
The pre-configured value ```edsc``` doesn't have a trailing dot though.

File: db/seeds.rb [hunk 2]
```diff
@@ -204,6 +204,7 @@
...
+  {echo_id: "C1000000320-LPDAAC_ECS", has_browseable_granules: false, has_granules: true, browseable_granule: nil, granule: nil, searchable_attributes: {"attributes"=>[{"name"=>"CorrectionAchieved", "data_type"=>"STRING", "description"=>"The identifier to inform the user of the level of correction achieved for the product. (The choices are: Terrain+Precision, Terrain+Systematic, Systematic, Precision)"}]}, orbit: {}},
```
bilts: This is not the full list of attributes, and it'll be overwritten by the full list of attributes in a background job (except that there's a bug that I just discovered and am now fixing that has been preventing that background job from working, push coming soon).

mightynimble: I add the line there mainly to make the new test pass.

bilts: Yes, but when it gets overwritten, there's a big usability problem.

app/views/search/_granule_search.html.erb
```diff
@@ -78,7 +78,7 @@
           <span data-bind="visible: !cmrGranuleQuery.validateCloudCoverRange(cmrGranuleQuery.cloudCover.min(),cmrGranuleQuery.cloudCover.max())" class="error">Minimum must be less than Maximum</span>
         </p>
       </dd>
-      <!-- ko if: $root.labs() && searchable_attributes -->
+      <!-- ko if: ($root.labs() || has_feature('<%=Rails.configuration.cmr_tag_namespace.gsub(/\.$/, '')%>', 'attribute_search')) -->
```
bilts: The && and || are still wrong. Now if you have the labs parameter set, the heading will show up even if there are no searchable attributes.

Also, Can you avoid passing the cmr_tag_namespace in erb and just get it from Javascript?  See note I'll be making above.

app/assets/javascripts/models/data/collection.js.coffee
```diff
@@ -299,4 +299,7 @@ ns.Collection = do (ko
       this[prop] ?= ko.observable(undefined)
       this[prop](jsonObj[prop] ? this[prop]())
 
+    has_feature: (namespace, feature_key) ->
+      return @tags()?[namespace + ".features." + feature_key]?.data && @searchable_attributes
```
bilts: This should use the existing method to read tags, which lets us be more robust against future tag API changes.

Also "@searchable_attributes" should not be in this method. What if we add another feature that has nothing to do with additional attributes?

So,
    
    has_feature: (key) ->
      @getValueForTag("features.#{key}")

db/seeds.rb
bilts: Please replace with the full list of attributes.

PR #165, opened 2016-04-22
EDSC-984: Set default model data only when the form is up to date.

mightynimble:


app/controllers/data_access_controller.rb
```diff
@@ -192,6 +197,17 @@ def options
 
   private
 
+  def echo_form_outdated?(access_config, methods)
+    return true if access_config.nil? || access_config.echoform_digest.nil? || methods.nil?
```
bilts: What if there is no echoform, i.e. if the user had previously selected "Download", one of the OPeNDAP methods, or an order without a form?

app/assets/javascripts/models/data/project.js.coffee
```diff
@@ -87,9 +87,12 @@ ns.Project = do (ko,
       options = @serviceOptions.serialize()
       $(document).trigger('dataaccessevent', [@collection.id, options])
 
+      form_data = []
+      form_data.push method.form for method in @granuleAccessOptions().methods
       id: @collection.id
       params: param(@collection.granuleDatasource()?.toQueryParams() ? @collection.query.globalParams())
       serviceOptions: options
+      form: form_data
```
bilts: This addition seems to be adding all of the forms, in their entirety to the saved data in a project, which is potentially quite large (AST_L1A). Can we pare it down to just the digest. And, really, just the selected digest?

bilts: Also, I don't 100% trust that the form is clean and unmodified at this point.

app/controllers/data_access_controller.rb
mightynimble: It will consider the (non-exist) echoform out dated and won't try to load from AccessConfigure.

app/models/access_configuration.rb
```diff
@@ -1,19 +1,17 @@
...
 
-  def self.get_default_options(user, cmr_concept_id)
-    access_config = self.find_by(user: user, dataset_id: cmr_concept_id)
-    if access_config
-      access_config.service_options
-    else
-      nil
-    end
+  def self.get_default_access_config(user, cmr_concept_id)
+    self.find_by(user: user, dataset_id: cmr_concept_id)
```
bilts: Why was the if-statement there before, and why was it removed?

mightynimble: the ```get_default_options``` used to return ```service_options```, not the whole ```access_config```. So the if-else is a nil check. It is removed because now the method returns the whole ```access_config```.

spec/features/data_access/access_option_defaults_spec.rb
```diff
@@ -39,16 +39,15 @@
...
+        'accessMethod' => [{'method' => 'stale', 'model' => '<broken/>', 'rawModel' => '<broken/>', 'type' => 'order'}]
       }
-      AccessConfiguration.set_default_options(user, collection_id, options)
+      AccessConfiguration.set_default_access_config(user, collection_id, options, ['form data changed'])
       load_page 'data/configure', project: [collection_id]
     end
 
     it "presents default options for the order" do
       expect(page).to have_unchecked_field('Download')
-      expect(page).to have_checked_field('FtpPushPull')
```
bilts: I never noticed this line before, but it should have never been in the baseline. Ugh.

Also, I think there needs to be another test here. This spec tests that an order form that doesn't exist doesn't cause problems. We need another one where an order option that exists but has changed doesn't cause problems.

app/assets/javascripts/models/data/project.js.coffee
mightynimble: I agree that sending hash to rails backend is the cleaner way! On the other hand, sha-1 will be introduced in addition to murmur3. I used murmur3 to hash the excluded cwic granule ids since it is shorter than sha-1 and the collision rate is acceptable. I used sha-1 in rails because it is just an easy grab from the digest lib. Should I switch to use murmur3 in rails as well?

I like the idea of hashing the whole form more than just the selected (easier :))..

Is there a way to guarantee that the form is not dirty?

spec/features/data_access/access_option_defaults_spec.rb
mightynimble: Will do.

app/assets/javascripts/models/data/project.js.coffee
bilts: No need to do any of that. Have the server tell the client what the sha1 is, then have the client pass the sha1 back to the server. The client doesn't need to compute the sha1.

app/controllers/data_access_controller.rb
bilts: It will set the defaults to `{service_options: nil}` which means that the 'Download' radio won't be checked, right? Or am I misreading that?

mightynimble: Right, the download radio won't be checked in this case.

app/controllers/data_access_controller.rb
```diff
@@ -192,6 +197,26 @@ def options
...
+  # The digest in access_config may be dirty.
+  def echo_form_outdated?(access_config, methods)
+    return true if access_config.nil? || access_config.echoform_digest.nil? || methods.nil?
+
+    form_digest = []
+    methods.each do |method|
+      if method[:type] == 'download'
+        digest = access_config.echoform_digest.select {|digest| digest['id'] == method[:type]}
+      else
+        return true if method[:form].nil?
```
bilts: This isn't right, is it? If there is no form for an order (which I think is possible) it needs to be selected as default.

I think this method is potentially overcomplicated. Check if the form has changed (to include the addition / removal of a form). If it has, return false, otherwise true.

PR #197, opened 2016-05-13
EDSC-1054 Add error message when un-supported file is uploaded.

mightynimble:


Gemfile.lock
```diff
@@ -299,6 +299,3 @@ DEPENDENCIES
   unicorn
   vcr
   whenever
-
-BUNDLED WITH
```
bilts: What happened here?

bilts:
This is good, but I think we should additionally have some one-off instructions specifically for .shp.

Gemfile.lock
mightynimble: After v1.10, bundler adds it automatically. I was using an older version which removes it on ```bundle install```

bilts: No, I was just curious.

bilts:
When I hover to remove the file, the error text shows up in both the tooltip and inline. Can we make it so the error text only shows up inline?

PR #199, opened 2016-05-16
EDSC-1006: Fix a CWIC issue where non-zero count shows up when granul…

mightynimble:
…es are filtered out for some CWIC collections.


bilts:
I can no longer reproduce this problem in UAT. Can you?

mightynimble:
I can still reproduce this.Firstly, click on the link in the ticket. Then open the granule filter panel and clear the temporal, click apply. Finally, add the temporal back, you will see ```showing 15 of 0 granules```.

PR #203, opened 2016-05-23
EDSC-1010: Maintain facet panel open/close state when returning to co…

mightynimble:
…llection list from project-overview or granule-list.

### DIFF
--- a/app/assets/javascripts/models/ui/state_manager.coffee
+++ b/app/assets/javascripts/models/ui/state_manager.coffee
@@ -131,6 +131,7 @@
     _readPath: ->
       state = @overlayState()
       path = @_pathForState(state)
+      @overlayState(state)
       @_onPathChange(path)
 
       path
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/models/ui/state_manager.coffee
```diff
@@ -142,15 +143,23 @@
 
       root = '/search'
 
-      return root if state.parent && state.current != 'collection-details'
+      if state.parent && state.current != 'collection-details'
+        if state.current == 'granule-list' || state.current == 'project-overview' || state.current == 'granule-details'
```
bilts: I think I'd prefer to see all of these if-statements (lines 147, 155, 160) combined into a single if-statement on line 144 that says something like:

    state.parent = false unless state.current == 'collection-results'

app/assets/javascripts/models/ui/state_manager.coffee
```diff
@@ -177,7 +187,10 @@
       component = components.shift()
 
       state.visible = component != 'map'
-      state.parent = !component
+      if component == 'collections'
+        state.parent = false
+      else
+        state.parent = state.manualShowParent
```
bilts: Shouldn't these 4 lines be something like:

    state.parent = !component && state.manualShowParent

app/assets/javascripts/models/ui/state_manager.coffee
```diff
@@ -177,7 +187,10 @@
       component = components.shift()
 
       state.visible = component != 'map'
-      state.parent = !component
+      if component == 'collections'
```
bilts: Thinking about this, isn't this basically the only necessary trigger, with nothing necessary on the master overlay or in the template? If you came from /search/collections, the facets should be collapsed when you go back, otherwise they should be expanded. That also feels more correct / future-proof in case we use the parent for other things in the future.

spec/helpers/overlay_helpers.rb
```diff
@@ -146,5 +146,10 @@ def nth_panel(root, n)
     def nth_ccol(root, n)
       root.find(".ccol:nth-child(#{n})")
     end
+
+    def manual_close_facet_list
+      page.find('#master-overlay-parent a.master-overlay-close').click
+      sleep 0.5 # wait for animation
```
bilts: I don't like sleeps in our tests. Use synchronize instead so you wait exactly as long as needed. Also, I'm not sure there are animation in test mode (I can't remember if they're shut off for CSS or just in JS)

mightynimble:
Cleaned up the state manager (basically reverted all the changes made to ```_pathForState```.)
Set ```state.parent``` in

PR #209, opened 2016-06-06
EDSC-384 EDSC-770 Manually enter point and bounding box 

mightynimble:


app/views/search/index.html.erb
```diff
@@ -261,14 +261,16 @@
...
+
 <div class="landing-area landing-visible">
 
   <div class="landing-dialog">
     <div class="noscript no-js">
       <h3><%= site_name %> uses JavaScript</h3>
       <ul>
         <li>Please use a JavaScript-aware browser.
-        <li>Make sure that JavaScript is enabled.
+        <li>Make sure that JavaScript is <enabled class=""></enabled>
```
bilts: Is this a mistake?

app/assets/stylesheets/modules/toolbar.css.scss
```diff
@@ -15,8 +15,31 @@
...
+
+  input {
+    min-width: 200px;
+    padding: 0;
+  }
+}
+
 .leaflet-top.leaflet-right {
-  top: 50px;
+  top: 145px;
```
bilts: Why was this 50 and what changed it?

app/assets/stylesheets/modules/toolbar.css.scss
```diff
@@ -15,8 +15,31 @@
   margin: 0;
 }
 
+.manual-coord-entry {
```
bilts: I don't so much like the number of one-off styles required for this feature. Could some of it be reusable?

app/assets/javascripts/modules/map/spatial_selection.js.coffee
```diff
@@ -148,6 +150,8 @@ ns.SpatialSelection = do (window,
         @_layer = @_oldLayer
         @_oldLayer = null
         @_drawnItems.addLayer(@_layer)
+        currentPage.query.spatialCondition.visible(true)
```
bilts: "visible" is a presentation concern, but the spatial condition is part of a data model, not a UI model. The condition shouldn't care whether or not its UI is visible; that should be tracked elsewhere.

app/assets/javascripts/models/data/spatial_condition.js.coffee
```diff
@@ -0,0 +1,72 @@
...
+            for range in ranges
+              match = range.trim().match(/^-?\d+(\.\d+)?$/)
+              unless match?
+                return "Invalid coordinate: #{range}"
+              [all, min, max] = match
+              min = parseInt(min, 10)
+              max = parseInt(max, 10)
+        null
+
+      @queryCoordinates = @computed =>
```
bilts: The @computed should be a separate method on the class (ditto to line 15 above)

app/assets/javascripts/models/data/spatial_condition.js.coffee
```diff
@@ -0,0 +1,72 @@
...
+        if coordinates? && coordinates.length > 0
+          coords = coordinates.split(/:/)
+          for coord in coords
+            ranges = coord.split(',')
+            return "Coordinate must be two comma-separated numbers: #{coord}" unless ranges.length == 2
+            for range in ranges
+              match = range.trim().match(/^-?\d+(\.\d+)?$/)
+              unless match?
+                return "Invalid coordinate: #{range}"
+              [all, min, max] = match
```
bilts: These three lines don't appear to do anything for the error message

app/assets/javascripts/models/data/spatial_condition.js.coffee
```diff
@@ -0,0 +1,72 @@
...
+        null
+
+      @queryCoordinates = @computed =>
+        @visible(spatial().length > 0)
+        type = spatial().split(':')[0]
+        newSpatial = ''
+
+        return null unless !!@previousType || !!spatial()
+
+        if type == @previousType
```
bilts: I don't quite understand the keeping and testing against previousType and previousSpatial. This is a computed property; you should be able to just re-computed it without regard for its current value.

app/assets/javascripts/models/data/spatial_condition.js.coffee
```diff
@@ -0,0 +1,72 @@
...
+
+        if newSpatial?.length > 0
+          value = newSpatial
+          newSpatial = "#{type}:#{value}"
+          if type == 'point'
+            @coordinates(value)
+          else
+            @swPoint(value.split(':')[0])
+            @nePoint(value.split(':')[1])
+            @coordinates(@swPoint() + ":" + @nePoint())
```
bilts: This ko.computed method is a getter. Generally you shouldn't be setting values in a getter, only reading them. The exception is if you're trying to use the ko.computed to periodically perform some external action like fire off an XHR when something changes. Those are rare and tend to make things complicated, though.

vendor/assets/javascripts/leaflet.draw-0.2.2/leaflet.draw-src.js
```diff
@@ -55,7 +55,7 @@ L.drawLocal = {
 			},
 			rectangle: {
 				tooltip: {
-					start: 'Click and drag to draw rectangle.'
+					start: 'Click and drag to draw rectangle. test test.'
```
bilts: Remove

app/views/search/_spatial_manual_entry.html.erb
```diff
@@ -0,0 +1,16 @@
+<div id="manual-coord-entry-container" class="manual-coord-entry" data-bind="with: query.spatialCondition">
+  <p data-bind="visible: !visible()"><i class="fa fa-fw fa-crop"></i></p>
+  <p data-bind="visible: visible()">
+    <label for="manual-coord-entry"><!-- ko text: spatialType --><!-- /ko -->: </label>
+    <!-- ko if: spatialType() == 'Point' -->
+    <input type="text" id="manual-coord-entry" name="manual-coord-entry" data-bind="value: coordinates, valueUpdate: 'blur'" placeholder="lon, lat (e.g. 130, 44.2)">
```
bilts: Can we do lat/lon instead? That seems to be how everyone else in the world thinks.

bilts: It seems like it could simplify things a lot if you had separate inputs for lat and lon as well as separate inputs for south, west, north, and east in the bounding box case.

PR #210, opened 2016-06-07
EDSC-1071 Correct keyword search param to 'free_text' and regenerate …

mightynimble:
…fixtures.


mightynimble:
The failures in build #909-#914 are due to CWIC granule temporal changes. It happened in master as well. I re-generated those fixtures and changed test's expectations. (The temporal now doesn't have ```Thh:mm:ssZ``` any more.)

PR #214, opened 2016-06-14
EDSC-770 and EDSC-384: Manual spatial entry (point and rectangle)

mightynimble:


bilts:
I'm noticing lots of bugs and glitches. The first couple are visible, the last few seem to relate to state leaks.

1. When I'm not selecting anything, there's a big gap between the spatial box and the layer selector
2. When I've selecting a bounding box, the entry box draws above the layer selector
3. When I'm selecting a polygon, the text says "polygon:" but then when I finish selecting the polygon, it shows the same thing as when I have no spatial filter. I think this only happens when I selected a bounding box and then a polygon.
4. The error message causes the entry box to expand horizontally in an ugly way
5. The coordinates are not clipped to +/- 90 lat wrapped across +/- 180 lon, even though the query / bookmark is
6. If I've entered a coordinate in the box and I try to manually clear it, instead of having my point go away, the filter stays in place and I get an error message
7. If I hit "Clear Filters" when there's an error message up, the entry text fields go away and the filter gets cleared, but the error message stays visible
8. (Probably the same source as the previous item) If I have an error visible and I manually select a new value by dropping a point on the map, the text box gets updated with the point I selected but the error remains visible
8. When entering a bounding box, if there's an error, as soon as both fields are populated, the pillbox gets replaced by whatever the current spatial is (e.g. it disappears if you had no filter before the error) though the error message remains visible

mightynimble:
The one failed spec left was caused by some weird CWIC fixture issue. Somehow CI uses a different client_id for CWIC requests.

app/assets/javascripts/models/ui/spatial_type.js.coffee
```diff
@@ -7,25 +7,76 @@ ns.SpatialType = do (ko, $=jQuery) ->
...
+      if spatialParam then spatialParam else @name()
+
+    _clearDisplaySpatialName: (value) =>
+      console.log "---- #{value}"
+      null
+
+    _getQuerySpatial: ->
+      @query.spatial()?.split(':')[0]
+
+    _setMapControlPosition: ->
```
bilts: This is a model class. It shouldn't know anything about the UI, especially about specific pixel values.

app/assets/javascripts/models/ui/spatial_type.js.coffee
```diff
@@ -7,25 +7,76 @@ ns.SpatialType = do (ko, $=jQuery) ->
...
+      null
+
+    _getDisplaySpatialName: =>
+      spatialParam = @_toReadableName(@query.spatial()?.split(':')[0])
+      if @name() == 'Spatial'
+        # Order matters
+        if spatialParam && @name() != spatialParam && @displaySpatial() != spatialParam && @name() != @displaySpatial()
+          # on spatial type changes from one (e.g. point) to another (e.g. rectangle).
+          return @displaySpatial()
+#        else if spatialParam == null && @name() =='Spatial' && (@displaySpatial() == 'Point' || @displaySpatial() == 'Rectangle')
```
bilts: Clean up unnecessary comments

app/assets/javascripts/models/ui/spatial_type.js.coffee
```diff
@@ -7,25 +7,76 @@ ns.SpatialType = do (ko, $=jQuery) ->
     constructor: (@query) ->
       @icon = ko.observable('fa-crop')
       @name = ko.observable('Spatial')
+      @displaySpatial = ko.computed(read: @_getDisplaySpatialName, write: @_clearDisplaySpatialName, owner: this, deferEvaluation: true)
+      @querySpatial = ko.computed(read: @_getQuerySpatial, owner: this, deferEvaluation: true)
+      @mapControlTop = ko.computed(read: @_setMapControlPosition, owner: this)
```
bilts: By convention, computed read methods start with `_compute` elsewhere in the code. Here you have three read methods, two of which start with `_get` and one of which starts with `_set`, which I think is confusing.

app/assets/javascripts/models/ui/spatial_type.js.coffee
```diff
@@ -7,25 +7,76 @@ ns.SpatialType = do (ko, $=jQuery) ->
...
+#          return @name()
+        if spatialParam?.length > 0
+          @manualEntryVisible(true)
+          return spatialParam
+        return @displaySpatial() if @displaySpatial()
+#        return '' if spatialParam == null
+#      if spatialParam != @name() then return @name() else return @name()
+      if spatialParam then spatialParam else @name()
+
+    _clearDisplaySpatialName: (value) =>
```
bilts: If I call a writer on a compute property, I expect the backend values to be set. This doesn't do anything. It looks like this is just a debug method?

app/views/search/_spatial_manual_entry.html.erb
```diff
@@ -0,0 +1,19 @@
+<div id="manual-coord-entry-container" class="manual-coord-entry">
+  <p data-bind="visible: !ui.spatialType.manualEntryVisible() && !ui.spatialType.querySpatial()"><i class="fa fa-fw fa-crop"></i></p>
+  <p data-bind="visible: ui.spatialType.manualEntryVisible() || ui.spatialType.querySpatial()">
+    <label for="manual-coord-entry-point"><!-- ko text: ui.spatialType.displaySpatial() + ':' --><!-- /ko --></label>
+    <!-- ko if: ui.spatialType.displaySpatial() == 'Point' -->
+    <input type="text" id="manual-coord-entry-point" name="manual-coord-entry-point" data-bind="value: spatialEntry.point, valueUpdate: 'blur'" placeholder="lat, lon (e.g. 44.2, 130)">
+    <!--/ko-->
+    <p>
+    <!-- ko if: ui.spatialType.displaySpatial() == 'Rectangle' && ui.spatialType.manualEntryVisible() -->
+    <label for="manual-coord-entry-swpoint">Southwest Point</label>
```
bilts: Just a quick trick to avoid having to set `id` and `for` attributes on labels and inputs, this is valid HTML that does the same thing:

    <label>Southwest Point <input type="text" data-bind="value: spatialEntry.swPoint...."></label>

bilts:
Are you able to reduce the gap between the coordinate entry and the rest of the toolbar when no errors are visible?

I think it's okay to have errors make them overlap, but with browse imagery, vertical real estate over there is at a premium.

bilts:
If I have a point, then click on the bounding box icon, but then click cancel, my point spatial re-appears but the bounding box manual entry remains visible (and is cleared).

bilts:
If I have a polygon selected, the spatial window shows "Polygon:" with no points.

PR #216, opened 2016-06-17
EDSC-1070: Demonstrate GIBS granules

bilts:


bilts:
To test: search for AST_L1T and view granules from 2016-02-17T17:__:__Z. They should have imagery.

PR #221, opened 2016-06-24
EDSC-1075: Add ajax calls to pull additional pages directly from CMR.

mightynimble:


app/controllers/granules_controller.rb
```diff
@@ -95,6 +95,63 @@ def each
     end
   end
 
+  def fetch_links
```
mightynimble: This action is quite similar to ```download```. I had to steal a lot from ```download``` and ```GranuleUrlStreamer#each```. This ticket only addresses ```/granules/download.HTML```. ```/granules/download.sh``` still has the timeout issue if there are too many links to retrieve from CMR.

```EDSC-1124``` was created to address the timeout error caused by script file download. Once it is resolved,  the ```#download``` action will be much slimmer than it is now since most of the logic will be commonly used to generate the HTML page and the shell script file.

File: app/views/granules/download.html.erb [hunk 2]
```diff
@@ -1,19 +1,90 @@
...
-    <ul>
-      <% @urls.each do |url| %>
-      <li><a href="<%= url %>"><%= url %></a></li>
-      <% end %>
-    </ul>
-    <% if  @errors.present? %>
-      <%= @errors %>
-     <% end %>
-  </body>
+<head>
```
mightynimble: Steals @bilts ```cwic_granule``` code. However, to find all download links, we can't query CMR directly but have to go through our rails app to parse links for opendap granules.

app/controllers/granules_controller.rb
bilts: Make sure EDSC-1124 gets prioritized for next sprint.

app/views/granules/download.html.erb
```diff
@@ -1,19 +1,90 @@
...
+  (function () {
+
+    var browseOnly = <%= @query.delete('browse') %>;
+    var currentPage = <%= @query['page_num'] %>;
+    var project = <%= @query['project'] %>;
+    var collection = '<%= @query['collection'] %>';
+    var cmrHits = -1;
+
+    function getNextPageUrl() {
+      queryPageNum = parseInt(url.match(/page_num=(\d+)/)[1]);
```
bilts: I think CMR may be returning a link rel="next" element now. Check and use that if so.

bilts:
+1 on code (please address minor comments)

Need someone to review functionality, though

macrouch:
My only comment was about testing, I didn't know if we needed any more tests for the new way the download links page works.

Functionality looks good

mreese84:
@mightynimble This looks good to me.

charleshuang80:
seems fine to me as well

app/views/granules/download.html.erb
mightynimble: Checked in SIT and there is no such field yet.

PR #222, opened 2016-06-29
EDSC-1108 Integrate facet-v2 (humanize facets)

mightynimble:
For you to preview...


mreese84:
@mightynimble can you please comment/let me know when this is ready for "functional" review. Meaning when you're "done" with code. Just want to make sure we aren't each waiting on each other. Thanks man.

mightynimble:
@mreese84 It is not ready for functional review yet. I am still working on science_keywords facets. Then I will move on to query param compression and adding 'features' facets etc.

bilts:
I would expect this change to result in the net deletion of a ton of code surrounding facets, parsing, hard-coding, etc. The fact that it's almost 50/50 adds/removals right now has me a little wary, though I can see that the classes dealing with facet responses haven't been removed yet.

Make sure you get rid of the the facet pruning code in the fixtures, the facet presenters, etc. Try to assume as little as possible about the parameters facets are giving you (hint: they're all about to change with the work I just did) and just trust what the CMR gives you.

@mreese84 We're leaving this on a branch, right? IIRC Katie wanted a way to see them side-by-side.

mightynimble:
@bilts yes, I commented out the facet presenters in the collections#index and accepted whatever CMR is sending back to me. The majority of the changes happen on the coffeescript side, like in collection_facet.js. Basically some translation to convert CMR facet response into EDSC query params and to construct new CMR facet-v2 queries when a facet is clicked. And thank you for pointing out to remove facet pruning code - I will never think of that!

I am a little worried about the upcoming changes to the facets parameters. How they gonna change? Are you talking about the query params, for example the params in the CMR request below, is going to change?

https://cmr.sit.earthdata.nasa.gov/search/collections.json?include_facets=v2&science_keywords%5B0%5D%5Bcategory%5D=EARTH+SCIENCE+SERVICES&page_size=0

bilts:
That's exactly what I mean. It's going to be science_keywords_h for that parameter.

In general, the whole point of facets v2 is that you should not ever care about the specifics of what parameters or facets are returned. Look at the guarantees made by the media type and stick to that. If CMR decides tomorrow that we should really facet on shoe size, we should handle that without any modifications (in other words, we should be completely domain agnostic here). Just use the add and remove links provided and everything will be happy. In fact, for most of the facets, CMR will give you back a couple of extra parameters like options[platforms][and]=true, which you'd miss if you just parsed out the param you're expecting.

mreese84:
@bilts thanks for the feedback. And yes, we are going to leave this in its own branch. There were a few people concerned with going all-in, so we'll deploy this branch and see how things work first.

mightynimble:
@bilts One thing I don't quite understand - since I have to present the selected facets in the EDSC query, I have to keep in mind what CMR facet query params are so that I can do the conversion/forwarding. How can I do that if I don't track what CMR facet query params are?

If CMR adds some new facets, (say shoe size :)), I can list the new param in the list to accommodate that. Just can't think of a way to be 100% domain agnostic...

bilts:
Everything that CMR gives you that is not already tracked elsewhere
(spatial, temporal, keywords) must be facet-related, right?

mightynimble:
Why these params are facet-related?

No matter whether they are facet-related or not, we need to know where to
look for spatial/temporal/keywords params in CMR query and we have to do
query name translation to construct CMR requests. For example, CMR uses
'keyword' for keyword search. If they decide to change that to 'free_text'
someday, we will have to make the change when constructing CMR queries,
right?

mightynimble:
Filed https://bugs.earthdata.nasa.gov/browse/CMR-3190.

mreese84:
When a facet item that you have to scroll down to see is selected, the facet list is updated based upon the selection with the selected option at the top. However, since you had to scroll down and no longer see the top item, it's very easy to lose context as a user. It would be preferable if when you select an item after scrolling down, the list reset itself so that the selected option is visible at the top of the facet pane. Also, it would be preferable if that transition were to be animated (the scroll) so that it's not a snap/jerky movement. That's not good UX either.

So, in an ideal world it would be select item > updated list > scroll the list so the selected item is at the top of the facet list. Allows user to maintain context and confirm what they clicked on was what they thought they clicked on.

mreese84:
Admittedly, I don't know enough about old facets vs new facets to test edge cases. Other than the comment I provided earlier (which may or may not be out of scope for this ticket), it all seems to work.

mightynimble:
@mreese84 the scroll thing is actually a spot on - I had this exact experience once and a similar issue was filed and fixed (EDSC-1017) coming out of the LPDAAC user study. I tried to steal code from Patrick that he made in the other ticket but scrolling on facet list seems slightly complicated than scrolling the collection list (or maybe it is really easy to do but I am just being dumb right now). I will keep it in mind and think about it.

macrouch:
I selected Biosphere, then Aquatic Ecosystems, then Lakes, and I repeating facets

screen shot 2016-07-13 at 10 10 55 am

macrouch:
The repeating seems to only happen with Lakes, I tried a lot of other facets and that was the only time I saw it.

I also tried removing a middle level facet, like Aquatic Ecosystems in the above example, and it removes all applied facets. I was expecting parent facets (Biosphere) to still be applied.

mightynimble:
I think the two are related. It is a CMR issue (CMR-3190) that I thought I had fixed it in commit ecbf229. I will take a look. Thanks.

mightynimble:
Filed CMR-3207 for the issue above. However, I did find a minor issue when I looked at the code: the ```_findAncestors``` recursive can be more efficient if the ```child``` is passed in as an argument.

mreese84:
This looks good to me. On to the testing phase!

PR #224, opened 2016-07-01
Gem updates

mreese84:
Updating gem version numbers to address setup issues.


mightynimble:
:+1: as long as CI builds.

PR #226, opened 2016-07-06
EDSC-1116: Fix granule list action item positioning.

mreese84:
Slight refactor of action item positioning on collection list so that granule action items behave the same way.

### DIFF
--- a/app/assets/stylesheets/modules/panel_list.css.scss
+++ b/app/assets/stylesheets/modules/panel_list.css.scss
@@ -236,17 +236,13 @@ $panel-list-thumbnail-container-size: $panel-list-thumbnail-size + 10px;
   position: relative;
   top: 2px;
   left: 2px;
+  float: right;
 }
 
 .panel-list-item:after {
     content: "";
     display: block;
     clear: right;
-}
-
-.panel-list-item-actions-container {
-  display: block;
-  float: right;
 }
 
 #granule-list:focus, #granule-list.has-focus {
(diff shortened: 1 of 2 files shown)


mightynimble:
The icons on granules list page is slightly off from what I see in OPS. (https://search.earthdata.nasa.gov/search/granules?p=C1237207600-LARC_ASDC)
They look like vertically aligned in the middle while in OPS the icons are more at the right bottom corner.

mreese84:
@mightynimble I'm seeing the opposite. This is what I see in ops...
screen shot 2016-07-06 at 10 54 27 am

On my branch I see the action buttons located in the lower right corner (consistent with the collection list action items).

charleshuang80:
I see what Mark is seeing in ops. My fault, I didn't realize there was a different panel list that would also be affected.

@mreese84, in the lists I looked at there seems to be a little extra space on the bottom of the panels. If you want to keep it that way it seems acceptable but maybe not optimal. I could also take a stab at taking out the extra space if that would help, so let me know.

mightynimble:
@mreese84 and @charleshuang80 yes, when the view port is between 1200px and 1400px, the icons in ops are on top of the texts. If you widen the browser to >1400px, you will see the icons shifts to the right bottom corner.

On your PR branch, the icons move up to the middle of the granule panel when the width is greater than 1400px.

mreese84:
@mightynimble give that a shot. Previously there were a single set of rules controlling both the collection list items and the granule list items, but they were behaving differently at the different breakpoints. I added a new media query to handle the granule list item action buttons by themselves. Looks good across all browser widths to me on both lists.

mightynimble:
🚢 it

charleshuang80:
+1

PR #227, opened 2016-07-11
EDSC-1123: Update Colormaps rake task to use role v1.2

charleshuang80:


lib/tasks/colormaps.rake
```diff
@@ -21,9 +33,19 @@ namespace :colormaps do
...
     layers.each do |layer|
       id = layer.xpath("./ows:Identifier").first.content.to_s
-      url = layer.xpath("./ows:Metadata/@xlink:href").to_s
+
+      # get v1.2 role metadata node
+      target = layer.xpath("./ows:Metadata[contains(@xlink:role, '1.2')]")
+      if target.empty?
+        # layer does not have a metadata node with xlink:role v1.2,
+        # so try to use url from another metadata node (no version or v1.0)
+        url = layer.xpath("./ows:Metadata/@xlink:href").to_s
```
mightynimble: Not sure if this is the right url to get. The code changes look good to me. 👍

charleshuang80: This was the original method to get the url, and is only used if there is no `<ows:Metadata` tag that has a `xlink:role` attribute that ends in `/colormaps/v1.2`. So if the `gibs_url` is an updated version that has the `xlink:role` that has v1.2, it won't use this method, but if the `gibs_url` hasn't been updated to include versions in the role yet it will still pull the urls the old way.

macrouch:
It looks like you have to be on the VPN to use the GIBS SIT link?

I'm getting this:
```
> rake colormaps:load
Loading GIBS colormap data...
6 error(s), 588 file(s)
```
Is that right? Seeing errors seems bad

charleshuang80:
@macrouch you were right about needing to be on vpn to access the SIT GIBS link - which is odd, because it was working just fine for me last week without being on the vpn. But as we are actually pulling from the other link, that should not affect the rake task.

Digging into the errors, it looks like the 6 errors were produced because the (updated) colormaps have an opening tag of `<ColorMap title="No Data">`, and therefore exits out (in line 26 of lib/colormaps/xml_to_json.rb, as designed. An example is http://sit.gibs.earthdata.nasa.gov/colormaps/v1.2/OMI_Aerosol_Optical_Depth.xml. There are still values in the body, though, so I'm not sure if we need to check with Patrick or the GIBS team to know if we should process the colormap even though it has `title = No Data`.

mightynimble:
@charleshuang80 I have a question on the errors that @macrouch discovered. When GIBS promotes their SIT to production, are we gonna see the errors when we run the ```colormap``` rake task in production? I agree with you that we should let @bilts know about this error.

lib/colormaps/xml_to_json.rb
```diff
@@ -22,10 +22,19 @@ def self.xml_to_json(id, url, output_dir)
 
         if entries.size == 0
           # Colormaps 1.2
-          colormaps = xml.xpath("//ColorMaps/ColorMap").find {|map| map.attribute('title').to_s != 'No Data'}
+          no_data = xml.xpath("//ColorMaps/ColorMap").find { |map| map.attribute('title').to_s == 'No Data' }
```
macrouch: `xml.xpath("//ColorMaps/ColorMap")` is repeated 4 times here. Clean it up some so that you aren't making the same call more than once.

I think the `if` can be simplified some too. It looks like you want `colormaps = xml.xpath("//ColorMaps/ColorMap")` either way, so do that outside the `if`, then just check `if !no_data.nil?` or maybe `if no_data.present?` would work

macrouch:
looks good

PR #236, opened 2016-08-04
EDSC-1145 Monkey patch to enable faraday check proxy exceptions via '…

mightynimble:
…no_proxy' env var.


mreese84:
@mightynimble heads up, it looks like the long-term fix is being fast-tracked and may actually be in place by the end of the day tomorrow. So we'll hold off on merging this until we get a final word one way or the other.

PR #243, opened 2016-08-16
EDSC-1153: Store digest as String in db. Need to put in a test before…

mightynimble:
… merging.

### DIFF
--- a/app/models/access_configuration.rb
+++ b/app/models/access_configuration.rb
@@ -8,14 +8,14 @@ class AccessConfiguration < ActiveRecord::Base
 
   def self.get_default_access_config(user, cmr_concept_id)
     access_config = self.find_by(user: user, dataset_id: cmr_concept_id)
-    access_config.echoform_digest = JSON.parse(access_config.echoform_digest) if access_config.present? && access_config.echoform_digest.present?
+    access_config.echoform_digest = JSON.parse(access_config.echoform_digest) if access_config.present? && access_config.echoform_digest.present? && (access_config.echoform_digest.is_a? String)
     access_config
   end
 
   def self.set_default_access_config(user, cmr_concept_id, options, form_hashes)
     access_config = self.find_or_initialize_by(user: user, dataset_id: cmr_concept_id)
     access_config.service_options = options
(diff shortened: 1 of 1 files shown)


app/models/access_configuration.rb
```diff
@@ -8,14 +8,14 @@ class AccessConfiguration < ActiveRecord::Base
...
-    access_config.echoform_digest = JSON.parse(access_config.echoform_digest) if access_config.present? && access_config.echoform_digest.present?
+    access_config.echoform_digest = JSON.parse(access_config.echoform_digest) if access_config.present? && access_config.echoform_digest.present? && (access_config.echoform_digest.is_a? String)
     access_config
   end
 
   def self.set_default_access_config(user, cmr_concept_id, options, form_hashes)
     access_config = self.find_or_initialize_by(user: user, dataset_id: cmr_concept_id)
     access_config.service_options = options
-    access_config.echoform_digest = form_hashes.nil? ? nil : form_hashes.to_json
+    access_config.echoform_digest = form_hashes.nil? ? nil : form_hashes
```
xuebingli: IMO a cleaner alternative
```ruby
access_config.echoform_digest = form_hashes || form_hashes
```

mightynimble: Seriously, ```access_config.echoform_digest = form_hashes``` should do it. Thanks for pointing that out @xuebingli.

macrouch:
Couple of things

First, I don't think we need a new test, spec/features/data_access/access_option_defaults_spec.rb hits the data access page multiple times and checks that it remembers the previous access methods.

Second, this might have already been fixed. I had added the line, `serialize :echoform_digest, JSON` on an earlier ticket because of failing tests. This ensures that field is always handled as JSON when using it in Rails. That change was made a couple of days before this bug was created, so the fix wasn't to OPS yet.

If I remove the serialize line from the file I get failing tests again, but if I leave it in and remove line #18 completely, I still get everything passing.

mightynimble:
Looks like the issue was fixed. But we do need the cleanup commit to be merged to master.
👍

macrouch:
Thanks

PR #245, opened 2016-08-19
EDSC-1143: Ensure error message shows when service option request fails

charleshuang80:
because of bad/empty bounding box


app/controllers/data_access_controller.rb
```diff
@@ -115,6 +115,11 @@ def retrieval
           s['service_options']['number_processed'] = status['numberProcessed']
           s['service_options']['total_number'] = status['totalNumber']
           s['service_options']['download_urls'] = urls
+
+          if s['order_status'] == 'failed' && response_json['Exception'].nil? && !process_info.nil?
+            s['error_message'] = Array.wrap(process_info['message']).join("\n")
```
charleshuang80: @macrouch, @mightynimble, if there are multiple error messages, this does not actually allow them to show up on separate lines. If you have any suggestions to quickly change it so the messages can show up on separate lines (here or on the ko side), please let me know.

macrouch: Pushed a change, hopefully it gets you where you wanted to go with this

app/views/data_access/_access_error.html.erb
```diff
@@ -1,6 +1,14 @@
...
-  <p><strong>Message: </strong> <span data-bind="text: error_message"></span></p>
+  <p>
+    <strong>Error: </strong>
+    <span data-bind="text: error_code"></span>
+  </p>
+  <p>
+    <strong>Message: </strong>
+    <ul data-bind="foreach: error_message">
+      <li data-bind="text: $data"></li>
+    </ul>
```
macrouch: @mreese84 This is what this looks like with multiple errors. Even with one error it puts the message below the `Message:` text. How can we make that look better?

screen shot 2016-08-22 at 1 12 34 pm

PR #248, opened 2016-08-27
Avoid FOIA Redirect Page by Direct Link

rye:
The existing FOIA link resolves to a [(presumably) old URL](http://www.hq.nasa.gov/office/pao/FOIA/agency/), one which does a meta redirect to [the new page](http://www.nasa.gov/FOIA/index.html).  Instead of routing FOIA request through the old URL, it is probably better to simply link to the new FOIA page.

This could be classified as a low-impact change, as it only affects one link on the main application layout.


macrouch:
Thanks for the fix! (Sorry about the delay!)

PR #249, opened 2016-08-29
EDSC-1177 Add a method to time collection and granule controller acti…

mightynimble:
…ons.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -246,4 +246,12 @@ class ApplicationController < ActionController::Base
     end
   end
 
+  def log_execution_time
+    time = Benchmark.realtime do
+      yield
+    end
+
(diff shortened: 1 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -246,4 +246,12 @@ def metrics_event(type, data, other_data={})
     end
   end
 
+  def log_execution_time
+    time = Benchmark.realtime do
+      yield
+    end
+
+    Rails.logger.info "#{params[:controller].singularize.capitalize} request took #{(time.to_f * 1000).round(0)} ms"
```
macrouch: Should we also add the controller action being called?

mightynimble: Will do. Just asked Doug and he agrees to add the action to the log message.

macrouch:
looks good

PR #250, opened 2016-08-30
Edsc 1133

macrouch:


app/assets/javascripts/util/url.js.coffee
```diff
@@ -147,6 +139,18 @@ this.edsc.util.url = do(window
     new ChildCompressor('pg', new CwicGranuleIdListCompressor('cx'))
   ]
 
+#   new ParamFlattener(['science_keywords_h', '0', 'category'], 'fsc', false)
+  keywords = []
+  for index in [0...6]
```
mightynimble: With the change of CMR-3207, the ```index``` can be any integer, picked by CMR, as long as they indicates hierarchical relationships between ancestor/child facets. It is highly unlikely but possible that variable_level_1 could look like: ```science_keywords_h[34][variable_level_1]```. I can't think of a way to address that issue though (maybe using regex?).

mightynimble:
I noticed that after the CMR change was deployed, it broke our "removeHierarchicalFacets" feature:
When unchecking an ancestor/parent facet, all the applied facets are unchecked. Only the clicked one and its child and grandchildren should be cleared.

For example, click 'Biosphere' > 'Aquatic Ecosystems' > 'Lakes', then click 'Aquatic Ecosystems' to remove 'Lakes' and 'Aquatic Ecosystems'. Notice that 'Biosphere' is also removed.

macrouch:
I don't see that behavior on this branch when using cmr_env=sit

app/assets/javascripts/util/url.js.coffee
mightynimble: Scratch what I said... Although CMR is capable of returning an index number that is larger than 6, EDSC limits users to select only one topic which will set the index to always be 0.

mightynimble:
You are right. I forgot to set cmr_env=sit. 👍

PR #253, opened 2016-09-06
EDSC-1185: Redirect logs to STDOUT for UAT and SIT.

mightynimble:


macrouch:
The rails_12factor gem is supposed to be doing this for us. Can we figure out why that isn't working?

macrouch:
+1

PR #254, opened 2016-09-07
EDSC-0000 Temporarily remove 'foreman' command from the cronjobs when…

mightynimble:
… deploying to production since it is not in the cloud yet.


macrouch:
looks good

PR #255, opened 2016-09-07
EDSC-1164 Wait for everything loads before append the search box to the page.

mightynimble:


mightynimble:
For a race condition like this, I don't know if we can add a test to it...

macrouch:
This is what I see when waiting for the page to load. I don't really like the proposed solution of hiding the text box though. I think we need a better way to communicate "wait a second" to the user
screen shot 2016-09-08 at 1 23 25 pm

macrouch:
screen shot 2016-09-08 at 2 37 44 pm
screen shot 2016-09-08 at 2 37 36 pm

mreese84:
@mightynimble So this looks okay, but in cases of a really slow connection it could be confusing as to what is going on. Ideally we would have a "Please wait while the application loads" message with the "loading" indicator. For consistency you can use the same loading indicator that is used in the collection list panel when searches are performed.

mightynimble:
A 'please wait' message is added. You will see it if the network speed is slow enough. To simulate a slow network speed, use chrome and press ⌘+alt+i to open developer's tool, then click 'network' tab. Click on 'no throttling' dropdown at the top right of the tab. Select 'good 3G'. And reload the page.

config/schedule.rb
```diff
@@ -20,12 +20,13 @@
 # Learn more: http://github.com/javan/whenever
 set :environment, Rails.env
 set :job_template, "/bin/bash -c 'PATH=#{File.dirname(`which ruby`)}:$PATH; :job'"
-job_type :foreman_rake, "cd :path && :environment_variable=:environment foreman run bundle exec rake :task --silent :output"
+#TODO: This should be reverted when production is in the cloud
```
mreese84: Production has been moved to the cloud.

mreese84:
Assuming 1) a clean build, 2) merge conflicts resolved, and 3) a +1 from @macrouch on code review this is good to go for me.

mightynimble:
Looks like the TODOs are already addressed on master.

PR #257, opened 2016-09-07
EDSC:1168 Changed call to innerHTML to textContent

jboukes:
Fix for issue EDSC-1168 to merge with master

### DIFF
--- a/app/views/granules/prepare_script.html.erb
+++ b/app/views/granules/prepare_script.html.erb
@@ -131,7 +131,7 @@ fetch_urls <<'EDSCEOF'
   (function () {
 
     function downloadScriptFile() {
-      var content = document.getElementById("script").innerHTML;
+      var content = document.getElementById("script").textContent;
       var dl = document.createElement('a');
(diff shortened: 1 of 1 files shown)


macrouch:
Is there a way we can write a test for this?

jboukes:
I'm not entirely sure.  Here's how it could be tested manually (follow this to recreate the bug on production):

1. Click on 'Browse All Collections' and click on a resulting collection
2. Click the green 'Retrieve Collection Data' button
3. In the following form, by the label 'Select Data Access Method', select 'Download' and then click Submit
4. Click 'Download Access Script'
5. Click 'Download Script File' 

At that point the browser will download the script file.  It is within the file that the error occurs.  It appears correct in the browser - it is only the downloaded file that is broken.

If the script file downloaded contains the string '&amp;&amp;', then it would fail the test.  Success condition would be the lack of the string '&amp;&amp;'.  This isn't an entirely direct method of testing but it should be accurate.  The reasoning is that '&amp;&amp;' is the HTML entity code for '&&' - if the file is generated correctly, then that entity code should not be in it.

macrouch:
see #266

PR #259, opened 2016-09-08
EDSC-1138 added logging for collection relevancy metrics

macrouch:


app/controllers/collections_controller.rb
```diff
@@ -46,6 +46,19 @@ def use
     render :json => result, status: :ok
   end
 
+  def collection_relevancy
+    number_of_collections = 20
+
+    data = {
+      query: params[:query],
+      collections: params[:collections][0..number_of_collections - 1],
```
mightynimble: It looks like that we only return the first 20 collections on purpose. Do we care about which collection the user clicked if it is the 32nd collection?

mreese84: Just discussed this with Matthew and Jason. The selected_index still returns, which is good. We are going to add selected_collection to cover cases of a collection selected that is outside the top 20.

mightynimble:
👍

macrouch:
@mightynimble don't approve this yet, there are a lot of changes needed

app/controllers/application_controller.rb
```diff
@@ -111,26 +112,6 @@ def current_user
...
-      @@recent_lock.synchronize do
-        RecentCollection.find_or_create_by(user: current_user, echo_id: id).touch
-      end
-    else
-      # FIXME This does not work for guests loading directly to a project with more then 1
-      # collection, the session gets session does not carry over between the multiple calls to
-      # collections_controller:show
-      recent = session[:recent_collections] || []
-      recent.unshift(id)
-      session[:recent_collections] = recent.uniq.take(RECENT_DATASET_COUNT)
```
mightynimble: ```RECENT_DATASET_COUNT``` at line 28 should also be removed

app/models/recent_collection.rb
```diff
@@ -1,7 +0,0 @@
-class RecentCollection < ActiveRecord::Base
-  self.table_name = "recent_datasets"
```
mightynimble: Should we remove this table from db as well?

PR #262, opened 2016-09-14
EDSC-1181 Handle errors fromm CMR when generating download script.

mightynimble:


mightynimble:
You can follow the test to trigger a 400 error from CMR and see how it is handled in the script: 
 - it doesn't have any download links in the script. When it is run, a message will be shown to the user
 - an error message will also be shown on the script download page and because of the error , the download script button is not provided.

mreese84:
This looks good to me. We are now returning error messages from the CMR when submitting empty query parameters. Previously, a 500 error was thrown. This was the "quick fix" and a new issue has been filed to clear out empty query parameters to avoid sending them in the first place.

PR #263, opened 2016-09-14
EDSC-1165: Added 'find' functionality to arrays where it is not provided by the browser

jboukes:
Ticket 1165 describes an issue wherein a page does not load due within IE11.  Investigation revealed use of 'find' function applied to arrays - a function that is (surprisingly) not implemented in all Internet Explorer versions (though it was finally added in Edge).

The answer then was to provide a support function that provides the 'find' capability in the instance that the browser does not naturally provide it.

### DIFF
--- a/app/assets/javascripts/util/array.js.coffee
+++ b/app/assets/javascripts/util/array.js.coffee
@@ -9,3 +9,13 @@
 
   exports =
     pairs: pairs
+
+  if Array::find == undefined
+
+    Array::find = (callback, thisArg) ->
+      i = 0
+      while i < @length
+        if callback.call(thisArg or window, @[i], i, this)
+          return @[i]
+        i++
+      undefined
--- a/app/assets/javascripts/util/array.js.coffee
+++ b/app/assets/javascripts/util/array.js.coffee
@@ -10,12 +10,12 @@
   exports =
     pairs: pairs
 
-  if Array::find == undefined
+if Array::find == undefined
 
-    Array::find = (callback, thisArg) ->
-      i = 0
-      while i < @length
-        if callback.call(thisArg or window, @[i], i, this)
-          return @[i]
-        i++
-      undefined
+  Array::find = (callback, thisArg) ->
+    i = 0
+    while i < @length
+      if callback.call(thisArg or window, @[i], i, this)
+        return @[i]
+      i++
+    undefined


macrouch:
This pull request seems to have a lot more test failures than your other pull request. Do you get failures in your tests locally? We need to make sure this change didn't break anything.

jboukes:
'Rake Spec' as well as 'Rake Test' have never worked for me - that is another issue I'm working on.  This fix was tested manually.

If you look at the file changed, the only thing added was a bit of CoffeeScript - script which adds the 'find' function to arrays.  The other ticket which also failed Travis was similar yet even less invasive - changing a called to 'innerHTML' to 'textContent'.

It seems likely that these Travis failures are the result of a forked version of the code pushing to master.   The travis docs do mention that this is a danger.  Do we have any 'encrypted variables or data' which - if Travis does not have access to them - could cause these failures?

_**Pull Requests and Security Restrictions #**

The most important restriction for pull requests is about secure environment variables and encrypted data.

A pull request sent from a fork of the upstream repository could be manipulated to expose any environment variables. The upstream repository’s maintainer would have no protection against this attack, as pull requests can be sent by anyone with a fork.

Travis CI makes encrypted variables and data available only to pull requests coming from the same repository. These are considered trustworthy, as only members with write access to the repository can send them.

Pull requests sent from forked repositories do not have access to encrypted variables or data.

If your build relies on these to run, for instance to run Selenium tests with BrowserStack or Sauce Labs, your build needs to take this into account. **You won’t be able to run these tests for pull requests from external contributors.**

To work around this, restrict these tests only to situations where the environment variables are available, or disable them for pull requests entirely.

Here’s an example of how to structure a build command for this purpose:

script:
```
   - 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then bash ./travis/run_on_pull_requests; fi'
   - 'if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then bash ./travis/run_on_non_pull_requests; fi'
```_

[Source](https://docs.travis-ci.com/user/pull-requests)

jboukes:
Oh, it does too work - though perhaps a typo snuck in there.  I will re-investigate and collect screenshots of it working (or not).

I tested it and it worked within IE 11 as well as Chrome and Firefox (though it always worked with Chrome and Firefox).

If you could provide a screenshot of it bombing out with the console errors, that would be awesome.  **But before you do that** wait for me to collect screenshots (because if it doesn't work for me, there's no reason for you to proceed further).

Will report back in a bit.

jboukes:
I believe I have it now - it *did* work - it just also unintentionally clobbered another helper function, 'pairs'.  I found that the granules were not loading as Matthew said, and console pointed to the lack of the 'pairs' function.

I've restructured the CoffeeScript so that the new find function is not stepping on the toes of the pairs function, and now granules are loading, and the original bug is fixed as well.  I'll do a new pull to master.

macrouch:
See #267

PR #264, opened 2016-09-15
Edsc 990 test

macrouch:
added logging and split rake tasks into their own health checks

### DIFF
--- a/app/models/collection_extra.rb
+++ b/app/models/collection_extra.rb
@@ -9,6 +9,7 @@ class CollectionExtra < ActiveRecord::Base
   store :orbit, coder: JSON
 
   def self.build_echo_client(env=(@cmr_env || Rails.configuration.cmr_env))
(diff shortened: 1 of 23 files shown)


app/models/collection_extra.rb
```diff
@@ -9,14 +9,18 @@ class CollectionExtra < ActiveRecord::Base
...
+    puts "Building echo client for env: #{env || 'env not working'}"
     Echo::Client.client_for_environment(env, Rails.configuration.services)
   end
 
   def self.create_system_token(client)
     response = client.create_token(ENV['system_user'], ENV['system_user_password'])
     if response.success? && response.body['token']
-      response.body['token']['id']
+      token = response.body['token']['id']
+      puts "token: #{token}"
```
mightynimble: Is this token sensitive? Is it okay to be seem in the splunk by other users?

macrouch: I'll remove it to be safe

PR #269, opened 2016-09-16
EDSC-1173: use --netrc-file to specify the temp netrc file in curl.

mightynimble:


mightynimble:
The issue was replicable if a user doesn't have .netrc file in the home directory. When the script is run, some of the curl commands use a n option which indicates to use the default .netrc file instead of the script generated temp .netrc. The -n opsion is replaced with -netrc-file so that we can specify to use the temp one.

PR #271, opened 2016-09-20
EDSC-1079: Updated 'Native' collections link so that it downloads the XML represention of the data instead of displays it in HTML; added new 'HTML' link to display content as HTML

jboukes:
- 'native_url' was updated to include .native extension, which signals CMR to generate and download to the client the requested collection data in its native format.
- A new field was added to the collections object, 'html_url', to store the newly requested path to the HTML representation of the data.
- Test files were also updated so as to expect these new links; rspec was run successfully with no failures (a few pendings).


jboukes:
Fair enough - will update!

jboukes:
Update is made!  It now uses .native instead of .xml, and the .html extension is now explicit.

I had to make another update to a test for it to pass rspec; all tests passed on my local machine.

macrouch:
See #272

PR #274, opened 2016-09-22
EDSC-1188 fix so the log file gets picked up correctly

macrouch:


macrouch:
That branch is currently in SIT right now. You can search splunk for "Starting data:load" and see logs from the rake tasks that haven't been running

PR #276, opened 2016-09-28
EDSC-1153 Fix an error that shows up when loading collections in UAT.

mightynimble:


macrouch:
Can we write a test for this? Manually put a string into the database field and make sure order options still load?

mightynimble:
@macrouch I can't find a way to store a String in the db because it is prevented in the ```AccessConfiguration```.

PR #281, opened 2016-10-06
EDSC-1206 Fix download button in firefox and IE

mightynimble:
…loyed.

### DIFF
--- a/app/views/granules/prepare_script.html.erb
+++ b/app/views/granules/prepare_script.html.erb
@@ -28,6 +28,7 @@
       padding: 8px 12px;
       border-radius: 3px;
       font-size: small;
+      text-decoration: none;
     }
 
     div {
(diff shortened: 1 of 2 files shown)


app/views/granules/prepare_script.html.erb
```diff
@@ -132,16 +133,25 @@ fetch_urls <<'EDSCEOF'
...
+      window.navigator.msSaveOrOpenBlob(blob, e.target.scriptFileName);
     }
 
     document.addEventListener('DOMContentLoaded', function () {
-      document.getElementById('download-button').addEventListener('click', downloadScriptFile);
+      var dlLink = document.getElementById('download-button');
+      dlLink.scriptFileContent = document.getElementById("script").textContent;
+      dlLink.scriptFileName = 'download.sh';
+
+      var isIE = /*@cc_on!@*/false || !!document.documentMode;
```
mightynimble: @cc_on is only for IE. So if it is IE,` /*@cc_on!@*/false` compiles to `!false`.

PR #283, opened 2016-10-13
EDSC-1214: Removed granule filter from href

jboukes:
Added code to remove the granule filter from the link returning to collection results.

### DIFF
--- a/app/controllers/data_access_controller.rb
+++ b/app/controllers/data_access_controller.rb
@@ -25,6 +25,13 @@ class DataAccessController < ApplicationController
     if !@back_path || ! %r{^/[\w/]*$}.match(@back_path)
       @back_path = '/search/collections'
     end
+
+    #This block removes the granule filter per EDSC-1214
+    @paramsWithoutGranuleFilter = request.query_string
+    params = CGI.parse(@paramsWithoutGranuleFilter) 
+    params.delete('sgd')
+    @paramsWithoutGranuleFilter = URI.encode_www_form(params).to_s
+    @paramsWithoutGranuleFilter = URI.unescape(@paramsWithoutGranuleFilter)
   end
 
   def retrieve
--- a/app/views/data_access/configure.html.erb
+++ b/app/views/data_access/configure.html.erb
@@ -9,7 +9,7 @@
 
 <div class="data-access">
   <nav class="data-access-nav">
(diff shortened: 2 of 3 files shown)


jboukes:
Rspec test is now also submitted ;)

macrouch:
Can you open this pull request from your newly cloned project, and not from the fork so the tests have a chance to pass

macrouch:
Your test doesn't fully replicate the steps from the ticket.

Steps:

1. Download a single granule
2. On the data access retrieval page, choose "Back to Search Results"
3. Click on the collection that you downloaded the granule from

For step 3 you should test that more than one granule exists in the granule list for the selected collection.

jboukes:
>>"Can you open this pull request from your newly cloned project, and not from the fork so the tests have a chance to pass"

Conflicting requests, here - Mark asked that I continue to use the fork for this particular ticket and use the clone for all future items.

jboukes:
>>"Your test doesn't fully replicate the steps from the ticket."

It really does, unless I am missing something.  I took advantage of an existing test that did exactly that - test a collection with multiple granules and viewing only a single granule:

```
  context 'when the user is logged in' do
    before :all do
      load_page :search, focus: downloadable_collection_id
      login
      wait_for_xhr
      **first_granule_list_item.click_link "Retrieve single granule data"**
      wait_for_xhr
    end

    after :all do
      Capybara.reset_sessions!
    end

    it 'only configures one granule' do
      **expect(page).to have_content "1 Granule"**
    end

    **it 'does not have granule filter in back url' do
      expect(find_link('Back to Search Session')[:href]).not_to include("&sgd=")
    end**

    it 'limits the data access to only the selected granule' do
      click_link 'Expand List'
      expect(page).to have_content 'FIFE_RAIN_30M.72981621.r30'
    end
  end
```

This test is specifically using this collection: https://search.earthdata.nasa.gov/search/granules?p=C179002914-ORNL_DAAC&tl=546393600!4!!

This collection has 117 granules.

So, the logic fits our needs; the test is **already** viewing a single granule out of a multi-granule collection, and I added a check to make sure the granule filter is removed from the 'back' link.  This feels like the right spot to add this test.  If you like, I could add an 'expect' for it to have 117 granules, though I'm not sure if that's really necessary.

PR #284, opened 2016-10-17
EDSC-1214: Removed granule filter from href and updated associated rspec tests...

jboukes:


macrouch:
Moving the test discussion to the open pull request.

What you added only tests the change you made. That is good the verify, but you never actually test what the ticket is about, viewing a collection again after retrieving a single granule. You never go back to the search results and verify that viewing the collection results in being able to view all the granules.

jboukes:
All tests have passed!  That's a beautiful thing, at least.

I see what you are saying Matthew - that is a good point.  I was thinking more in proving that the granule filter was removed; not that removing it accomplished what the ticket was requesting.

I'll add another test where it goes back and confirms that 117 granules are showing.

jboukes:
Alright - I'd had it backwards and I think (hope) it's correct now - added a test to check for those 117 granules *after* returning to the page.  I did also remove the unnecessary check.

spec/features/data_access/single_granule_access_spec.rb
```diff
@@ -36,10 +36,21 @@
...
+      expect(find_link('Back to Search Session')[:href]).not_to include("&sgd=")
+    end
+
     it 'limits the data access to only the selected granule' do
       click_link 'Expand List'
+      wait_for_xhr
       expect(page).to have_content 'FIFE_RAIN_30M.72981621.r30'
     end
+
+    it 'does not filter the granules upon returning to the collection' do
```
macrouch: We try not to follow links within an `it` statement. You would want a new context `context 'when returning to the search results' do` to handle this. Inside the context will be a `before` block that does the `click_link` and then and `it` that asserts the list is correct

jboukes:
Round 3!  Following your suggestion, it seemed that the check to see if the granule filture (e.g., '&sgd=") is in the href text really isn't necessary - it's too code-literal and not checking actual functionality.

I created a new context block that is more to the point - clicking a single granule of a multi-granule collection, then clicking 'Back to Search Session', and then confirming the presence of all granules.

spec/features/data_access/single_granule_access_spec.rb
```diff
@@ -38,10 +38,27 @@
 
     it 'limits the data access to only the selected granule' do
       click_link 'Expand List'
+      wait_for_xhr
       expect(page).to have_content 'FIFE_RAIN_30M.72981621.r30'
     end
   end
 
+  context 'while the user is viewing a single granule of a multi-granule collection' do
```
macrouch: Now you're on the right track. This does do everything I was looking for, but you put it in the wrong place. The context right above this one (line 22) already loads the page and clicks the single granule link. You can nest contexts. The nested context's `before` block will take over the state of the page where the parent context left it, so you don't have to repeat all the setup in code, and it executes the tests faster because you don't have to do a full page load from the `load_page` call.

jboukes:
Round 4 :)  Nested context and all!

Thank you for your explanations and patience - it's helped to make the team's testing philosophy clearer.

PR #292, opened 2016-10-31
EDSC-1236 Prompt user differently (only asking for email addr) on (NS…

mightynimble:
…IDC) anonymous downloading.


mightynimble:
@mreese84 - need your input on the wording and the way that users are (temporarily) prompted on script-downloading the NSIDC data. (NSIDC uses anonymous download right now and it is said they are switching to URS "pretty soon".) Asking only for email address is what we ( @macrouch and @jboukes ) discussed during the PI planning.

mreese84:
👍

PR #294, opened 2016-11-02
EDSC-1234 Fix the js error mentioned in the ticket.

mightynimble:


mightynimble:
If you see requests are still referring to the old endpoint `api.echo.nasa.gov`, restart the rails app *and background_jobs*.

PR #296, opened 2016-11-03
EDSC-1235: Initial fix to use ID instead of short name; remove spinner in case of exception

jboukes:
This commit fixes two issues.

CWIC expects the usage of 'concept_id' - simply noted as 'id' within our code - and not short_name.  The use of short_name has been retired.  This change alone fixes the issue described in the ticket - it will correctly fetch the granules (albeit zero in this case).

The second issue fixed is dismissing the 'loading granules...' spinner in the case that an exception is hit.  The changes to Granules.jsx accomplish this - by setting isLoading to false as well as the other values in the 'error' condition of  the AJAX call, the app successfully dismisses the spinner and replaces it with what would be expected normally.

The second issue will be difficult to test, as the first issue - changing short_name to ID - stops it from occurring.  A dev may explore this code though by altering their local instance of edsc-plugin.datasource.cwic.jsx to call short_name instead of ID.  Then repeat the replication described in the ticket to initiate the error.  If the Granules.jsx has been updated with this commit, then the front end should behave as would normally be expected.


app/views/search/_granule_list.html.erb
```diff
@@ -97,7 +97,7 @@
               </div>
             </li>
           </ul>
-          <p class="busy panel-list-load-more" data-bind="visible: granules().isLoading()">
+          <p class="busy panel-list-load-more" data-bind="visible: granules.isLoading">
```
macrouch: This change completely removes the "Loading granules..." message

jboukes: It does :(  I see that now.

The issue I am having is setting isLoading to false - which should trigger the spinner to disappear - is not working for some reason.  I thought it might be a syntax error within the view.

However, that means this other code in index.html.erb must be incorrect?  Or is there any reason it would use this syntax? 

```            <p class="panel-list-load-more" data-bind="visible: collections.isLoading">
              Loading collections...
            </p>```

jboukes: Fix found!  It's the third permutation.  visible: granules().isLoading works - the spinner shows and disappears as it is supposed to.

While unrelated, it should be pointed out that the syntax does seem to be incorrect within index.html.erb.

PR #297, opened 2016-11-04
EDSC-1205 support for different gibs layer resolutions in different p…

macrouch:
…rojections

### DIFF
--- a/app/assets/javascripts/modules/map/leaflet-plugins/granule_layer.js.coffee
+++ b/app/assets/javascripts/modules/map/leaflet-plugins/granule_layer.js.coffee
@@ -171,6 +171,19 @@ ns.GranuleLayer = do (L
       matched = false
       for optionSet in @multiOptions when @_matches(granule, optionSet.match)
         matched = true
+
+        # set resolution to {projection}_resolution if it exists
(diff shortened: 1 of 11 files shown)


macrouch:
To test locally we have to make sure that the tags are up to date from this branch in SIT. Get with me to make sure it is ready for you

app/assets/javascripts/modules/map/leaflet-plugins/granule_layer.js.coffee
```diff
@@ -171,6 +171,19 @@ ns.GranuleLayer = do (L
       matched = false
       for optionSet in @multiOptions when @_matches(granule, optionSet.match)
         matched = true
+
+        # set resolution to {projection}_resolution if it exists
+        oldResolution = optionSet.resolution
+
+        if this._map.projection == 'geo'
```
mightynimble: `newResolution = optionSet.geo_resolution if this._map.projection == 'geo'` makes it more uniform with line 180 and 181 :)

macrouch: oops, I just saw this comment after I merged. I'll make that change on master

PR #301, opened 2016-11-10
EDSC-975: Initial check-in including fix and rspec test updates

jboukes:
Initial fix for 975 - the code fetches the user's email from URS, and then uses it to fill any empty email xml field within the model before the model is used to create the ESI form.


spec/features/map/nlp_spatial_temporal_extraction_spec.rb
```diff
@@ -40,11 +40,11 @@
     end
 
     it 'filters collection results' do
-      expect(page).to have_content('6325 Matching Collections')
+      expect(page).to have_content('11553 Matching Collections')
```
macrouch: This file sticks out in the review. I don't think you shouldn't have changed anything that caused this test to need to be updated. I'm also unsure of why the collections_requests.yml and collections_responses.yml files were changed

spec/features/data_access/service_access_spec.rb
```diff
@@ -14,15 +14,22 @@
...
+      end
+      after :all do
         fill_in 'Email Address', with: "patrick+edsc@element84.com\t"
         click_on 'Continue'
         click_on 'Submit'
       end
-
+    end
+    
+    context 'and submitting an ESI service request' do
```
macrouch: this context needs to be nested within `context 'when choosing the ESI Service' do`

spec/features/data_access/service_access_spec.rb
```diff
@@ -14,15 +14,22 @@
...
-    context 'and submitting an ESI service request' do
+    context 'when choosing the ESI Service' do
       before :all do
         choose 'AE_Land.2 ESI Service'
         wait_for_xhr
+      end
+      it 'prepopulates the email address field' do
+        expect(find_field('Email Address').value).not_to be_blank
+      end
+      after :all do
```
macrouch: This block should be a `before` block under `context 'and submitting an ESI service request' do`

app/assets/javascripts/models/knockout_extensions.js.coffee
```diff
@@ -92,14 +92,30 @@ do (ko, $=jQuery) ->
 
   ko.bindingHandlers.echoform =
     init: (element, valueAccessor, allBindings, viewModel, bindingContext) ->
+      @email = ""
```
macrouch: I don't think we need to do all of this. In Chrome, `edsc.page.account.email()` while on /data/configure gives me the URS email address of the logged in user. I think accessing the email address already stored on the client would be better.

jboukes: Nice - I will try this.  Thanks!

spec/features/data_access/service_access_spec.rb
macrouch:
```
context 'when choosing the ESI Service' do
  before :all do
    choose 'AE_Land.2 ESI Service'
    wait_for_xhr
  end
  
  it 'prepopulates the email address field' do
    expect(find_field('Email Address').value).not_to be_blank
  end

  context 'and submitting an ESI service request' do
    before :all do
      click_on 'Continue'
      click_on 'Submit'
    end
    
    # Cannot reliably display a progress bar using recordings
    it 'displays an error message', pending_fixtures: true do
      wait_for_xhr
      within '.access-error-code' do
        expect(page).to have_content('CollectionDisabled')
      end
      within '.access-error-message-list' do
        expect(page).to have_content('This collection is currently not configured for subagent HEG')
      end
    end
  end
end
```

jboukes: Thank you!

mreese84:
@jboukes why are there changes to nlp_spatial_temporal_extraction_spec.rb? Those are the tests that are failing and I'm not sure why updating fixing the pre-populated email would need an update to that those tests.

jboukes:
I thought I said this in an email last night - there are changes to nlp_spatial_temporal_extraction_spec.rb simply because the test was failing.  The count has changed.  I mentioned the changing counts of tests as well days ago in an email.  Why has the count changed?  I have no idea.  I presumed it was because we're dealing with a living database and the data had simply changed.  But you're right - this email update did nothing to influence that.  I'm just trying to get the test to stop failing.

app/assets/javascripts/models/knockout_extensions.js.coffee
```diff
@@ -1,4 +1,4 @@
-do (ko, $=jQuery) ->
+do (ko, $=jQuery, ajax=@edsc.util.xhr.ajax) ->
```
mreese84: Drop this helper!

PR #302, opened 2016-11-11
EDSC-1275 Find form element by exact name. Add a test and new hand ed…

mightynimble:
…ited fixtures.


app/services/esi_client.rb
```diff
@@ -93,7 +93,7 @@ def add_parameter(field_symbol, value)
   end
 
   def find_field_element(field_symbol, data_type = 'ecs')
-    find_by_xpath("//*[contains(name(), '#{data_type}:#{field_symbol.to_s}')]")
+    find_by_xpath("//#{data_type}:#{field_symbol.to_s}")
```
macrouch: This change was made to fix EDSC-1233, can you very that bug no longer exists

mightynimble: The test is passing so I didn't pay attention. Let me check

mightynimble: The revert of that line doesn't affect the bbox:
http://n5eil01u.ecs.nsidc.org/ops/esir/5000000015504/requestSummary.txt

PR #307, opened 2016-11-16
EDSC-1200: Added Faraday rescue for connection failure

jboukes:
A rescue method has been added to urs_client which will recover from a Faraday::Error::ConnectionFailed exception and will return the appropriate error.

To convenience testing, a parameter hook has been provided that can be used to trigger this error.  This technique uses the external site https://expired.badssl.com to throw the exception.  To trigger the error in the health check, try: http://edsc.dev/health?urs_test=true

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,7 +5,7 @@ class ApplicationController < ActionController::Base
   before_filter :validate_portal
 
   rescue_from Faraday::Error::TimeoutError, with: :handle_timeout
-  rescue_from Faraday::Error::ConnectionFailed, with: :handle_connectionfailed
+  rescue_from Faraday::Error::ConnectionFailed, with: :handle_connection_failed
 
   def redirect_from_urs
     last_point = session[:last_point]
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,6 +5,7 @@ class ApplicationController < ActionController::Base
   before_filter :validate_portal
 
   rescue_from Faraday::Error::TimeoutError, with: :handle_timeout
+  rescue_from Faraday::Error::ConnectionFailed, with: :handle_connectionfailed
 
   def redirect_from_urs
     last_point = session[:last_point]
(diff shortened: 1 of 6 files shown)


jboukes:
We're not really at the 'add a test' stage yet, despite the PR.  Please see the "Instructions Needed to Replicate EDSC-1200" email.  

Once we get the conversation there settled, then we can come up with an rspec test.

jboukes:
*Now* we're at the 'create a test' point :)

mreese84:
Pinging this PR. @macrouch requested changes... what's the status?

jboukes:
@mreese84: We're near the end - I've been working with @macrouch to come up with a method of testing.

The issue is this - the app needs to hit a bad certificate (or simulation thereof) in order to throw the rescue and test this fix.  That's not easy from rspec perspective.  I've found a way to do it by calling a remote site (a handy tool site that provides bad certificates for testing purposes).  However, this requires code to be in the application to help that test, and Matthew is not fond of that (and I agree with him).

The solution then is to record via VCR the call to the handy certificate tool site - then, with these cassettes in play, *remove* the testing code from the application code and just use the cassettes for testing.  

The final problem that we are hitting is that the call does not want to record!  I found recently that it will record when wrapped in a Faraday object but then it also breaks the rescue and thus the fix.  I have a new idea that I'll be trying today that *might* get passed this issue.

tl;dr: It's very close but one last thing is messing up testing

app/controllers/application_controller.rb
```diff
@@ -77,6 +78,10 @@ def handle_timeout
     end
   end
 
+  def handle_connectionfailed
```
macrouch: should be `handle_connection_failed`

spec/helpers/mock_response.rb
```diff
@@ -13,6 +13,10 @@ def self.edsc_dependency(body)
     MockResponse.new(body, {}, 200)
   end
 
+  def self.connection_failed(body)
+    MockResponse.new(body, {}, 500)
```
macrouch: in application_controller you check for a 504, but here you throw a 500

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,7 @@ class ApplicationController < ActionController::Base
   before_filter :validate_portal
 
   rescue_from Faraday::Error::TimeoutError, with: :handle_timeout
+  rescue_from Faraday::Error::ConnectionFailed, with: :handle_connectionfailed
```
macrouch: you missed `handle_connectionfailed` -> `handle_connection_failed ` here

PR #308, opened 2016-11-18
Edsc 1265

macrouch:


app/models/retrieval.rb
```diff
@@ -157,4 +168,11 @@ def update_access_configurations
       end
     end
   end
+
+  def self.test_query_params(client, params, token)
```
mightynimble: a better name? :)

app/models/retrieval.rb
```diff
@@ -64,24 +64,34 @@ def self.process(id, token, env, base_url, access_token)
...
+                                                     method['id'],
+                                                     method['method'],
+                                                     method['model'],
+                                                     user_id,
+                                                     token,
+                                                     client,
+                                                     access_token)
+                method[:order_id] ||= []
+                method[:order_id] << order_response[:order_id]
+                method[:dropped_granules] = order_response[:dropped_granules]
```
mightynimble: should this be ` method[:dropped_granules] << order_response[:dropped_granules]`? If it is an ASTER order and is split into 2 orders. The first order drops, say 130 granules and the second one drops 500 granules. A total of 630 granules should be shown in the dropped granules panel.

I think you'd better let ASTER (LP DAAC?) know before you submit orders with large numbers of granules. I was black-listed last time when I did this...

mightynimble: You don't have to submit an order with 2k+ granules. Just set the split threshold to 50 granules and submit a 100-granule order.

app/controllers/data_access_controller.rb
```diff
@@ -74,58 +74,70 @@ def retrieval
...
-
-        orders.each do |order|
-          echo_order = echo_orders[order['order_id']]
-          if echo_order
-            order['order_status'] = echo_order['state']
-          else
-            # echo order_id doesn't exist yet
-            order['order_status'] ||= 'creating'
+      order_ids.each do |order_id|
+          order_response = order_id && order_id.compact.size > 0 ? echo_client.get_orders({id: order_id}, token) : nil
```
macrouch: `order_ids` is the array of orders per collection. so it could look like this

```
[collection_1_order_id, [collection_2_id_1, collection_2_id_2]]
```

I'm not sure if it will work sending all of them at once

app/controllers/data_access_controller.rb
```diff
@@ -74,58 +74,70 @@ def retrieval
...
+          if order_response && order_response.success?
+            echo_orders = order_response.body.map {|o| o['order']}.index_by {|o| o['id']}
+
+            orders.each do |order|
+              Array.wrap(order['order_id']).each do |id|
```
macrouch: This is looping through the order ids of split orders

app/models/retrieval.rb
macrouch: Maybe? I don't know what dropped_granules are

macrouch:
@mreese84 How do you think we should reword "Access methods other than "Download" are limited to the first 2000 granules."?

mreese84:
@macrouch I have some changes queued up but want to work all the way through the workflow. I forgot the rake (i think) command to workoff the order queue. I can't seem to find it anywhere.

mreese84:
I'm good with this. @mightynimble I've updated (and moved) the red text in question.

PR #309, opened 2016-11-20
EDSC-1235: Removed unnecessary call which was causing the granules table to blink

jboukes:
This is a quick add-on for EDSC-1235.  When a collection is being viewed and the granules are loading, the granules table would 'blink' (as in, appear, disappear, and then re-appear) very rapidly upon loading.  This is only with CWIC collections (so far as is known).

The 'callback' function that was removed from the AJAX success condition stored the function _load - which loads the data again.  Since it is a success callback, it only happens when the data is already successfully loaded!  It appears as if this call was actually causing granule data to load _twice_ within the table, and this was causing the 'blink'.

This particular blink bug is currently on production and so can be viewed by looking up 'USGS_EDC_EO1_ALI' (for convenience: https://search.earthdata.nasa.gov/search/granules?p=C1220566654-USGS_LTA&tl=1463858789!4!!&q=USGS_EDC_EO1_ALI&ok=USGS_EDC_EO1_ALI).  When the collection data is retrieved, watch the table of granules on the left - it will load, blink out of existence, then reappear.  Typically it will still be on 'loading granules' when it first disappears.

Testing seems to show this second '_load' is not necessary - granules appear just fine after its removal.  Local testing was done with CWIC collections USGS_EDC_EO1_ALI (which has many granules) and collection AIRX2RET_NRT (which has 0 granules).  Neither table blinks any longer and app behavior is fine.

### DIFF
--- a/edsc/plugins/datasource/cwic/src/js/Granules.jsx
+++ b/edsc/plugins/datasource/cwic/src/js/Granules.jsx
@@ -128,6 +128,7 @@ let CwicGranules = (function() {
       this._loadOsdd(() => this._load(params, current, callback));
       return;
     }
+
     let url = this._urlFor(params);
     this.abort();
     this.isLoading(true);
@@ -173,7 +174,7 @@ let CwicGranules = (function() {
     this.currentRequest = ajax(xhrOpts);
   };
 
-  CwicGranules.prototype._parseOsdd = function (doc) {
+  CwicGranules.prototype._parseOsdd = function (doc, callback) {
 
     let urls = doc.getElementsByTagNameNS(xmlNamespaces.os, 'Url');
     for (let i = 0; i < urls.length; i++) {
@@ -184,6 +185,7 @@ let CwicGranules = (function() {
         };
       }
     }
+    if (callback) callback();
     return null;
   };
 
@@ -196,7 +198,7 @@ let CwicGranules = (function() {
       url: this.osddPath,
       retry: () => this._loadOsdd(callback),
       success: (data, status, xhr) => {
-        this.osdd(this._parseOsdd(data));
+        this.osdd(this._parseOsdd(data, callback));
       },
       complete: () => {
         this.isLoading(!this.currentRequest);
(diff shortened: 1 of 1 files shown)


mreese84:
Anything holding this up?

jboukes:
Just waitin' for the OK to merge ;)

mightynimble:
I am still seeing blinks here: http://edsc.dev/search/granules?p=C1233769002-GES_DISC&tl=1464815168!4!!&q=AIRX2RET_NRT&ok=AIRX2RET_NRT

From the js console I can see that one request was aborted as another one fired, which could be contribute to the blinking. It also makes the fixing of the blinking much more difficult (for we need to hunt down why the first request aborted). Should we write a ticket and throw it into backlog?

jboukes:
I've changed the nature of the fix and I believe it is now more in line with the original intent of design.

The issue appears to be this call in the beginning of _load:
```
if (!this.osdd()) {
   this._loadOsdd(() => this._load(params, current, callback));
   return;
 }
```
This is (essentially) saying "if osdd is not there yet, then load osdd".  The function call includes 'callback' which is the _load function itself.  The idea is that loadOsdd is supposed to - well load osdd - and then is supposed to reload _load, which - with osdd now loaded - will bypass this conditional and carry on.

However, this is what happens (prior) with loadosdd:

```
success: (data, status, xhr) => {
        this.osdd(this._parseOsdd(data, callback));
        if (callback) callback;
      },
```

The call to parseOsdd is meant to load the osdd, and then it is supposed to use the callback function to redo _load.  HOWEVER - an asynchronous issue hits as the callback function fires faster than this.osdd - the callback returns to _load before this.osdd can complete its work!  This means when it re-enters _load, it again hits the if (!this.osdd()) conditional, and *again* calls loadosdd  - thus the blink.

So, the new fix strategy is to force callback to only be called when we're sure parseOsdd is complete.  It was removed from the success condition and instead is passed as a parameter to parseOsdd and then is executed at function end.  This forces the callback to be performed when osdd is loaded, thus when it hits _load it successfully bypasses the if (!this.osdd()) (as osdd is now loaded) and carries on without reloading, and thus no more blink.

jboukes:
@mightynimble - I made my comment before seeing yours :)  I'll check out that collection with this different fix and report back...

PR #312, opened 2016-12-01
EDSC-1280 display all temporal extents on collection details view

macrouch:


mreese84:
Please note, from a browser standpoint we cannot test this as the CMR doesn't send multiple temporal extents yet. This PR changes how we request dates and should request individual dates and list them out on the collection details panel. As long as the code/tests look good, we should approve this PR, merge it, and wait for the CMR to send multiple At that point we can have our testers verify the issue.

PR #313, opened 2016-12-01
EDSC-1281 added exact_match to collection relevancy logging

macrouch:


app/assets/javascripts/models/ui/collections_list.js.coffee
```diff
@@ -67,11 +67,20 @@ ns.CollectionsList = do ($ = jQuery
       query = @query.params()
       collections = @collections.results().map (collection) -> collection.id
       index = @collections.results().indexOf(collection)
+
+      exact_match = switch query.free_text
+        when collection.dataset_id then true
+        when collection.id then true
+        when collection.short_name() then true
+        when "#{collection.short_name()}.#{collection.version_id}" then true
```
mightynimble: I don't know how detail this ticket asks.
AE_Land.002 is more likely to be type in than AE_Land.2. So maybe a regex match here?

app/assets/javascripts/models/ui/collections_list.js.coffee
```diff
@@ -67,11 +67,20 @@ ns.CollectionsList = do ($ = jQuery
       query = @query.params()
       collections = @collections.results().map (collection) -> collection.id
       index = @collections.results().indexOf(collection)
+
+      exact_match = switch query.free_text
```
mightynimble: The strings should be compared in lower (or upper) cases. So AE_Land matches ae_land.

app/assets/javascripts/models/ui/collections_list.js.coffee
macrouch: I don't think we need to do anything here (besides change the `.` to a `_`, oops)

My reasoning is if the actual entry id is `AE_Land.002` and they search for `AE_Land.2`, they won't get search results from CMR. Same if it is reversed.

mightynimble: Why change the dot to an underscore?

PR #317, opened 2016-12-14
EDSC-1298 add browse_scaler_url_(env) for each environment

macrouch:


lib/echo/client.rb
```diff
@@ -12,7 +12,7 @@ def initialize(service_config, urs_client_id)
       clients << CmrClient.new(@config['cmr_root'], urs_client_id)
       clients << EchoClient.new(@config['echo_root'], urs_client_id)
       clients << UrsClient.new(@config['urs_root'], urs_client_id)
-      clients << BrowseScalerClient.new(ENV['browse_scaler_url'] || @config['browse_scaler_root'], urs_client_id)
+      clients << BrowseScalerClient.new(ENV["browse_scaler_url_#{Rails.env.to_s}"] || @config['browse_scaler_root'], urs_client_id)
```
mightynimble: `Rails.env.to_s` evals to `production`, not `prod`. This line is neat, I like it. Change the one in js.coffee.erb to `browse_scaler_url_production`?

PR #321, opened 2016-12-20
EDSC-1170 timeline auto pans to temporal constraint

macrouch:


mreese84:
@mightynimble I agree that the highlighted temporal range should be centered horizontally.

mightynimble:
@mreese84 this is ready for review, I moved the selection to the center of the timeline.

mreese84:
@mightynimble Hmm, something isn't working here. Here are the steps I'm doing...

* Perform a search
* Set a temporal
* Click a collection to show timeline
* Scroll timeline to some section that doesn't show the temporal range
* Back to collections
* Select a collection to show timeline

At that point I was expecting the temporal range to be centered in the timeline. However, it is still showing the last place I scrolled the timeline to. Am I missing something?

mightynimble:
It is more complex to fix. Roll it over to next sprint...

mreese84:
So what do do when you said "I moved the selection to the center of the timeline."

mightynimble:
I didn't take it into account the second time you return to granule list of a different collection. The timeline is still the same one the first time initialized. I think I need to refresh some of its variable values. I don't know which ones need update.

mightynimble:
@mreese84 Things to test out besides the cases you mentioned above:

 - Set a temporal range (start AND stop) and click on a collection, verify the selected area is at the center of timeline. Go back and pick another collection, verify again.
 - Set only start time. Verify that the timeline starts at the set start time.
 - Set only stop time. Verify that the timeline ends at the set stop time.

mreese84:
@mightynimble this looks good. Just need to fix the failing test though.

PR #325, opened 2017-01-10
EDSC-1287: GIBS Imagery Fixes

jboukes:
This fix adjusts the filter of function 'match' so that it does not consider time stamps in its date comparisons.


jboukes:
There are two issues addressed by this update.

First - GIBS imagery was being filtered out of some granules due to an error in how dates are filtered.  The fix was to remove the timestamp from the filter comparison.

Secondly - previously it was assumed that all GIBS layers (Arctic, Antarctic, Geo) are available in all projections (of the same name).  This is not the case.  Code was added to check to make sure the layers exist before trying to fetch them.  In addition, in the collection details pane, the available GIBS layers are now shown.

app/assets/javascripts/modules/map/leaflet-plugins/granule_layer.js.coffee
```diff
@@ -170,22 +170,27 @@ ns.GranuleLayer = do (L
...
-        matched = true
 
         # set resolution to {projection}_resolution if it exists
+        # and if the layer exists within optionSet
         oldResolution = optionSet.resolution
-
-        newResolution = optionSet.geo_resolution if this._map.projection == 'geo'
-        newResolution = optionSet.arctic_resolution if this._map.projection == 'arctic'
-        newResolution = optionSet.antarctic_resolution if this._map.projection == 'antarctic'
+        if this._map.projection == 'geo' && optionSet.geo == true
```
mightynimble: ```if this._map.projection == 'geo' && optionSet.geo```

app/assets/javascripts/models/data/collection.js.coffee
```diff
@@ -62,6 +62,7 @@ ns.Collection = do (ko
       @details = @asyncComputed({}, 100, @_computeCollectionDetails, this)
       @detailsLoaded = ko.observable(false)
       @gibs = ko.observable(null)
+      @gibsLayers = ko.observable(null)
```
jboukes: The problem with this is that the available layers are given in the fromJson function.  Since it is dependent upon the JSON data that function receives, I'm not sure how to break it out of that function.

jboukes: I believe I figured out a way to make it better - let me know what you think ;)

app/assets/javascripts/models/data/collection.js.coffee
```diff
@@ -62,6 +62,18 @@ ns.Collection = do (ko
       @details = @asyncComputed({}, 100, @_computeCollectionDetails, this)
       @detailsLoaded = ko.observable(false)
       @gibs = ko.observable(null)
+      
+      @gibsLayers = ko.computed( (->
+        available = []
+        if @gibs()
+          available.push('Antarctic') if @gibs()[0].antarctic
+          available.push('Arctic') if @gibs()[0].arctic
+          available.push('Geo') if @gibs()[0].geo
```
mreese84: Let's change this to 'Geographic' so that it's consistent with our map tools.

mightynimble: I might be wrong but I *think* ```geo``` is from CMR's response. So it is a question of whether we want to be consistent with the map tools or with CMR.

mreese84: I prefer "Geographic". While neither Geo nor Geographic are the official names of the projection, Geographic is what was requested by end users and a customer for the map tool. So it at least carries meaning with the community.

jboukes: 'Geo' is the language used internally; however, for a label, 'Geographic' is a little more recognizable.

I'll make the change.

mightynimble:
The tests look good to me. As long as the build is green, you can merge it.

PR #328, opened 2017-01-18
EDSC-1313 Add checkboxes to facet list.

mightynimble:


mreese84:
3 main things stand out to me.
1) The lag between clicking a box and it actually becoming checked seems a bit extreme. May just be a connection thing, but it seems slow.
2) This is probably out of scope for this ticket, but can you do anything quickly about fixing browse imagery locally? The missing images is maddening.
3) I've attached a screenshot that shows some check boxes are not sized correctly. That definitely needs to be fixed.
screen shot 2017-01-20 at 2 11 00 pm

mreese84:
This all looks good to me now.

PR #333, opened 2017-01-28
EDSC-1326: Initial fix to add new hook to 'View Projects' button

jboukes:


mreese84:
So the project window fires now, which is great. But if you add multiple collections to a project it is not possible to see all of them because the panel is partially covered up by the timeline. I _think_ that Jeff had this issue with the collection details panel and addressed that with this commit: https://github.com/nasa/earthdata-search/pull/244/commits/3d77c18f627ce717ab9f0a6286a8d70cfe360066

app/assets/javascripts/models/ui/project_list.js.coffee
```diff
@@ -89,6 +89,12 @@ ns.ProjectList = do (ko
       @project.focus(collection)
       @configureProject()
 
+    viewProject: () =>
+      base = urlUtil.cleanPath()?.split('?')[0]
+      parameters = urlUtil.cleanPath()?.split('?')[1]
```
mightynimble: I think you can combine these two lines:
```[base, parameters] = urlUtil.cleanPath()?.split('?')```

jboukes:
From what I can tell, 3d77c18 didn't make viewing the added projects all that easier - only a 'little' easier.

Line 1: collectionDetailHeight = @root.find('.master-overlay-content').height()
Line 2: $('.tab-pane .master-overlay-content').height(@root.find('.master-overlay-content').height() / 2.3)

Line 1, above, actually isn't necessary - collectionDetailHeight is never used and so this line can be safely deleted.

Line 2, above, seems to enlarge the scrollable area of the project list pane.  Without this line, there isn't a scroll bar at all - which means you can literally only see the first project in the list.  With this line, there is a scroll bar that you can use to drag the list down to view others - but it is still very, very cramped with the timeline.

Update 3d77c18 also changes the padding of .master-overlay-info but this is a little worrisome as that padding is being set twice (same file) - .master-overlay-info has a padding of 1 em, set here:

.master-overlay-info .panel-list-meta, .panel-list-meta {
  text-align:right;
  padding: 1em;
  margin: 0;
  font-size: .8em;
  color: #666;
  line-height: 1.5;
}

I definitely see that the timeline blocking the project list is an issue - but I don't think this code update caused it, at least based upon what I know right now.  This update only adds a new function to the projectList object as well as a hook from the href to call it - that shouldn't change the CSS in any way.  :\


Suggestion: What if the timeline had a toggle 'show/hide' button?  Or, what if this button toggled *between* the list of projects and the timeline - so you only saw one at a time?  Just a couple of thoughts.

app/assets/javascripts/models/ui/project_list.js.coffee
jboukes: That combine worked great!

PR #335, opened 2017-01-30
EDSC-1067 Use both pump calendar and status app for events.

mightynimble:


mreese84:
Functionality works as expected. Only change requested is to use the styles provided by the status app to be more consistent with other ESDIS apps.

mreese84:
@mightynimble the current notification bar actually overlays on top of the app. I'd rather the div be inserted into the normal flow and bump everything down (instead of overlaying). It should show up in between Tophat the main EDSC toolbar.

mightynimble:
@mreese84 when the notification is dismissed, should the whole page shift up?

mreese84:
@mightynimble, yes. Things should shift up. It would be nice if there was a transition instead of just jumping. So the page would slide up to fill the void.

mightynimble:
@mreese84 should we show one notification at a time or follow the Status App to show them all?

This is how it looks like when you have multiple notifications by default:
screen shot 2017-01-31 at 5 24 18 pm

mightynimble:
Note two things that are different from the default Status styles:
1) The close/dismiss button is floated to the right instead of appending to the message.
2) Title (if exists), start and end date are prepended to the message. Status only displays the message. It doesn't support title field (should they support this field if it is designed to replace PUMP calendar?). Nor does it displays start and end time of the notification.

PR #345, opened 2017-02-22
EDSC-1257 Select multiple facets. Fix failed specs and update fixtures.

mightynimble:


mreese84:
@mightynimble this seems to be working except for the keywords facet. If I select a top level keyword I'm only able to drill down further into that keyword. Take a look at collection C1200056431-MMT_1 (in sit). It has keywords ATMOSPHERE > ATMOSPHERIC WINDS > WIND CHILL and CLIMATE INDICATORS > LAND SURFACE/AGRICULTURE INDICATORS > LANDSLIDES

I can only select one or the other.

mreese84:
@mightynimble let's look at this together tomorrow. Just want to make sure I understand whats going on.

PR #349, opened 2017-02-27
EDSC-1352: Initial fix for multiple collections having appropriate options in data access

jboukes:


mreese84:
@jboukes, if i select collection C179014698-NSIDC_ECS, select the AE_5DSno.2 ESI Service for my first access option, add a second one and select the same option, the form still does not populate on the screen. I'm using Chrome.

jboukes:
@mreese84 - the way the app handles suboptions forms is a little misleading, I believe.  From testing with that collection, it looks like both access options were actually accepted.  The form for the second access options actually writes over the first form, which gives the illusion that the first form never left.

I just tried with C179014698-NSIDC_ECS with the AE_5DSno.2 ESI Service - in the first form, I entered 'Burt@raytheon.com' as the email entry, then added a second access option (which makes the Burt@raytheon.com entry disappear).  In the second access options, I entered 'Ernie@Raytheon.com', and then went on to hit continue and Submit.

After hitting Submit, I could see that C179014698-NSIDC_ECS was listed twice - once for each access form entry.  This feels correct.

Within the markup conversation in the output, I found a block involving "\u003cecs:email\u003e**Ernie**@raytheon.com\u003c/" as well as a block for "\u003cecs:email\u003e**Burt**@raytheon.com\u003c/ecs:email\u003e\n ".  This seems to prove that both access methods are actually being accepted - just the way the form writes over previous forms when creating additional access methods makes it look like it's not working.

Let me know if this explains it, or if there might be something wrong that I'm not getting (happens).

mreese84:
So I guess my expectation (which should have been more clearly stated admittedly), is that a new set of fields would be rendered. Think of it from a user's perspective, they'd have to be paying really close attention and then they'd then be just making an assumption that both forms are being submitted. It would be better and instill more confidence seeing both sets of fields.

jboukes:
@mreese84 -  I agree the way it is, is confusing.  I can look at improving it as you described.

Should we preserve the suboptions entry forms for all collections in the project, or should we do it only when there are 'additional access options' on a single collection?  I would assume the former.

Not sure how hard/easy this change will be - will have to dig in a bit.

For the moment though - does it pass the original bug report for you?  The original are collections ' MOD10A1 V005, MYD10C2 V006, and AE_L2A V003' - if you could confirm those at least are working, then we know we at least have the core solution for the acceptance criteria.  Prior to this PR, a project with those three collections was truly broken - you couldn't access sub-options for the second and third collection at all.

I'll start looking into how to preserve suboptions field forms for multiple entries and will report back when I know more. ;)

mreese84:
@jboukes, this is getting closer. A couple of things I noticed in the latest review.

1) I do see multiple forms rendering on the page, which is great.
2) On the first form, when selecting various bands for subsetting I see it change from something like 4 of 4 selected to 1 of 4 selected. (screenshot attached). However, on subsequent forms when selecting bands nothing is changing in the "x of x selected" section. We need to ensure that the prompt is getting updated, but more importantly that those bands are being sent in the request.
3) And finally, if you add multiple collections into a project, select an ESI form for the first collection, set some filters and click continue... selecting ESI on the subsequent forms does not render a new form. We need to handle that too.

Screenshot of band subsetting message issue.
First ESI Form
screen shot 2017-03-03 at 10 53 20 am
Subsequent ESI Forms
screen shot 2017-03-03 at 10 53 30 am

jboukes:
@mreese84 I will take a look at 2.); however 3.) _should_ be working.  

I am able to add multiple collections and select ESI (or any other suboption), complete the form, hit next, and select ESI on the second form, hit next, and hit an ESI for the third form, too.  The rspec test that's included also does this (it selects an ESI on the second and third collection, though it does do FtpPushPull on the first).  

Somehow you found an exception that doesn't work - what collection are you using to test?

mreese84:
Hmm, let me take another stab at recreating.

jboukes:
@mreese84 thanks!  I did try the other issue and it's definitely a bug.  It looks like it is likely an Echoforms issue.  The label is actually handled by ID (i.e., "<div id="bands-count"><span id="selected-bands-count">6 of 6</span> bands selected</div>"), which are supposed to be unique on the page.  Since the fix allowed for more than one form to be present on the page at once (in the case that they were added by 'Add Access Option'), then this means you have two labels with the same ID if you use two ESI methods, and so when you interact with the first one, it actually updates the first.  

Now to see if that can be fixed...

mreese84:
Ok, disregard my third comment. I must have completely spaced while testing that. It's working for me.

Just need to fix #2.

spec/features/data_access/data_access_workflow_spec.rb
```diff
@@ -8,6 +8,62 @@
...
+
+    context "when clicking 'FtpPushPull'on the first collection" do
+      before(:all) do
+        within '.access-item-selection:nth-child(1)' do
+          choose 'FtpPushPull'
+          wait_for_xhr
+        end
+      end
+      it "displays the distribution options for the first collection" do
+        page.find('.access-form').should have_content("Distribution Options", wait: 5)
```
mightynimble: Why ```wait: 5``` here? I know that the ESI form will take some time to load but shouldn't ```wait_for_xhr``` be sufficient or am I missing something?

jboukes: Wait_for_xhr just wasn't working in in this case.  As you likely know, it's an in-house function and it does seem to work in most cases.  In this case, it wasn't working at all.

I found 'should' waits for elements to show up, up until the default wait time set in capybara, which is 2.  At the default wait time the test was passing about 1 out of 3.  Adding the 'wait' parameter uses that time instead of the default wait time, and increased it to 5.  It never failed at 5, and so I stuck with it.

I'll revisit it today and see if I can learn more why wait_for_xhr doesn't work, and maybe to look at changing it to 'expect' rather than 'should'.  I would also like to get rid of that one 'sleep'  - made me feel guilty to use it (as it should).

mightynimble: Search for ```synchronize do``` in ```*_spec.rb``` you will see examples that wait for much longer time (30 seconds).

The method is defined in ```earthdata-search/spec/helpers/page_helpers.rb```

jboukes: @mightynimble - you were right to question those waits, as after running the test more times I have managed instances where it fails.  :(

I tried using 'expects' insteads of 'shoulds' - still fails at times.  I tried using synchronize - even synchronize(60), and it still intermittently fails - taking the full minute to fail.  A method that does not occasionally fail is being elusive.  For some reason the form generated by echoforms is not showing up within any reasonable time, on occasion.

Could the fact that it is waiting for a submodule to return a form have anything to do with it?  Other ajax items that are loaded from within the immediate scope of edsc seem to be fine after using wait_for_xhr.  But elements loaded from echoforms - at least anecdotally (so far) - seem to be a little flaky and wait_for_xhr has been no solution at all.  

I'll keep throwing rocks at it and see if I can find something that consistently works.  Suggestions welcome.

app/assets/javascripts/models/data/service_options.js.coffee
```diff
@@ -68,7 +68,11 @@ ns.ServiceOptions = do (ko, edsc = @edsc, KnockoutModel = @edsc.models.KnockoutM
         clickedMethod = m
         break
 
+
       echoformContainer = if @index then document.getElementsByClassName('access-form')[@index] else document.getElementsByClassName('access-form')[0]
+      if (typeof echoformContainer != 'undefined')
```
mightynimble: ```echoformContainer.empty?() if echoformContainer?```

jboukes: Thanks!  

This is repairing a bug where the form generation was really flaky.  There were instances where you could keep pushing 'FtpPushPull' and the form would be generated with every button push so you could have dozens of forms.

This fixes it.

app/assets/javascripts/models/data/project.js.coffee
```diff
@@ -57,7 +57,7 @@ ns.Project = do (ko,
     constructor: (@collection, @meta={}) ->
       @collection.reference()
       @meta.color ?= colorPool.next()
-
+
```
macrouch: don't add nothing to commits!

jboukes: It was a deletion of a previous change that was committed.  Thanks!

jboukes:
I finally found why the rspec test was being 'flaky'.

The issue was the inclusion of this line "@index = document.getElementsByClassName('access-form').length" - this was meant to establish the 'index' of the ServiceOptions object.  Since the access-form divs are generated as the Service Options are created, it made sense.  Worked live well enough, too.

It also worked with the rspec test perfectly - _the first time it was ran_.  The issue is the created 'access-forms' were somehow caching, so the next time the test was ran, @index would count instances that it didn't see the first time it ran, and so would given the wrong value, and then the form gets written to the wrong div.

The fix (as I see it) is to go back to having Project Collection keep track of the index and pass it down to its children in a more governed manner.  The test seems very stable after doing this - great! - but it's also broken Mark's interest in being able to add access options on a single collection.  That is again an issue.

So, work in progress label is on, last bug is solved, need to fix the issue of keeping multiple forms if additional accesses are added, then _should be_ done.  Will remove label and ping you all when that's complete.  Thanks.

jboukes:
@mreese84 @macrouch @mightynimble give it another shot.  Completely different strategy (plan 'C') - though I like it more than the others, actually.  Test works consistently and live seems to do everything we want.

Now just to adjust that Echoforms pr...

PR #354, opened 2017-03-01
EDSC-1357 show all contacts from both DataCenters and ContactPersons.

mightynimble:


jboukes:
@mightynimble - is there a good collection to try this out on?

mightynimble:
The collection is mentioned in the ticket. I think the short name is NSIDCV0

PR #355, opened 2017-03-01
EDSC-1350 Limit number of granules of AST_* orders to 2000.

mightynimble:


mreese84:
Tried testing locally but my searches aren't working. Will try again in the morning.

config/services.yml
```diff
@@ -28,6 +28,15 @@ earthdata:
 status_app_root: "https://status.earthdata.nasa.gov/api/v1/"
 cwic_root: "http://cwic.wgiss.ceos.org/"
 
+#Earth Data Search Client specific configs
```
mreese84: Should be "Earthdata". Also, let's explain a little better what's going on here. The collection concept IDs included below will be limited to 2000 granules when ordering. Would hurt to also make a note where that limit is set.

config/services.yml
```diff
@@ -28,6 +28,15 @@ earthdata:
...
+edsc:
+  sit:
+    aster_collections: C1002-LPDAAC_TBD
```
mreese84: Please change these labels to "limited_collections"

PR #360, opened 2017-03-03
EDSC-1370 Lock coffee-loader and webpack versions.

mightynimble:


mightynimble:
@mreese84 I have successfully deployed this to SIT.

PR #364, opened 2017-03-07
EDSC-1385 handle metadata values not being present in temporal details

macrouch:


mightynimble:
Approved assuming the build is green and the test expectation is correct.

PR #365, opened 2017-03-07
EDSC-1382 update UMM version to 1.9 for collection details page

macrouch:


macrouch:
I need to get my tests working locally so I can update this branch with new cassettes

macrouch:
Do you know of an easy way to remove them?

mightynimble:
Not a very easy way. I'd try to revert the fixture changes and remove all v1_4s. I think they should be all next to each other. It doesn't hurt to delete more than we should - they will be added back when running the tests. This is what I did for all the .echo10 requests/responses.

PR #370, opened 2017-03-14
EDSC-1392 Use https to prevent a redirection from http to https.

mightynimble:


mightynimble:
@mreese84 removed you from the reviewer since the PR may need more work.

mightynimble:
OK, @mreese84 , it is ready

PR #373, opened 2017-03-15
EDSC-1335 display related urls on collection details page, rubocop fixes

macrouch:


mreese84:
@macrouch, a couple things I noticed on first review.
- Check out C61468234-NSIDC_ECS. There's a "View More" link with 3 empty `li` elements. This is the case with all of the NSIDC collections I tested. Sometimes it's more than 3 empty elements.
- I think the link should say "View All Related URLs" so that it's super clear.
- I was looking for a record that had some of the 4 types we were going to show but couldn't find any. Then I realized that MMT doesn't have Related URL yet. Were you testing with any collection(s) in particular?

macrouch:
@mreese84 are you using `cmr_env=sit`?

mreese84:
@macrouch, I am... http://localhost:3000/search/collection-details?cmr_env=sit&p=C1000001160-NSIDC_ECS&fdc=NSIDC

macrouch:
@mreese84, I'm not sure. What you described sounded like what I saw with ops data. I'll look first thing tomorrow

mreese84:
This MMT collection has a URL populated in the view more modal: C1200230699-MMT_1

Make sure you looking at CMR SIT and logged in.

mreese84:
One more thing, these links are being appended onto localhost:3000/search. So what ends up being linked is "localhost:3000/search/www.nasa.gov", which of course isn't going to work.

macrouch:
@mreese84 I can't find C61468234-NSIDC_ECS or C1000001160-NSIDC_ECS on SIT

I did see C1200230699-MMT_1 on SIT, and that URL problem is the same problem MMT was dealing with. I'm beginning to think MMT should require a full URL with the protocol included (http, https, etc.). I'll look at MMT to see how they fixed it.

I tested with C1000000577-DEV07, almost every collection I clicked on looked very similar with their links.

app/views/search/_collection_details.html.erb
```diff
@@ -137,20 +146,20 @@
...
             <dd>Not Available</dd>
-            <!--/ko-->
-            <!--ko if: $data.science_keywords.length > 0 -->
+            <!-- /ko -->
+            <!-- ko if: $data.science_keywords.length > 0 -->
             <dd data-bind="foreach: $data.science_keywords">
               <div class="arrow-tag-group">
                 <ul class="arrow-tag-group-list">
-                  <!--ko foreach: $data-->
+                  <!-- ko foreach: $data -->
```
mightynimble: move ```foreach``` binding into ```<ul>``` tag

app/views/search/_collection_details.html.erb
```diff
@@ -137,20 +146,20 @@
         <div class="row">
           <dl>
             <dt>Science Keywords:</dt>
-            <!--ko if: $data.science_keywords.length == 0 -->
+            <!-- ko if: $data.science_keywords.length == 0 -->
```
mightynimble: And here (and one more place below when the length > 0)

mightynimble:
screen shot 2017-03-21 at 3 08 34 pm

app/views/search/_collection_details.html.erb
```diff
@@ -94,16 +94,25 @@
               <!-- /ko -->
             </dd>
           </dl>
+          <!-- end of second row -->
+
+          <ul>
+            <!-- ko foreach: $data.highlighted_urls -->
+            <li><a data-bind="attr: {href: $data.url}, text: $data.name" target="_blank"></a></li>
+            <!-- /ko -->
+            <!-- ko if: $data.related_urls.length > 0 -->
```
macrouch: If I do this here then there will be an empty `<li>` element when there are no related urls. I'm also just not sure why, the file uses `<!-- ko if .... -->` all throughout

mightynimble: I happened to read a stackoverflow Q/A that ```<!-- ko``` can be used for 'virtual elements'. But it doesn't say that it is a convention to move that into the data-bind attr. It is just my interpretation :)

mreese84:
@macrouch this is looking good to me. The only thing I think we need to do now is give the related urls a section title (like Metadata Formats and API Endpoints sections) and make the formatting match (for both the section title and the related url links). Notice that the links under metadata formats and API endpoints are bolded.

PR #379, opened 2017-03-23
EDSC-1296 Add metric to capture searches which result in user directly entering another search

mightynimble:


db/schema.rb
```diff
@@ -11,15 +11,15 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160920144648) do
+ActiveRecord::Schema.define(version: 20170316054224) do
```
macrouch: shouldn't there be a migration for this?

app/controllers/search_controller.rb
```diff
@@ -5,6 +5,9 @@ class SearchController < ApplicationController
   respond_to :json
 
   def extract_filters
+    isImmediateReentered = (params.delete :rerun) == 'true'
```
macrouch: this variable should be named `is_immediate_reentered`, along with `previous_keyword` below

db/schema.rb
mightynimble: It is the migration that John added the other day. I think he just forgot to update the schema.rb.

macrouch: A migration was added?

mightynimble: https://github.com/nasa/earthdata-search/pull/372

macrouch: Ah, yep. Merged before I saw it

PR #381, opened 2017-03-31
EDSC-1337 add ability to submit service requests for more than 2000 g…

mightynimble:
…ranules


mreese84:
@mightynimble, now that ESI orders support order chunking, are there any other access methods that limit to 2,000 granules? If no, we can remove that red banner that says some orders may be limited to the first 2,000 granules.

mightynimble:
@mreese84 the message is removed.

PR #382, opened 2017-04-04
EDSC-1396 Auto pan to point or bounding rectangle for manually entere…

mightynimble:
…d spatial constraints.


mreese84:
@mightynimble, this looks much better. I like the panning of the bounding rectangle, but I think the point search panning/zooming could be improved. As it currently stands, the point pans so that it falls right on the edge of the collection panel. Can we tweak it so that it falls in the middle of the viewable area of the map? Screenshot attached...

screen shot 2017-04-06 at 2 44 06 pm

mightynimble:
Please take another look.

mreese84:
@mightynimble, I'm not sure why, but the first time I tested bounding rectangle it panned and zoomed to an open area. However, now it's not doing that. We need to make sure that the bounding rectangle manual search behaves like the point search (panning/zooming to the open area). The point search works great now.

mightynimble:
@mreese84 I couldn't find a way to correctly/properly set the zoom level on finish drawing the rectangle. ```fitBounds``` can do that but it is not what we exactly want...

PR #391, opened 2017-04-17
EDSC-1303: Updated header message for data access and modified visib

jboukes:
…lity to hide it at appropriate time

The data access header 'Success' message was changed to: 

_Request in Progress:  Your request is being processed.  Results will appear shortly._

This message disappears once the results have been obtained.

Do we need a test for this?  It feels so innocuous and tiny as to not need one - let me know.

### DIFF
--- a/app/views/data_access/retrieval.html.erb
+++ b/app/views/data_access/retrieval.html.erb
@@ -14,9 +14,9 @@
   <nav class="data-access-nav">
     <a href="<%=edsc_path "/data/configure?#{@retrieval.source}" %>"><i class="fa fa-arrow-circle-o-left"></i> Back to Data Access Options</a>
   </nav>
-  <header class="data-access-header">
-    <h1>Success!</h1>
-    <p>Your request has been processed.  See below for information on accessing your data.</p>
+  <header class="data-access-header" data-bind="visible: ui.projectList.collectionsToDownload().length == 0">
+    <h1>Request in Progress</h1>
+    <p>Your request is being processed.  Results will appear shortly.</p>
   </header>
   <div class="data-access-container">
     <section class="data-access-content data-access-downloads" data-bind="visible: ui.projectList.collectionsToDownload().length > 0">
--- a/app/views/data_access/retrieval.html.erb
+++ b/app/views/data_access/retrieval.html.erb
@@ -16,7 +16,7 @@
   </nav>
   <header class="data-access-header" data-bind="visible: ui.projectList.awaitingStatus()">
     <h1>Request in Progress</h1>
-    <p>Your request is being processed.  Please standby.</p>
+    <p>Your request is being processed.  Please standby for status.</p>
   </header>
   <div class="data-access-container">
     <section class="data-access-content data-access-downloads" data-bind="visible: ui.projectList.collectionsToDownload().length > 0">
--- a/app/views/data_access/retrieval.html.erb
+++ b/app/views/data_access/retrieval.html.erb
@@ -14,9 +14,9 @@
   <nav class="data-access-nav">
     <a href="<%=edsc_path "/data/configure?#{@retrieval.source}" %>"><i class="fa fa-arrow-circle-o-left"></i> Back to Data Access Options</a>
   </nav>
-  <header class="data-access-header" data-bind="visible: ui.projectList.collectionsToDownload().length == 0">
+  <header class="data-access-header" data-bind="visible: ui.projectList.collectionsToDownload().length == 0 && ui.projectList.collectionOnly().length == 0 && ui.projectList.submittedOrders().length == 0 && ui.projectList.submittedServiceOrders().length == 0 && ui.projectList.collectionLinks().length == 0">
     <h1>Request in Progress</h1>
-    <p>Your request is being processed.  Results will appear shortly.</p>
+    <p>Your request is being processed.  Please standby.</p>
   </header>
   <div class="data-access-container">
     <section class="data-access-content data-access-downloads" data-bind="visible: ui.projectList.collectionsToDownload().length > 0">
--- a/app/views/data_access/retrieval.html.erb
+++ b/app/views/data_access/retrieval.html.erb
@@ -14,7 +14,7 @@
   <nav class="data-access-nav">
     <a href="<%=edsc_path "/data/configure?#{@retrieval.source}" %>"><i class="fa fa-arrow-circle-o-left"></i> Back to Data Access Options</a>
   </nav>
-  <header class="data-access-header" data-bind="visible: ui.projectList.collectionsToDownload().length == 0 && ui.projectList.collectionOnly().length == 0 && ui.projectList.submittedOrders().length == 0 && ui.projectList.submittedServiceOrders().length == 0 && ui.projectList.collectionLinks().length == 0">
+  <header class="data-access-header" data-bind="visible: ui.projectList.awaitingStatus()">
     <h1>Request in Progress</h1>
     <p>Your request is being processed.  Please standby.</p>
   </header>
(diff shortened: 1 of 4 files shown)


app/views/data_access/retrieval.html.erb
```diff
@@ -14,9 +14,9 @@
   <nav class="data-access-nav">
     <a href="<%=edsc_path "/data/configure?#{@retrieval.source}" %>"><i class="fa fa-arrow-circle-o-left"></i> Back to Data Access Options</a>
   </nav>
-  <header class="data-access-header">
-    <h1>Success!</h1>
-    <p>Your request has been processed.  See below for information on accessing your data.</p>
+  <header class="data-access-header" data-bind="visible: ui.projectList.collectionsToDownload().length == 0">
+    <h1>Request in Progress</h1>
+    <p>Your request is being processed.  Results will appear shortly.</p>
```
mreese84: Personally, I would remove the "Results will appear shortly" sentence. We don't know how long it will take in some cases. There is an effort underway by our underlying services to provide us with a time estimate, so that would be something we could tackle when that is in place.

mreese84:
One small comment added. And yes, I would test that the message is removed from the screen when the order is ready.

mreese84:
@jboukes when the message is removed, the data access container needs some padding at the top so it doesn't butt right up against the "back to data access options" element.

mreese84:
@jboukes, looks like the message needs to be removed for any non-pending state for ESI orders. Here's a screenshot of a successful ESI request that still has the message.

screen shot 2017-04-18 at 10 22 45 am

app/views/data_access/retrieval.html.erb
jboukes: How about a nice "Please standby." instead.  It's not as promising as "shortly" but still conveys there's a wait ahead.

Rspec is throwing a weird error on this - I'll have to sort that out.

jboukes: For some reason this silly github won't let me reply to each of your comments :)

Nice catch on the instances of it not disappearing; it must be logically based upon only the 'success' block.  I can look at that as well as adding a bit of padding for the data access container.

I'll put a WIP flag on this and then rework it a bit.

jboukes: @mreese84 Alright, have new changes in - let me know what you think.  I added a bit of padding so that the data-access isn't riding up on the back to collections links, updated the text to a more less-promising 'Please standby.', and also found other visibility cases to hide the header for when data-access-container has content.

I checked and it does now work for those ESI collections (I used the same collection you did), but I couldn't find one that'd fail.  Can you

mreese84:
@jboukes, Bo is waiting on you to add a test before reviewing the code. Please knock that out for us. Thanks!

mreese84:
@mightynimble, can you make sure you review this first thing Thursday morning. If approved, go ahead and merge it as well.

mightynimble:
What is the expected behavior? Should "Request in Progress" show up as long as the order/service state is not "complete"? Should it change to "Success" when the order or service is done successfully?

mreese84:
The expected behavior is that the request in progress message should show up as long as it is not "complete". Once it is complete, that top header is redundant because you get the main content in the window below (be it a success or failure notice). So we are just removing it from the page. It's not needed at all.

app/views/data_access/retrieval.html.erb
```diff
@@ -14,9 +14,9 @@
   <nav class="data-access-nav">
     <a href="<%=edsc_path "/data/configure?#{@retrieval.source}" %>"><i class="fa fa-arrow-circle-o-left"></i> Back to Data Access Options</a>
   </nav>
-  <header class="data-access-header">
-    <h1>Success!</h1>
-    <p>Your request has been processed.  See below for information on accessing your data.</p>
+  <header class="data-access-header" data-bind="visible: ui.projectList.collectionsToDownload().length == 0 && ui.projectList.collectionOnly().length == 0 && ui.projectList.submittedOrders().length == 0 && ui.projectList.submittedServiceOrders().length == 0 && ui.projectList.collectionLinks().length == 0">
```
mightynimble: These conditions won't guarantee that the header will show up before an order or a service order reaches 'complete' state. For example, submit an "FtpPushPull" order using a granule from C204690560-LAADS will hide the header when the order is in "Submitting" state.

It might be cleaner to move these conditions into a function.

spec/features/data_access/data_download_spec.rb
```diff
@@ -209,6 +209,9 @@
       wait_for_xhr
     end
 
+    it 'does not display a header erroneously declaring success' do
+      expect(page).should_not have_content("Success!")
+    end
```
mightynimble: Need to test the cases when "Request in Progress" shows up and when "Request in Progress" disappears. 
 - an order in different states ("Creating" - show, "Submitting" - show, "Complete" - hide, etc.)
 - a service order in different states (similar, see above.)

app/views/data_access/retrieval.html.erb
jboukes: @mark, @bo,I think this is actually okay - to further clarify what Mark had said, the initial request in the ticket was to just remove the header entirely.  It was useless information because it said 'Success!' no matter what happened.

Removing it was a bit unsightly - when the page was awaiting the state, we just had an empty page, so that was actually no good.  *Something* needed to be there, and so I thought replacing the header with a vague 'Please standby' type message would help inform the user that yes, something was happening and it should be ready sometime soon.

Once the data-access-container has the 'real' status the header should disappear, regardless of what that 'real' status is.  So it doesn't need to be truly complete for it to disappear - just data-access-container needs to be complete in getting filled with information.

So, if it says it's in a submitting state, it's actually fine to hide the header as the data-access-container is clearly stating the status at that point.

At least that was the mindset I used in making this change ;)

spec/features/data_access/data_download_spec.rb
jboukes: I'll take a shot at it ;)

app/views/data_access/retrieval.html.erb
jboukes: @mightynimble - Added new function to tidy it up, I like it much better.

jboukes:
@mightynimble, @mreese84 - please take a look at the new test.  It does check that the 'Please standby' message is _not_ present after the status has loaded, which is good.

However, finding a way to test that the message appears at all is not apparent.  The message disappears too rapidly.  I tried something like the below, for example, and it does not work.  I have the check immediately after the submit and it still doesn't work.

Is there a way to do this that I'm not seeing?

```  
context "when retrieving data for download" do
  before :all do
      load_page 'data/configure', {project: [downloadable_collection_id, non_downloadable_collection_id], temporal: ['2014-07-10T00:00:00Z', '2014-07-10T03:59:59Z']}
      wait_for_xhr
      choose 'Download'
      click_on 'Continue'
    end

    it 'displays a header instructing the user to standby for status' do
     click_on 'Submit'
      expect(page).to have_content("Your request is being processed.  Please standby for status.")
    end
end
```

app/views/data_access/retrieval.html.erb
mightynimble: @mreese84 's call. If the 'stand by' message should disappear as long as there is a state for the order, I think we need to make sure that it does disappear when the order is in submitting/creating/etc. states. Otherwise, we need to make sure that the message is on the page when the order is not in 'complete' state.

jboukes: @mightynimble I agree that it needs checking whether it disappears for those states - I'll get that done tomorrow.

mreese84:
This looks fine to me. I think we might be over-thinking this particular issue.

PR #392, opened 2017-04-18
EDSC-1076 Add checkbox to filter collections that don't have granules

mightynimble:


mreese84:
@mightynimble, the checkbox should be moved down below the tip so that it's on its own line.

Also, I'm seeing a discrepancy between the number of collections locally vs what is in prod. Even when I set my cmr_env to prod. Are you seeing the same numbers?

mreese84:
@mightynimble, fixtures need to be updated for this PR.

mreese84:
@mightynimble, in general... let's make Thursday morning "clean up" time. So getting this PR green should be prioritized.

PR #401, opened 2017-04-26
EDSC-1419 Bring back 2d coordinate search

mightynimble:


mreese84:
@mightynimble, upon initial view of the filter stack work... I'd add a little padding between the end of the value input field and the "x" icon. I'd also make the color of the help text grey so it blends in a little. Seems too harsh/in-your-face as it is. Maybe even bump the font size down a size or two.

mreese84:
Also, looks like there's a merge conflict that needs to get knocked out.

mreese84:
@mightynimble, also noticed that the value input field is so long that it is stretching out the "clear filters" icon box. Notice in the screenshot that the clear filters button is longer than the other 2 icons. We should shorten the value input field so that it doesn't stretch out the icon.

screen shot 2017-05-01 at 9 32 11 am

app/assets/stylesheets/modules/map/map.css.scss
```diff
@@ -306,6 +306,32 @@ p.map-popup-info {
...
+    height: 22px;
+  }
+
+  label {
+    min-width: 0;
+  }
+}
+
+.landing-toolbar-content .error {
+  max-width: 590px;
```
mightynimble: This is more like a hack to me - is there a place that defines the max width of the filter stack? @mreese84

mightynimble:
Pushed css changes that you mentioned, @mreese84 . Also see my comment above.

mightynimble:
Right now I am trying to find a way to style the 2d coord input to fill the remaining space of the container. If you know something off the top of your head, I am very appreciated to learn :)

screen shot 2017-05-01 at 10 42 22 am

PR #403, opened 2017-04-27
EDSC-1449: Updated map_zooming test as needed for new UI

jboukes:
Granule_filters was fixed as a bonus.  These changes were tested on Travis by restarting the build six times; the build went green every time.

This PR replaces the 'FlakyTestFinder' that I was using as a sandbox.

### DIFF
--- a/app/assets/javascripts/modules/map/map.js.coffee
+++ b/app/assets/javascripts/modules/map/map.js.coffee
@@ -45,20 +45,9 @@ ns.Map = do (window,
         @_zoom = @_limitZoom(zoom)
         return this
 
-      # if overlay is visible, adjust center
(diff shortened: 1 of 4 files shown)


mreese84:
@jboukes, so while the tests are passing, I think they might be testing the wrong thing. Please notice the acceptance criteria on the ticket:

> All map zooming tests pass and **when zooming in and out, the center of the map stays in place.**

To test this in a browser, draw a bounding box around the center of the map, then zoom in/out. Now... compare that to how Google Maps behaves. Much different. When our collection panel moved to the bottom, that righthand portion that used to be the viewable area became invalid. So the middle of the map has changed.

This ticket is to resolve both the zooming issue AND the corresponding tests.

jboukes:
Got it!  I'll look into getting that part settled as well.

jboukes:
@mreese84 - give this update a shot, thanks!

PR #406, opened 2017-04-28
EDSC-1076 Fix an invalid test that fails on master.

mightynimble:


mightynimble:
@jsiarto @mreese84 @jboukes - need a quick review on the latest commit when it is green. It should fix the failed master.

PR #407, opened 2017-05-01
EDSC-1284: Update access method titles.

mreese84:


mreese84:
@mightynimble, I'm having issues with this one last test. Can you take a look this morning? I'd like to get this merged this morning.

mightynimble:
Some CSS issues. See the screenshot:
screen shot 2017-05-02 at 1 23 04 pm

 - I only see three borders (add some padding to teh bottom?).
 - when selecting one of the access methods, the page below the green line shifts down a few pixels.
 - See this screenshot taken from a test: when there are more than one order options, user won't know which one is needed. This is very rare and I couldn't find an example in real data.
screen shot 2017-05-02 at 12 36 51 pm

mreese84:
@mightynimble, the bottom border is missing on purpose as it's supposed to resemble a tab and butt right up against the element below it. Looks fine in my browser. The design accommodates 4 access methods, which is the most I've seen even in our test environments. We'll roll with this and see out it performs. Thanks for your help.

PR #408, opened 2017-05-02
EDSC-1447: Fixed Shapefile dropzone interface

jsiarto:


mightynimble:
The failed one could be caused by tophat:
screen shot 2017-05-03 at 11 14 49 am

PR #410, opened 2017-05-03
EDSC-1441 Remove calendar events from EDSC

mightynimble:


jboukes:
Code looks reasonable to me.  ;)

mreese84:
@mightynimble, I'm currently waiting on approval for Status app access to test this fully. I want to make sure EDSC is just straight using the javascript widget and not the API.

mightynimble:
We are using the API, not the javascript widget. Because we need to be albe to
 - show/hide the banner
 - show one environment's banners in another. (with cmr_env query param)

mreese84:
@mightynimble, the status app has a "dismissable" flag that will allow users to dismiss the notification. And the cmr_env parameter wouldn't work because the Status app isn't tied to the CMR.

mreese84:
This is no longer needed. It was taken care of in #417

PR #411, opened 2017-05-03
EDSC-1337 Use correct production env name.

mightynimble:


mightynimble:
cmr_env can have value "sit", "uat", "prod"
if it is "prod", it needs to be converted to "production", since "prod" is not a rails env name but "production" is.
It wasn't noticed until the code went to ops where it failed deployment because only production has this naming difference.

PR #415, opened 2017-05-07
EDSC-1395: New Site Tour

jboukes:


jboukes:
First comment - this PR needs an rspec test!  

I'll start that Monday.

app/assets/javascripts/models/data/preferences.js.coffee
```diff
@@ -24,11 +25,16 @@ ns.Preferences = do (ko
         @fromJson(data)
         @isLoaded(true)
       else
-        getJSON '/users/site_preferences', (data, status, xhr) =>
-          console.log "Loaded site preferences, #{JSON.stringify(data)}"
-
-          @fromJson(data)
-          @isLoaded(true)
+        $.ajax '/users/site_preferences',
```
jboukes: The reason for this change, just to proactively explain it, is that there was a race condition with the instantiation of the preferences object.  When preferences is instantiated it is often immediately checked; however, this can get an erroneous result  as the constructor has not completed the getJSON call yet (asynchronous activity).  I had to switch this call to an $.ajax so that I could set async to false for this single call, meaning it will wait to populate itself during instantiation so that when it's immediately called, it has the right goods.

mreese84:
One more comment... the initial dialog is still not in the center:
screen shot 2017-05-08 at 8 07 04 am

jboukes:
@mreese84: Fixed the checkbox so that it's a normal single-click, and also changed the initialized state of the doNotShareTourAgain observable.  So, only one click to change the checkbox and the tour should show from the start.

Oddly centering the first box in the page is being more difficult than those two.  Currently it is attached to an element - and that element is #map-center - and doing that gets it kind of close (as your image shows).  Bootstrap popovers seem to love being attached to elements - I'm researching how to get it to be straight up the center of the page, and will report back.

mreese84:
@jboukes, thanks for the update. Re: initial box... who says it has to be a popover? I'd be okay with using a modal for now since this is such a drastic change.

jboukes:
@mreese84 - I suppose it's me saying that it needs to be a popover, lol.  Programmatically, using popovers for this tour is just really fluid and works well.  The tour is essentially an array of popovers that are ticked through by index, set by the next and prev buttons (this is based upon how the old tour went).  Tucking a red-headed step child of a modal into that process feels programmatically awkward - it can be done, of course, I am sure - just awkwardly.

Let me poke at getting that first one centered; if worse comes to worse, I can figure out how to squeeze a modal in there.

Further note - we have new test failures, and I believe this is due to fixing the issue where the tour was not showing up by default.  Now that it is showing up, it's breaking some tests.  This is my guess, in any case.

mreese84:
@jboukes, ok. Do what you can to get the first box centered up, but don't kill a full day doing it or anything. It can't be _that_ bad to make the first box a modal 😏

mreese84:
@jboukes I really would prefer to use a modal the more I think about it. There are a few reasons:

1. I think it adds more visual weight to the initial CTA
2. It centers automatically
3. It closes with the escape button
4. If you resize the browser window, it remains centered
5. I think that manually re-positioning the initial window is hacky

What are the downsides of using the modal for that initial dialog?

jboukes:
If you really would prefer, then I'll be happy to add it.

Reasons for keeping it a popup:

Consistency - it feels awkward in a UI sense to have a modal mixed with popups, and so having it be all popups feels consistent.  A modal will have at least a slightly different look and feel and so there is this native inconsistency between using the two together.

Required User Interaction - A modal must be dealt with while a popup and be somewhat ignored.  The priority of the tour should not be such that it dominate the page and insist on interaction (imo).

All of the other popups can also be closed with the escape button if I code them to do that :)

`$(document).keyup(function (event) {
    if (event.which === 27) {
        $('#example').popover('hide');
    }
});`

And lastly, if you think the manual repositioning of the initial window is hacky - well, so will the injection of a modal into this process.  I'm not saying it's 'hard' - just hacky.  The tour is an index of popovers that are ticked through.  Having the first item be a modal means that index 1 needs to have special code that says 'wait!  do a modal here instead', and that the 'Next' button then leads to the actual first popup in the array.  The first popup in the array will have to be 'forced' with a special 'Previous' button that will re-open the modal, as that first popup will not really be the first dialogue anymore.  Again, not hard - but squirrelly and hacky (or 'awkward' as I like to say).

BUT - I can tell you really want a modal, so I'll do that.  👍

mreese84:
😄 I do really want the modal. I guess I'm viewing that first dialog not really as part of the tour. In fact, the main CTA in that dialog shouldn't be "Next". It should be "Start Tour". Just noticed that.

jboukes:
You got it, buddy.  👍  I'll get something together by tomorrow to show (I will be out the rest of the day starting pretty much now) - I did setup a 'stub' modal injected into the process just to get started and to see what the tests do.  The rspec tests are screwy - 127 failures - but fixing them should just be a matter of copying and pasting a line to close the modal in each one, which shouldn't be that bad.

I'll ping you when I have something to show and tell ;)

jsiarto:
I agree with Mark's comments on the modal--especially as it relates to the welcoming the user to EDSC. The modal will be a clear call to action and will make it very obvious that the user is about to start the tour/interact with EDSC in a different way.

jboukes:
@mreese84 @jsiarto - I believe it clicked for me when Mark mentioned that the modal wasn't really part of the tour.  Give this new version a shot.

Since the modal is not truly part of the tour, you cannot go 'previous' back to it.  Further, when the 'Show Tour' buttons are clicked when the tour is hidden, it does *not* start at the modal - it starts at the beginning of the actual tour.

I took the liberty of adding a NASA logo in the initial modal screen as it just felt like it needed something.  Also, to make it look nice and stand out, I used the FontAwesome font-family (which really *is* nice).  Of course this stuff can be tweaked.

Let me know what you think.

app/assets/javascripts/modules/help.js.coffee
```diff
@@ -0,0 +1,253 @@
...
+                  </label>
+                </div>
+              </div>"
+
+  tour = [{
+      title: "Search"
+      content: '<p>Use Earthdata Search Client\'s natural language processing-enabled search tool to quickly narrow
+                down to relevant collections.  An example search phrase could be "Land Surface Temperature over
+                Texas last month".  Results will be displayed in the collection panel below.</p>
+                <p>If you would prefer to pick a temporal range from a calendar (<i class="fa fa-clock-o"></i>),
```
mreese84: Let's change this to:
`<p>If you would prefer to pick a temporal range from a calendar (<i class="fa fa-clock-o"></i>) or manually set spatial boundaries (<i class="fa fa-fw fa-crop"></i>), use the buttons to the right of the search box.</p>`

Since you can't actually see the shapefile upload from this initial screen I feel like mentioning it specifically will cause confusion.

app/assets/javascripts/modules/help.js.coffee
```diff
@@ -0,0 +1,253 @@
...
+      showNext: true
+      top: null
+    }, {
+      title: "Map Tools"
+      content: '<p>Use these standard map tools to configure and position the map as well as enable certain spatial search tools.</p>'
+      placement: 'left'
+      element: '.leaflet-control-zoom-in'
+      showNext: true
+    }, {
+      title: 'Toolbar'
```
mreese84: On this last stop of the tour, it would be nice to turn the "End Tour" button blue as that is the primary call-to-action at that point.

app/views/search/_sitetour_modal.html.erb
```diff
@@ -0,0 +1,22 @@
+<div class="modal fade sitetour" id="sitetourModal" style="font-family: FontAwesome;" tabindex="-1" role="dialog">
```
mreese84: Why are you setting inline styles here (font-family)? We should be setting things like that in a sass file. Also, the modal text is not inheriting the global font styles. Notice that all fonts throughout the app are sans serif while the font in the modal is serif.

jboukes: Because I think FontAwesome is pretty and it was a one-timer.  I'll remove it ;)

mreese84: Yeah, I'm cool with using FA in this modal... we just want to make sure that we are setting our CSS in CSS files rather than inline with our HTML.

jboukes: I'll add that to the CSS and make it so - we should have a sideline chat about inline CSS at some point.  I want to make sure I know the team rules; I tend to use inline for what appears to be 'very minor' bumps, but that gets subjective.

mreese84:
@jboukes, I've removed some unnecessary CSS. There's another issue that I'm seeing though that I can't figure out. I'm not seeing the tour or the option to trigger the tour consistently. Sometimes I can get it to load and see the appropriate buttons. But more often, I see the "show tour" button for a split second and then it disappears. Almost as if there's a logic check that is removing it.

app/assets/javascripts/modules/help.js.coffee
jboukes: This is done :)

spec/features/address_bar_spec.rb
```diff
@@ -10,6 +10,9 @@ def query_string
   context 'when loading collection details page from search page' do
     before :all do
       visit '/search'
+      if page.has_link?('closeInitialTourModal')
```
mightynimble: has_link? (I think) makes our tests _really_ slow. (150% slower overall)

spec/features/address_bar_spec.rb
```diff
@@ -10,6 +10,9 @@ def query_string
...
+        find("#closeInitialTourModal").click
+      end
```
mightynimble: The whole block of code can be extracted into a method and put it in helper file.

app/assets/javascripts/models/page/search_page.js.coffee
```diff
@@ -41,6 +45,9 @@ ns.SearchPage = do (ko
       $('#variablesModal').modal('show')
     $('.launch-customize-modal').click ->
       $('#customizeDataModal').modal('show')
+    preferences = new PreferencesModel()
+    sitetour = new SiteTourModel();
+    $('#sitetourModal').modal('show') if sitetour.safePath() && !window.edscportal && (preferences.doNotShowTourAgain() == 'false' || window.location.href.indexOf('?tour=true') != -1)
```
mightynimble: We need to make sure ```preferences.doNotShowTourAgain()``` is string when compare it to ```'false'```

app/assets/javascripts/models/ui/sitetour.coffee
```diff
@@ -0,0 +1,23 @@
...
+
+ns.SiteTour = do (help = @edsc.help, PreferencesModel = data.Preferences,  urlUtil=@edsc.util.url) ->
+  class SiteTour
+    startTour: () =>
+      help.startTour()
+
+    safePath: () =>
+      return if urlUtil.cleanPath() then (urlUtil.cleanPath().split('?')[0] == '/search' || urlUtil.cleanPath().split('?')[0] == '/') else false
+
+    toggleHideTour: () =>
```
mightynimble: This method can be simplified.

app/assets/javascripts/modules/help.js.coffee
```diff
@@ -0,0 +1,248 @@
...
+  index = 0
+  next = null
+  close = null
+
+  tourRunning = false
+
+  doNotShowTourAgain = false
+
+  toggleHideTour = ->
+    doNotShowTourAgain = if doNotShowTourAgain then false else true
```
mightynimble: Redundant

app/views/search/_sitetour_modal.html.erb
```diff
@@ -0,0 +1,22 @@
...
+      </div>
+      <div class="modal-body" style="margin-left: 20px;">
+        <%= image_tag "nasa.png", :class => "pull-left"%>  <p>This application allows you to search, discover, visualize, refine, and access NASA Earth Observation data.  If you are new to this application, please follow the brief tour to get an overview of the features that will 
+        help you achieve your goals.</p>
+      </div>
+      <div class="modal-footer">
+        <a class="button text-button" id="closeInitialTourModal" data-dismiss="modal" href="#">Close</a>
+        <button class='button' data-bind="click: function(){ $root.ui.sitetour.startTour(); }">Start Tour <i class='fa fa-arrow-circle-right'></i></button>
+        <label class='pull-left' style='padding-top: 5px;padding-left: 5px;'>
+          <input class='toggleHideTour' id='toggleHideTour' type='checkbox' data-bind="click: function(){ return true; }, checked: $root.ui.sitetour.toggleHideTour(); "/>
```
mightynimble: why ``` return true``` here?

app/views/shared/_login.html.erb
```diff
@@ -1,4 +1,5 @@
 <!-- ko if: window.tokenExpiresIn === null -->
+<a href="#" data-bind="click: function(){ $root.ui.sitetour.startTour(); }, visible:  $root.ui.sitetour.safePath();" class="toolbar-button">Show Tour</a>
```
mightynimble: 1. I am not sure if this is the right file to put the button.
2. If the tour is only gonna show up  on collections page, a function should be created in search_page and put the function name here.
3. You can put more logic in the method to check if it is on collections page so that the dependencies added on account and data_access js files can be removed.
4. ```visible``` binding hides the button for me.

app/views/shared/_login.html.erb
```diff
@@ -20,6 +21,9 @@
       <a href="<%=edsc_path '/projects' %>" class="dropdown-link dropdown-link-projects">Saved Projects</a>
     </li>
     <li>
+      <a href="#" data-bind="click: function(){ $root.ui.sitetour.startTour(); }, visible:  $root.ui.sitetour.safePath();" class="dropdown-link">Show Tour</a>
```
mightynimble: same here. 
Also I can't see the "show tour" menu option.

mightynimble:
Overall, the button/menu option visibility and the behavior of initial page load are flaky
1) the modal won't show up for me on initial page load
2) when show tour button is clicked, I don't see the modal but the tooltip pointing to the search box (is this expected?)
3) the show tour menu option is not displaying for me after I log in
4) when the modal does show up and is dismissed, clicking on show tour won't make it display again (similar to 2))
5) when the modal does show up, there is no show tour button or the menu option.

jboukes:
Hi @mightynimble - thank you for the review!  My numbered answers are matching the numbers of your comments:

1.)  What was the URL used when you say that the initial modal would not show up for you?  The modal - by design - does not show up on pages other than '/' or '/search' - this is because the tour shouldn't be shown on the other paths.

2.) Yes - that is by design.  The modal is not actually part of the tour - it is the introductory splash screen for the tour.  The first *actual* tour window is 'Search'.

3.) What was the URL you were on when you noted that the 'Show Tour' menu option was not visible upon logging in?  As with 1), above, the menu option only appears when you are on '/' or '/search'.  It has a visibility binding to hide it on any other page.  Also, note that upon logging in 'Show Tour' is in the 'Manage Account' drop down and is not immediately visible.

4.)  As with 2.) - the modal isn't part of the tour, so clicking 'Show Tour' wouldn't do that.  However, if you close the 'Search' popup (or any one of the tour popups) with the 'Do not show again' unchecked, and then you refresh the app at '/' or at '/search', then you will again see the modal.

5.)  What does the modal look like when you open it if it doesn't have any buttons?  It should have a 'Start Tour' and a 'Close' button in the lower right hand corner of the modal - is this not visible?  If so, can you include a screenshot?  

Thanks!

app/assets/javascripts/models/data/preferences.js.coffee
```diff
@@ -24,11 +25,16 @@ ns.Preferences = do (ko
...
+          type: 'GET'
+          dataType: 'json',
+          async: false,
```
jboukes: I had a nice comment explaining this but then thought it was needed and then deleted it - lol.

The problem is there was a race condition when using the getJSON method.  Preferences, once instantiated, uses this AJAX call to in order to populate itself.  Therefore, if you instantiate it, and immediately check a preference - it won't be loaded yet and will return an incorrect value.  The asynchronous call to getJSON won't be done in time.

So, the answer is removing the asynchronous nature of the call.  Here I did this by switching getJSON to an .ajax call so as to have access to the async option.  It works and does fix the issue.  You're right that its been deprecated; but it hasn't been removed from our build of jQuery yet.  I doubt we'll be changing that jQuery version soon.  We can look at changing this do a different strategy for sure.

app/assets/javascripts/models/page/access_page.js.coffee
```diff
@@ -7,10 +7,12 @@
 #= require models/ui/service_options_list
 #= require models/ui/account_form
 #= require models/ui/feedback
+#= require models/ui/sitetour
 
 data = @edsc.models.data
 ui = @edsc.models.ui
 ns = @edsc.models.page
+sitetour = @edsc.sitetour
```
jboukes: I believe you're right - I'll snipe it (same with the others).

app/assets/javascripts/models/data/preferences.js.coffee
```diff
@@ -51,11 +57,13 @@ ns.Preferences = do (ko
     fromJson: (jsonObj) =>
       return unless jsonObj?
       @showTour(jsonObj.show_tour != 'false')
+      @doNotShowTourAgain(jsonObj.doNotShowTourAgain)
```
jboukes: @mightynimble - done.  All those console entries should show strings after I push here in a bit.

spec/features/address_bar_spec.rb
jboukes: Thank you for this one - I have seen a slowdown and did not know has_link was a trap.  I'll figure out a different method and update these tests.

app/views/search/_sitetour_modal.html.erb
jboukes: @mightynimble - I've the same question, now.  I feel like there was a purpose to that but the code doesn't seem to need it now.  Removing - thanks!

app/views/shared/_login.html.erb
jboukes: By the ticket's requirements, a 'Show Tour' needs to be included in the 'Manage Account' drop down, and this is where that drop down is defined, so - this is where the code went.  ;)

The visibility option will hide the button if you are on any page other than '/' or '/search'.

mightynimble:
OK - I think I can reproduce it constantly:
a) initial load page http://edsc.dev/search, the tour modal won't show up. Refresh the page, the modal shows up
b) Type in edsc.dev in url ('/' path), I am not seeing the modal showing up nor the "show tour" button
screen shot 2017-05-12 at 2 03 32 pm
c) Continue from b), click login and log myself in. I see the modal but not the "Show Tour" button 
screen shot 2017-05-12 at 2 04 16 pm
d) Continue from c), close the modal and click on the user icon, I can see "Show Tour" menu option but not the button.
screen shot 2017-05-12 at 2 05 16 pm

jboukes:
@mightynimble - at least some of that is by design.  Once the user logs in, the 'Show Tour' button in the upper right hand corner actually does go away - the link within 'Manage Accounts' suffices.  The only reason a button is used in the case the user is not logged in is that there isn't a drop down to put it in.

But - the part with the modal not coming up with http://edsc.dev/search is a little concerning - I am wondering if it is linked to how Pow works?  I am not using Pow - I just straight up use http://localhost:3000, and the modal shows up every time (unless 'Do not show again' has been set of course).

Do you see the same behavior when using http://localhost:3000 as the base?

spec/features/address_bar_spec.rb
jboukes: @mightynimble - this was a million times true - has_css is so much faster.  Thanks!

jboukes:
@mightynimble - I added a line of debug code to the console that will output something like 'Safepath has /search' to show what it is seeing.  Could you try again with that http://edsc.dev/search url?  I have a feeling POW is doing something here but am not sure.

app/assets/javascripts/modules/help.js.coffee
jboukes: @mightynimble - hmmmmm If I remove line 112 above, then the 'End Tour' button no longer works.  It's needed.  :\

app/assets/javascripts/models/data/preferences.js.coffee
jboukes: @mightynimble - I'm looking for an alternative approach but it's being a bit elusive.  I'll keep trying.

app/assets/javascripts/modules/help.js.coffee
mightynimble: I meant that line 115's if-then-else can be simplified.
For example, line 115 and 116 can be combined to something like:
```$('input:checkbox.toggleHideTour').prop 'checked', !doNotShowTourAgain```
if ```doNotShowTourAgain``` is a string, you can compare it with string "false

File: spec/features/address_bar_spec.rb (hunk 5 above)
Comment (mightynimble) [inline spec/features/address_bar_spec.rb]: I think ```has_css``` is also slow. You can try ```execute_script``` or something like that.

File: app/assets/javascripts/modules/help.js.coffee (hunk 9 above)
Comment (jboukes) [inline app/assets/javascripts/modules/help.js.coffee]: Hmmm - here, it is a boolean.  I believe the only time it's a string is when exported from preferences for the external check.  Line 115 is needed in order to toggle the boolean from true to false or vice versa.  This boolean is needed to keep track of the state - when the 'End Tour' button is clicked, this boolean is used to save the preference of 'do not show again'.

Line 116 then ensures the checkbox matches the new assigned preference.

The line you gave would set the property of the checkbox but it doesn't actually toggle the boolean.  :\

File: app/assets/javascripts/models/page/search_page.js.coffee (hunk 7 above)
Comment (jboukes) [inline app/assets/javascripts/models/page/search_page.js.coffee]: Now it's always exported as a string :)

**Comment by jboukes** [comment]
@mightynimble - I've further tried to reproduce the issue you are seeing where the modal does not appear when you open http://edsc.dev/search - I did (finally) setup Pow and it was not to blame.

If your preference was set to 'do not show again', then of course it wouldn't show - but otherwise I'm a bit at a loss as I can't recreate. 

I did add a console debug line that should report what safePath is seeing - if you might recreate and check that, it might help.

File: app/assets/javascripts/models/data/preferences.js.coffee (hunk 13 above)
Comment (mightynimble) [inline app/assets/javascripts/models/data/preferences.js.coffee]: First try to put the checks in ```success``` callbacks. If that doesn't work, I think you can use ```settimeout``` to check the preferences.

**Comment by mightynimble** [comment]
I didn't open http://edsc.dev/search. I opened http://edsc.dev which redirects to /search. (See step a)). Maybe the 'false' and false conversion resolved it?
Now what I am seeing is:
1) open a incongito session in chrome
2) type in "edsc.dev" and press enter
3) the modal doesn't show up but I can see the show tour button
4) refresh the page, the modal shows up.

**Comment by jboukes** [comment]
@mightynimble I was able to recreate that incognito bug - when I added a blank "" as an acceptable path, that issue went away as well.  Just pushed up.

File: app/assets/javascripts/models/ui/sitetour.coffee (hunk 8 above)
Comment (jboukes) [inline app/assets/javascripts/models/ui/sitetour.coffee]: Done :)

**Comment by jboukes** [comment]
@mreese84, @mightynimble - Recent pushups removed the usage of the deprecated 'async' usage, as well as simplified other code.  The tests are much faster now as the modal is being closed by a very direct method, rather than the has_link route (I was overthinking that prior).  The 'incognito' bug also appears to be dealt with.  

I _believe_ I've mitigated all of Bo's comments - Bo, please correct me if I'm wrong in saying that.  If there is anything remaining that blocks a technical approval on this PR, please let me know.

Thanks!

--JB

**Comment by mightynimble** [comment]
One more thing - the show tour button is visible when the page is navigated to other pages:
1. open edsc.dev
2. click on any link that takes the user to another page, e.g. a collection.
(I think we should add some tests for the visibility of the show tour button)

**Comment by mightynimble** [comment]
So tests to be added:
 - the visibility of the show tour button and the menu option
 - "do not show tour" as a guest

File: app/assets/javascripts/models/page/search_page.js.coffee [hunk 16]
```diff
@@ -49,8 +43,8 @@ ns.SearchPage = do (ko
       $('#variablesModal').modal('show')
     $('.launch-customize-modal').click ->
       $('#customizeDataModal').modal('show')
-      
-    setTimeout initModal, 2000 if !window.edscportal
+    if (url.cleanPath()?.split('?')[0] in ["/search", "/", ""]) && current.preferences.serialize()?.doNotShowTourAgain == 'false'
+      $('#sitetourModal').modal('show')
```
jboukes: This new technique was not reading the preferences correctly somehow - the tour modal was coming up every time regardless of preference.

app/views/search/_sitetour_modal.html.erb
```diff
@@ -10,10 +10,10 @@
       </div>
       <div class="modal-footer">
         <a class="button secondary" id="closeInitialTourModal" data-dismiss="modal" href="#">Close</a>
-        <button class="button" data-bind="click: function(){ $root.ui.sitetour.startTour(); }">Start Tour <i class="fa fa-arrow-circle-right"></i></button>
+        <button class="button" data-bind="click: $root.ui.sitetour.startTour">Start Tour <i class="fa fa-arrow-circle-right"></i></button>
         <p class="pull-left">
           <label>
-            <input class="toggleHideTour" id="toggleHideTour" type="checkbox" data-bind="click: function(){ $root.ui.sitetour.toggleHideTour(); return true;}" />
```
jboukes: Deleting this code broke the 'Do Not Show Again' checkbox - it would never check.  The other code was setup to let it check.

mightynimble:
@jboukes The load preference is synchronized. It just needs to be initialized in the constructor and everything should be fine. The code outside the ```class SearchPage``` initialize the ```Preferences``` and ```SiteTour``` one more time which is not necessary and will cause inconsistency. The 2-second ```timeout``` will introduce another race condition if the preferences loads too slow.

There was an issue in the view ```app/views/search/_sitetour_modal.html.erb``` where ```function(){ $root.ui.sitetour.toggleHideTour(); return true;}``` somehow executes the toggleHideTour() function and you can see that there are multiple save and load operations to the preferences in the js console which turn on and off the checkbox. I moved the method out of the "function" and also removed the brackets to prevent it from being called and executed. A "true" line is added to the ```toggleHideTour``` method to allow the event bubble up.

@mreese84 and I also found a minor issue that ```preferences``` could be null if there is nothing in the db and the session is clean. A nil-check is added.

I can still see that ```doNotShowTour``` is still being set to both string ```"false"``` and boolean ```false``` (in helper.js.coffee). I pushed that and hopefully it is the last place that needs the change.

jboukes:
@mightynimble - your refactoring broke things (again).  :(  

The modal no longer comes up at all no matter what the preference.  

If we had just fixed the show/hide tour button issue and then finished the test, this ticket would be mergable.

I'll take a look at why the modal is not coming up and will likely revert code.

app/assets/javascripts/models/page/search_page.js.coffee
```diff
@@ -50,9 +42,6 @@ ns.SearchPage = do (ko
       $('#variablesModal').modal('show')
     $('.launch-customize-modal').click ->
       $('#customizeDataModal').modal('show')
-    
-    setTimeout initModal, 2000 if !window.edscportal
-
```
jboukes: @mightynimble - I believe you were the one who suggested using setTimeout (I could find the comment)  Further, deleting this code completely broke the tour.  The modal does not come up at all now.

app/assets/javascripts/models/data/preferences.js.coffee
jboukes: @mightynimble - this is where I got the idea for using setTimeout to check preferences for the initial modal.

mightynimble:
Sorry @jboukes , I accidentally removed the lines when resolving conflicts and forgot to add them back. Pushing a change to make the modal work.

app/assets/javascripts/models/page/search_page.js.coffee
mightynimble: Right. It was me... (Although I think setinterval is a better option if we want to keep polling for the right preferences to come back.) I removed it because 1) the call is already synchronized (see an earlier commit in preferences.js.coffee) and 2) put it there requires initialize the sitetour and the preferences twice (one in the $(document).ready block, one in the class).

jboukes:
@mightynimble - the change you made to make the modal work is the same code I had to remove last night because it broke things.  Initiating the modal in that manner somehow ignores preferences - the modal will show up no matter how you've set the preferences.

I'm reverting it back (for the second time) to the original code.  I know it's not as pretty as the code you pushed up; however, it works, and working is more important than aesthetics.

PR #416, opened 2017-05-08
EDSC-1337 Fix an issue in ESI order chunking. Adding more logs.

mightynimble:


app/models/retrieval.rb
```diff
@@ -115,7 +117,11 @@ def self.process(id, token, env, base_url, access_token)
                 method[:error_message] = []
                 method[:error_message] << error_message
 
-                page_num = results_count / page_size + 1 if Rails.configuration.enable_esi_order_chunking
```
mightynimble: The original 'if' condition will stop the loop as long as the ```enable_esi_order_chunking``` is set to true.

app/models/retrieval.rb
```diff
@@ -115,7 +117,11 @@ def self.process(id, token, env, base_url, access_token)
...
+                # break the loop
+                if !Rails.configuration.enable_esi_order_chunking || limited_collection?(method['id'])
```
mightynimble: ```limited_collection...``` part is not necessary since we have already limited the total number of granules to 100 on granule list page. Just to be safe here in case someone uses a bookmarked link to re-order the granules.

PR #422, opened 2017-05-16
EDSC-1453

mreese84:
Update limited collection help text.

### DIFF
--- a/app/assets/stylesheets/modules/buttons.css.scss
+++ b/app/assets/stylesheets/modules/buttons.css.scss
@@ -293,3 +293,7 @@ $feedback-button-hover-color: lighten($feedback-button-color, 10%);
     padding-bottom: 10px;
   }
 }
+
+.button-disabled {
+  text-shadow: 1px 1px 1px rgba(0,0,0,0.15);
+}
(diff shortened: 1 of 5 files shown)


mightynimble:
The changes look good. 
It is not related to your change but I noticed that the button is clickable before the granule list loads. And user can still be led to the data/configure page. There is no issue code-wise since when user submits the esi order, it will be limited to the first 100 granules. However, should we change the red text on the page as well?
```This order contains
            <span data-bind="text: granuleAccessOptions().hits"></span> granules. Some access methods may not support access beyond the first 2000 granules.```

mreese84:
@mightynimble, I'd rather solve the problem of the button being clickable before the granule list loads. Is there a better way to handle this?

mightynimble:
I can disable the button by default and only enable it when the granules are loaded.

mreese84:
@mightynimble can you do that for these specific collections? @jsiarto, how do you feel about disabling the download button until the granules have loaded? I certainly makes sense for these "limited collections", but what about app-wide?

mightynimble:
I don't find a way to disable the button only for limited collections. (Making it disabled by limited collection is do-able but it also waits for granules to load then the binding is evaluated. So it is useless.)

mreese84:
@mightynimble, let's not worry about it for now then. We'll keep an eye out for this being an issue, but I don't want to make a sweeping change like that without really thinking through it or determining if there is a need for it or not.

PR #426, opened 2017-05-18
EDSC-1481 Clicking 'x' hides grid filter inputs.

mightynimble:


mreese84:
@mightynimble, the clearing button now works... but it's still possible to add 2 spatial search types. Check the ticket description and screenshot. Added screenshot here too.
screen shot 2017-05-18 at 11 52 59 am

mightynimble:
So if there are coordinates in the point input and user picks 'grid' filter, should the point filter be hidden? i.e. only show one spatial type at any given time?

mreese84:
Yep, that is how it currently works in production I assume. You can only send 1 type of spatial search at a time... right?

mightynimble:
For point/rectangle/polygon types, yes, you can only specify one of them. However, you can search by grid in addition to those three spatial types.

mreese84:
@mightynimble, ahh... I wasn't aware of that. If that's the case, then we can leave it in the stack. That's cool.

mightynimble:
I added a test to make sure that the applied grid param is removed from the query when the 'x' is clicked.

PR #427, opened 2017-05-18
EDSC-1480-2: UI Feedback

jsiarto:


mreese84:
@jsiarto, got a couple of notes for you.
1) Personally I feel like the granule names font size is too big. I bumped things down to 1em and it feels/looks a little better to me. If you recall, some of the feedback was that the fonts felt too big. In some cases I disagree, but in this case I agreed. Bumping things down a hair could also prevent granules names from wrapping to three lines, which isn't ideal at all.
2) Hit this link http://edsc.dev/search/granules?p=C203234444-LAADS&g=G1386759068-LAADS&m=-87.46875!-19.6875!0!1!0!0%2C2&tl=1479344561!4!!&ff=Map+Imagery and expand the granule list. You should be able to scroll down and find an odd thing happening when certain granule panels (or whatever we're calling those blocks) don't have browse imagery. I've posted a screenshot in this comment, notice how the flow of those blocks can get jacked up depending upon the height of elements within the row.
3) I noticed another oddity when you exclude a granule. Just close any granule and notice where the "Granule Excluded. Undo" element comes up. The positioning needs to be fixed and I think the "undo" text color is hard to read.
screen shot 2017-05-18 at 8 06 16 pm

PR #429, opened 2017-05-24
EDSC-1480: Hide project button from all non-collection views

jsiarto:


mightynimble:
Remove the ```showProject``` method in coffeescript as well.

PR #431, opened 2017-05-25
EDSC-1431 Reset and Revert echoform settings.

mightynimble:


app/views/data_access/configure.html.erb
```diff
@@ -178,6 +178,8 @@
 
         <div class="access-item-actions">
           <button data-bind="visible: $index() > 0, click: $root.ui.serviceOptionsList.showPrevious" class="button-back low-priority">Back</button>
+          <button data-bind="visible: !$root.ui.serviceOptionsList.isLastOption() && isResetable, enable: serviceOptions.readyToDownload(), click: $root.ui.serviceOptionsList.resetForm, attr: {id: 'reset-form-' + collection.id}" class="button-next">Reset</button>
```
mreese84: We should provide a hover tip that says "Reset all customization options to their default values."

app/views/data_access/configure.html.erb
```diff
@@ -178,6 +178,8 @@
...
+          <button id="reload-previous" data-bind="click: $root.ui.serviceOptionsList.reloadPrevious, attr: {id: 'reload-previous-' + collection.id}" style="display: none;" class="button-next">Revert</button>
```
mreese84: On this line we should provide a hover tip that says "Revert all customization options back to the saved values."

PR #432, opened 2017-05-25
EDSC 1464 - Added toggle for non-EOSDIS collections

jboukes:


app/views/search/index.html.erb
```diff
@@ -110,6 +110,8 @@
           <p class="collection-filters">
             <input type="checkbox" id='has-granules' data-bind="checked: query.hasGranules"/>
             <label for="has-granules">Include collections without granules</label>
```
mreese84: How hard would it be to do the inverse of this granules toggle? It would be preferable to have a checked check box and the text say "Only show collection with granules". I like have a checked check box by default as it draws a little more attention to the feature.

mightynimble: It will require an update on all the collection fixtures, which is why we gave this ticket a 3 points. (@john, you can manually edit the collection fixtures if it is easier than delete-all and re-gen them.)

mightynimble:
1. What @mreese84 said above - the third acceptance criteria is asking that by default, all collections should be shown. (It confused me as well and I had to confirm with Mark.)
2. When the new "non-EOSDIS" checkbox is checked, the url is not updated to reflect the change, meaning after a refresh, the checked state will not be preserved. 
3. I suggest to move the parameter filtering to the back end rails so that the tag values can be easily configured in yml files. This way we will be ready to support more 'search-by-tag-name' requests in the long term. If this is going to blow the scope and take multiple days to refactor… then let's wait and see and refactor when needed. You can take a look at this example on how to make the query param a true/false flag and let the back end handle the mapping from the flag to the real CMR param: (https://github.com/nasa/earthdata-search/pull/404)

app/views/search/index.html.erb
jboukes: @mark inverting the logic was easy up until making the URL parameters work.  The inverted has_granules checkbox works - it's the page refresh that's not working well.  Since it is inverted that means it needs to take an action when its state is *false* - not true - and somehow this isn't syncing right with how URL parameters are handled.

The exact same issue is also causing the new non-EOSDIS toggle to not handle a page refresh well, either.  As @mightynimble requested - I added code so that it is seen in the URL.  However, like the inverted hasGranules, it adds something to the URL only when it is *unchecked*.  When it is checked, nothing is added to the URL - it is only when unchecked that an action is taken.  This little inverted spaghetti knot is causing both checkboxes to get confused on page refreshes.

I'll work on it a bit more (though have an errand to run shortly) and will get back to you two later today.

jboukes:
@mark, @bo - the 'has-granules' checkbox is now logically inverted, and both checkboxes default to 'checked'.  It does take a moment from checked to unchecked when loading from a parameter (a loaded parameter gives them a 'false' setting), but it works.

@mightynimble - you mentioned the need to regen the collection_requests and _responses fixtures - I did check the new versions in, that were modified from running the tests.  Let me know if that suffices (or not).

PR #435, opened 2017-05-26
EDSC-1496: Remove dead space on echoforms bounding box

jsiarto:
Given how echoforms works and current state of markup and style--this is the best we're going to get.


app/assets/stylesheets/modules/echoforms.css.scss
```diff
@@ -6,6 +6,10 @@
     border-bottom: 1px solid rgba(0, 0, 0, 0.05);
   }
 
+  h1.echoforms-label:empty {
```
mreese84: yeah-uh!

PR #436, opened 2017-06-02
EDSC-1477: Horizontal map controls

jsiarto:


mreese84:
Just discussed with Jeff that we need to figure out the positioning issue when moving from collections panel to granule panel and figure out a way to overcome the map scale legend moving the map controls when zooming.

jsiarto:
@mreese84 I agree--it needs some polish, there is definitely a lot going on. I tried to at least centralize the code that handles the reposition, so when we do edit in the future it's a bit easier and already done the "EDSC" way.

PR #448, opened 2017-06-29
EDSC-1465: Triggers echoforms validation for <select> fields after in…

jboukes:
…itial form generation in order to remove incorrect 'pruned' occurrences


jboukes:
Original steps to recreate bug (from ticket):

Steps to reproduce:
1.Search for SPL2SMP_E with the following spatial filter (WSEN): -169,54,-141,74
2. Select SMAP_L2_SM_P_E_10070_A_20161220T020725_R14010_001.h5
3. Choose Customize on the Data Access page, check "Enter bounding box", reformat to HDF-EOS, re-projection to UTM Northern Hemisphere and enter 5 as the zone. Select all applicable parameters.

To test, follow those instructions at least twice.  The issue resulted from loading the data access options from cache and then submitting without change; the first time (without cache) it was working.  So, if you try this at least twice, and the email sent to you includes all pertinent options (particularly UTM Northern Hemisphere and 5 as the zone), then we should be good.

PR #449, opened 2017-07-06
EDSC-1517

mreese84:


mreese84:
Need to remove old ECHO calendar code.

mreese84:
@mightynimble @jsiarto There's a weird scrolling issue I'm seeing. If you position your mouse over the facets and use a scroll wheel or 2-finger scroll on a Mac, Tophat and the footer both scroll up and down, when they should be fixed.

I did some digging and app/assets/stylesheets/modules/layout.css.scss:20 sets a min-height. If I turn that off the scrolling issue is fixed. Do you know of any reason we would need to keep that? (not to mention the fact that we have min-height and TWO height rules applied. @jsiarto, what gives?)

mreese84:
@mightynimble I've committed all the changes that I know to make for removing Status app/banner related things with the exception of app/assets/javascripts/modules/banner.coffee. I'd rather you handle that file. Please review this extremely carefully to ensure we aren't losing any functionality anywhere. If you know of anything else that needs to be stripped out please either comment here or just strip it out when you handle banner.coffee as well.

mreese84:
Thanks @mightynimble. @jboukes can you please give this a review so we can get it merged and tested in SIT? Thanks!

PR #450, opened 2017-07-06
EDSC-1465: Moved validating toggle outside of knockout binding

jboukes:


app/assets/javascripts/models/data/service_options.js.coffee
```diff
@@ -119,7 +118,12 @@ ns.ServiceOptions = do (ko, edsc = @edsc, KnockoutModel = @edsc.models.KnockoutM
                 if !accessMethodMatched && matches[1] == jsonObj.collection_id && parseInt(matches[2], 10) == index
                   accessMethodMatched = true
                   @loadForm(true) if jsonObj.type == 'service'
-                  ko.applyBindingsToNode(echoformContainer, {echoform: this})
+                  # After the binding is complete, toggle select elements to trigger validation lest they be pruned
+                  $.when(ko.applyBindingsToNode(echoformContainer, echoform: this)).done ->
```
mightynimble: The issue is specific to reprojection params. So I would limit to only the reprojection param section of the form. For some huge smap forms, trggering each "change()"will probably take an extra second..

jboukes: @mightynimble  - Echoforms builds itself so generically that I'm not seeing an obvious way to limit the selects to those only within the reprojection section.  The only way I can think of atm is to update echoforms to add a generic class name around the reprojection section to mark it, and then use that as an anchor to somehow narrow down the scope of the toggle.  In other words  $('.echoforms-element-select').each will become something like  $('.projectionSection.echoforms-element-select').each (syntax might be off).

The con is that this would take a moment or two to figure out and then retesting and such - so, it wouldn't be done probably until the first part of next week (being this is late Thursday and I have a higher priority ticket).  Do we want to delay this current fix in order to _potentially_ save a second on very large forms?

Further - while this particular bug is specific to reprojection params, it isn't difficult to imagine that other select params that we're not aware of may also suffer this same bug and so will require toggling.  The bonus of doing all of them is it makes sure that all of them have been validated. 

@mreese84 - what do you think?

mightynimble:
Sorry I didn't make my self clear - your query selects all drop downs and reset their values. I would prefer to find the reprojection one and trigger a validation only on that one element.

app/assets/javascripts/models/data/service_options.js.coffee
mightynimble: Sorry I didn't make my self clear - your query selects all drop downs and reset their values. I would prefer to find the reprojection one and trigger a validation only on that one element.

app/assets/javascripts/models/knockout_extensions.js.coffee
```diff
@@ -159,24 +159,6 @@ do (ko, $=jQuery) ->
           try
             $el.echoforms(form: form, prepopulate: options.prepopulatedFields())
 
-            # this needs to be done to trigger echoform validation otherwise these nested values will be pruned.
-            if shortName?
-              projectionOptions = $('#' + shortName + '-projectionSettings:visible')
```
mightynimble: For example, this finds the projection options. If ```shortname``` is hard to find in the html, you can use wildcard to get all ```*-projectionSettings``` that are visible.

app/assets/javascripts/models/knockout_extensions.js.coffee
```diff
@@ -159,24 +159,6 @@ do (ko, $=jQuery) ->
...
-              if projectionOptions.length > 0
-                reprojectionOptionsLabel = $('label:contains("Re-projection Options"):visible')
```
mightynimble: If it exists, look for the drop down ```reprojectionSelect```

app/assets/javascripts/models/data/service_options.js.coffee
jboukes: @mightynimble @mreese84  - I think this'll do it.  I realized that projection select dropdowns are always going to have 'NORTH POLAR STEREOGRAPHIC' as an option - so why not just leverage that?  One line fix.  This new version only toggles selects with that option available.

mightynimble: This only works for "North Polar Stereographic". It doesn't even work for South Polar Stereographic...

See my earlier inline comments in knockout_extensions.js.coffee.

The following is a chunk of code that you can put within $.when...done block. It is just a general idea, I didn't test it and it may even have syntax errors:
```
projectionOptions = $("[id$=-projectionSettings]").filter(":visible")
if projectionOptions.length > 0
    reprojectionOptionsLabel = $('label:contains("Re-projection Options"):visible')
    if reprojectionOptionsLabel.length > 0
        # touch all select dropdowns
        reprojectionSelect = $('#' + reprojectionOptionsLabel.attr('for'))
        originalVal = reprojectionSelect.val()
        reprojectionSelect.val('&').change().val(originalVal).change() # change() is better than trigger('change') here, I think
        # touch all inputs
        inputs = projectionOptions.find('input[type=text]:visible')
        $.each(inputs, (i, obj) ->
            originalVal = $(obj).val()
            $(obj).val(originalVal + 1).change().val(originalVal).change()
        )


```

jboukes: @mightynimble - It doesn't need to work for "South Polar Stereographic".  What that line of code does is it checks to see if the select contains "North Polar Stereographic" as an option, and the projection select always has that as an option.  

To rephrase to pseudocode, it says "Find all selects on the page that contain 'North Polar Stereographic' as one of its options'

And it works.  It's one line of code.  Your solution is interesting, but why would I replace one line of functional code that's been tested and shown to work with many lines of code that have not?

mightynimble:
Very neat solution. Approved. Make sure to resolve the conflicts.

PR #452, opened 2017-07-06
EDSC 1542: Adding logging to Delayed_Job actions

jboukes:


jboukes:
@bo @mark - let me know what you think, whether there's additional info we can log that I missed, or just grammar/text cleanup, anything - thanks!

The general pattern of the log messages is so that, once you know the job ID, it is easy to grep the log to see its life cycle with Delayed_Jobs, as in:  grep 'Delayed Job (ID: XYZ' *.log

If you want to see it fail, grab a local copy and add a division by zero line '1/0' somewhere in the self.process function of retrieval.rb.  You can then create a data access job locally, use rake jobs:workoff to make it go, and then view the logs to see it work.

app/models/retrieval.rb
```diff
@@ -57,29 +57,39 @@ def description
   # Failure: Executed when the work has errored with an exception three times in a row - after this, the job is abandoned
 
   def self.before(job)
-    queued_jobs = DelayedJob.where('failed_at IS NULL')
-    if job.attempts == 0
-      Rails.logger.info "Delayed Job (ID: #{job.id}) is beginning its work - there are #{queued_jobs.size - 1} orders waiting in line behind it."
-    else
-      Rails.logger.info "Delayed Job (ID: #{job.id}) has begun its work after failing #{job.attempts} time - there are #{queued_jobs.size - 1} orders waiting in line behind it."
+    logger.tagged("delayed_job version: #{Rails.configuration.version}") do
+      queued_jobs = DelayedJob.where('failed_at IS NULL')
```
mightynimble: Not sure if this will expose an SQL injection but ```DelayedJob.where(failed_at: nil)``` is a safer option.

jboukes: @mightynimble updated as requested ;)

mightynimble: Looks good 👍

mightynimble:
@mreese84 could you approve it?I can't - I am not on the reviewer list.

jboukes:
@mightynimble your on the list now!  :)  Apologies - from my view it looked like you were already there, but I think it was just a 'suggestion' by Github.

PR #455, opened 2017-07-13
EDSC-1423 Move search by granule IDs to granule list nav bar.

mightynimble:
The functionality is *moved* to granule list navigation bar.
Use input text box to search by single granule ID and click on "Search Multiple" to search more than one granules; just like how it used to work in the granule filters panel.


mreese84:
Some thoughts below...
* The search button should be on the far right
* The search button should be blue. It should match the button that is in the granule filters panel
* The granule panel position needs to be fixed (as mentioned on standup. ping @jsiarto)
* I think I'd remove the icon after "Search Multiple". I don't think it adds anything for users. Thoughts @jsiarto?
* When a user searches for multiple and sends more wildcards than allowed, the error banner appears as it should, but the granule filters panel still pops out. That needs to be addressed.

jsiarto:
Agree with @mreese84 that the icon after search multiple can be removed. Generally, I think this looks pretty good--I'd really like the search to be performed on keyup as that seems like a better user experience. This is especially true after you clear the search field--it feels like the interface should return all the granules again automatically.

mightynimble:
Last commit fixed two issues:
1. Hide granule filters on errors
2. Show search box, "showing 0 of 0 granules" and feedback link on errors.

Issues:
A multiple search:
```*foo1
*foo2
*foo3```
will populate ```*foo1*foo2*foo3``` in the single granule search box and set url param: ```pg[0][id]=*foo1!*foo2!*foo3```
However, after dismissing the modal and clicking on the search single input then "unfocus" the input box, the url param will change to ```pg[0][id]=*foo1*foo2*foo3```.  We used to use a button to trigger the single granule search to prevent this kind of issue. Now the button is gone and I couldn't think of a clean way to do it...

mightynimble:
Commit above fixes the positioning of granule filter panel. It is done in javascript. From reading the comment of the ```contentHeightChanged``` method, it seems that there is no easy way to do it in pure css. @jsiarto any ideas?

mightynimble:
New changes:
1) remove search multiple modal
2) use the one-liner for everything
3) and the allowed delimiters are ',' and ' '. 
Jeff's idea will require more work - the input loses focus on granule list refresh. I will include the feature if it is easy to do otherwise it will be pushed off to another ticket.

mreese84:
@mightynimble, this is looking pretty solid. I'm finding two oddities though.

1) Using collection C92711294-NSIDC_ECS, copy/paste the following granule IDs into the search field...
MOD10A1.A2017001.h34v09.005.2017003060855.hdf
MOD10A1.A2017001.h34v08.005.2017003060859.hdf
MOD10A1.A2017001.h31v13.005.2017003060339.hdf
MOD10A1.A2017001.h00v08.005.2017003055947.hdf
MOD10A1.A2017001.h12v13.005.2017003060343.hdf
MOD10A1.A2017001.h13v10.005.2017003055910.hdf
MOD10A1.A2017001.h19v07.005.2017003060015.hdf

Sometimes all 7 granules will return, but sometimes only one granule (the very first one listed) will return. Any explanation for that?

2) I wanted to see what would happen as I do a search for multiple granules leading with a wildcard. "*foo1, *foo2" would just fine. However, when I tried to search for "*foo1, *foo2, *foo3" the comma and space were stripped out and the search was transformed to "*foo1, *foo2*foo3". This is weird as there's no explanation what's going on. I think we could leverage the error we get back from CMR (maybe even display it down in the granule panels area?) Anyways... it's definitely odd and I'd at least like an explanation as to if it was intentional or what's going on.

Other than that, this looks great.

mightynimble:
There was an error on the regex that didn't match granuleIds() globally. Fixed.

mreese84:
@mightynimble, cool. Searches are all working as expected. I also noticed the tool tip is being cut off at the top. Screenshot attached.

screen shot 2017-07-20 at 7 58 32 am

PR #459, opened 2017-07-17
EDSC-1531: Implemented deferred method to load site tour in order to remove potential race condition

jboukes:
The site tour can potentially load despite being set to 'not show' due to a potential race condition involving a timer.  This new method of loading the site tour uses a 'deferred method' approach - this method adds precision to how the site tour must load after the user's preferences are fetched, and effectively removes the race condition.

### DIFF
--- a/app/assets/javascripts/models/page/search_page.js.coffee
+++ b/app/assets/javascripts/models/page/search_page.js.coffee
@@ -39,7 +39,7 @@ ns.SearchPage = do (ko
   sitetour = new SiteTourModel();
 
   initModal = () ->
-    $('#sitetourModal').modal('show') if sitetour.safePath() && (preferences.doNotShowTourAgain() == 'false' ||  window.location.href.indexOf('?tour=true') != -1)
+    $('#sitetourModal').modal('show') if sitetour.safePath() && (preferences.doNotShowTourAgain() == 'false' || window.location.href.indexOf('?tour=true') != -1)
 
   $(document).ready ->
     current.map = map = new window.edsc.map.Map(document.getElementById('map'), 'geo')
@@ -50,7 +50,7 @@ ns.SearchPage = do (ko
     $('.launch-customize-modal').click ->
       $('#customizeDataModal').modal('show')
 
-    setTimeout initModal, 2000 if !window.edscportal
+    preferences.ready.done(-> initModal()) if !window.edscportal
 
   class SearchPage
     constructor: ->
--- a/app/assets/javascripts/models/data/preferences.js.coffee
+++ b/app/assets/javascripts/models/data/preferences.js.coffee
@@ -13,10 +13,11 @@ ns.Preferences = do (ko
       @showTour = ko.observable(true)
       @doNotShowTourAgain = ko.observable('false')
       @isLoaded = ko.observable(false)
-
       @load()
 
     load: ->
+      readyDeferred = new $.Deferred()
+      this.ready = readyDeferred
       data = window.edscprefs
       window.edscprefs = null
 
@@ -28,6 +29,7 @@ ns.Preferences = do (ko
           console.log "Loaded site preferences, #{JSON.stringify(data)}"
           @fromJson(data)
           @isLoaded(true)
+          readyDeferred.resolve()
       null
 
     onload: (fn) ->


mightynimble:
1). The PR addresses part of the problem. Preferences and tour modal still initialize twice.
2). Dependency jQuery is introduced. At the very beginning "they" decided to purposely stay away from jQuery when possible. There are several reasons. Some of them are the performance, the memory usage, and the html5, etc. It will be good if not using jQuery here. (Maybe a ```setInterval``` will help?) If have to, a dependency line should be added to the top of the file, in case we extract some part of the edsc js into a plugin and break the load order. Anyways, addresssing 1) may eliminate the requirement of jQuery.

jboukes:
>> '1). The PR addresses part of the problem. Preferences and tour modal still initialize twice.'

The tour modal does not initialize twice.  If you have evidence that it does, let me know.

Preferences are maintained by an object that must be instantiated in order to be accessed; this means anything that loads depending upon a preference must instantiate a preferences object in order to check the related preference.  There are multiple (at least more than one) places where this occurs - and so this does mean that the preferences object becomes instantiated more than once during load, but this is by design and there is nothing we can do about that (short of redesign).  Therefore, preferences will always be instantiated multiple times - therefore, again, this means there is nothing I can do about 1.).

The race condition is specifically within the preferences object's constructor.  Since it loads through AJAX, the issue is that the preference object can be instantiated and accidentally checked before the AJAX routine in its constructor can complete.  In this case the default preference would be used, which can easily be wrong - this is what was happening in the race condition, and that's what this fixes.

**We use jQuery EVERYWHERE throughout our application so the second comment very much surprises me.**  For the love of everything - we use Twitter Bootstrap, which has jQuery as a dependency.  Given that, how can anyone suggest 'no jQuery'?  That is bizarre.  **I am not going to 'not use' jQuery.**  This is a solution, it is a good one, and it removes the race condition.

Please point me to the syntax of the 'dependency line' and I'll add it.

mightynimble:
First of all, I am perfectly ok with using jQuery. I was just saying that we don't prefer to use it on places that don't involve making event calls or interacting directly with page elements.

On the double initialization issue and where to add jQuery dependency, see code comments.

app/assets/javascripts/models/page/search_page.js.coffee
```diff
@@ -39,7 +39,7 @@ ns.SearchPage = do (ko
   sitetour = new SiteTourModel();
```
mightynimble: Line 38 creates a new PreferencesModel

app/assets/javascripts/models/page/search_page.js.coffee
```diff
@@ -50,7 +50,7 @@ ns.SearchPage = do (ko
     $('.launch-customize-modal').click ->
       $('#customizeDataModal').modal('show')
 
-    setTimeout initModal, 2000 if !window.edscportal
+    preferences.ready.done(-> initModal()) if !window.edscportal
 
   class SearchPage
     constructor: ->
```
mightynimble: Line 60 creates another new PreferencesModel

mightynimble: The "evidence" of multiple initializations can be seen in the js console. Everytime the preferences is initialized, an ajax call is made which is logged in the console.

The two consecutive calls to preferences will possibly make the code read dirty data if the preferences are changed between the two calls.

jboukes: @mightynimble I do see that - as I described above, any time Preferences need to be accessed, a preferences object must be instantiated.  These two are in two different locations due to scope; one cannot see the other.

In order for the application to check a user preference, it _must_ instantiate a preferences object.  If the preference needs to be checked in two different places in the logic, then that means it must be instantiated two different times (unless the scope is such that it can be used twice, which is not the case here).

However, the instantiation of the preferences object is *not* the problem.  It's not what is causing the race condition or any other issue.  It is actually even 'by design'.  

The race condition would still be a threat even if it were a single instantiation of the preferences object.  The problem lies within the jQuery based AJAX call within the preferences object's constructor.  If the code instantiates a preferences object and then immediately checks a preference, it may accidentally get a default preference as the asynchronous AJAX call within the preference's constructor may not have completed execution.

The beauty of this fix is that it allows communication between the preferences object and the calling code such that it can be confirmed to be 'ready' in a graceful fashion.  It is precise.  The initModal routine will not fire until the deferred object within the preferences constructor is resolved - and it is only resolved within the success condition of the AJAX call.

To sum - yes, the preferences object is instantiated more than once, however this is by design and is not causing the issue.

jboukes: @mightynimble Also, to make it clear - the race condition is due to the usage of an existing AJAX call in jQuery.  One could argue that we are using a jQuery based solution to fix a jQuery based problem.

jboukes: >>"The "evidence" of multiple initializations can be seen in the js console. Everytime the preferences is initialized, an ajax call is made which is logged in the console."

I was asking for evidence that the site tour was initialized more than once.  I acknowledge that preferences is instantiated multiple times; however, I've explained why that is by design.  It is also not breaking anything.

jboukes: >>"The two consecutive calls to preferences will possibly make the code read dirty data if the preferences are changed between the two calls."

They are two calls within the same load procedure (though different scope) - how in the world would any user preference change in that tiny fraction of a time?  That's like suggesting two different reads on a database during load is dangerous as the data might change in between the reads.  

The user preferences won't change if the user does not change them, and it is impossible for a user to change their preferences within the tiny fraction of a second between object instantiations upon page load.

jboukes:
@mightynimble - you did not state how you wanted the 'code dependency' defined.

Regarding the permissions object - we need to create another ticket for that issue, if it really is an issue.  The multi-instantiation of the permissions object has nothing to do with the site tour; it will not make this bug better; it will not make this bug worse.  It is completely unrelated and is scope creep.

@mreese84 - if you would consider approving this PR I would appreciate it.

mreese84:
@jboukes There's enough discussion on this thread that I think we should all sync up on a quick GTM or Screenhero session to make sure we're all on the same page. I'm on calls for the next 2 hours but we can sync up after that.

jboukes:
This whole ticket was WAY overthought.  This should have been approved yesterday.

PR #460, opened 2017-07-18
adding ABoVE project portal

eniad:
The ABoVE project would like a portal. ABoVE has data across DAACs.

I have omitted the ornl-daac-logo because it is not appropriate. If a logo is required, we should use the ABoVE logo. That will need to be added to repo.

### DIFF
--- a/config/portals.yml
+++ b/config/portals.yml
@@ -27,6 +27,15 @@ portals: &portals
     scripts:
       - edsc-portal.ornl.min.js
       
+  above:
+    org: ABoVE
+    title: Search
+    params:
+      project:
+        - ABoVE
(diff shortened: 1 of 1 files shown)


mreese84:
@eniad, I've accepted and merged the PR. I don't have an ABoVE logo. If you would like to add one you can send it over to me and I can take care of getting it added. It's not required, but does add a nice touch to custom portals.

eniad:
@mreese84, thanks. We preferred it without the logo. We aren't sure what link to use for it.

mreese84:
No worries. If you preferred without we can go with that for now. We can always add something later if need be.

PR #463, opened 2017-07-19
EDSC-1531 A test/example on show/hide tour

mightynimble:
This is just a test on show/hide tour and tour splash modal.
The pull request shows a way to initialize preferences only once on search page loading.
It may not address all the issues and may not even work at all. I created this while I was waiting for CI to complete, so it is just an idea of another possible way to fix the race condition.


jboukes:
It doesn't work - it broke the site tour as the race condition is always present now.  

The issue of multiple instantiations of the permissions object has nothing to do with the bug described in EDSC-1531.  Fixing it will do nothing for 1531.  It's simply an unrelated issue.

jboukes:
Also, if we're going to try to fix this, a singleton design pattern approach would be much cooler.

PR #464, opened 2017-07-19
EDSC-1499: Cleaned up remnant of original landing page

jboukes:
The issue was caused by a remnant of the old landing page - it was set to hide itself upon page load.  However, if the page load lags for a moment, it's visible for a split second before it gets hidden - such was the bug.

Removed the remnant and it looks good now.


mightynimble:
@mreese84 two things that I forgot to mention during screensharing:
1. I changed the "reverb" link to "Earthdata Access" on "old browser page":
screen shot 2017-08-04 at 12 22 22 pm
2. I increased the width of the black background div in the above screenshot so everything is on the same line.

PR #467, opened 2017-07-21
EDSC-1537 Fail gibs cron job on error.

mightynimble:


mightynimble:
Still not sure why "only generation of JSON objects or arrays allowed" showed up in the response. Hopefully, with all the ```rescue```s, the error will be caught and converted to a readable error for the ops team.

PR #471, opened 2017-07-26
EDSC-1448: Updated dropdown and datepicker behaviors

jboukes:
The issue is that the temporal dropdown is remaining open when the user selects the spatial dropdown or 'clear filters'; this is unexpected behavior for the user.

It's a tricky issue as the dropdown has to have special code to keep it open, as the datepicker within the temporal dropdown will close it unexpectedly when a date is choosen; this special code is what is erroneously keeping the dropdown open when the user clicks the spatial dropdown or 'clear filters'.

The fix was to modify how the dropdown stays open and coordinates with the datepicker, as well as new code triggered by the spatial dropdown and 'clear filters' that programmatically closes the temporal dropdown in the case it is still open.

### DIFF
--- a/spec/features/temporal/temporal_selection_spec.rb
+++ b/spec/features/temporal/temporal_selection_spec.rb
@@ -12,6 +12,17 @@ describe "Temporal" do
     reset_search
   end
 
+  context 'When opening the temporal dropdown, and then opening the spatial dropdown' do
+    before :all do
+      click_link "Temporal"
+      click_link "Spatial"
+    end
+
+    it 'closes the temporal dropdown' do
+      expect(page).to have_no_content("Start")
+    end
+  end
+
   context "range selection" do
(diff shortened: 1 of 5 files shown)


mightynimble:
Looks good. Please add a test in one of the spec files in spec/features/temporal folder.

jboukes:
@mightynimble Thank you - will do!  Once I have the test added I'll ping you, thanks!

spec/features/temporal/temporal_selection_spec.rb
```diff
@@ -12,6 +12,17 @@
     reset_search
   end
 
+  context 'When opening the temporal dropdown, and then opening the spatial dropdown' do
+    before :all do
+      click_link "Temporal"
+      click_link "Spatial"
```
mightynimble: insert ```expect(page).to have_content("Start")``` before clicking "spatial"

PR #475, opened 2017-07-26
EDSC-1544 Fix a flaky test that fails on master.

mightynimble:


mreese84:
Why was this flakey? If we add "wait for xhr" do we still need to point to the actual URLs instead of the `have_link` method?

mightynimble:
@mreese84 adding ```wait_for_xhr``` didn't fix it. There is no ```ajax``` calls fired when "metadata" is clicked. It takes time for the panel to switch from information to metadata and check xpath reliably assert that the links are there.

mreese84:
Gotcha.

PR #476, opened 2017-07-26
EDSC-1572 Fix a typo

mightynimble:
This PR fixes the single granule download issue in UAT and PROD

### DIFF
--- a/app/assets/javascripts/models/data/granule.coffee
+++ b/app/assets/javascripts/models/data/granule.coffee
@@ -23,7 +23,7 @@
       null
 
     download_now_url: ->
-      return link.href for link in @links when link.rel.indexOf('/data#') != -1 if @links? @links.length > 0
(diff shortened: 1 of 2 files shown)


mreese84:
@mightynimble, do we need to add a test showing that the previous code didn't work? I was chatting with Doug today and he said the idea is to write a test showing that it was broken, then fix it, then write a test showing that it works.

mightynimble:
It will fail with previous code.

mreese84:
@mightynimble, so to be clear, and to ease some concern over the commit message... this is not an edge case. It reads like the test checks the core functionality of the feature, which is good. But this is not an edge case. This is a highly desired feature that will most certainly be used often and promoted by those that are encouraging users to switch to EDSC. Are you confident that this feature is fully tested?

mightynimble:
Yes, I am sure it is working now. And I know it is an important feature.
The latest commit fixes a typo and the issue that Andy reported was indeed an edge case :)

PR #480, opened 2017-08-03
EDSC 1568 Improved tour text to be more descriptive

jennywagg:


jennywagg:
@jboukes Thanks! 

@mightynimble 1. Ok, moved the inline styles into classes. 2. That makes sense. I would copy the leaflet images into our image dir, but I'm seeing that leaflet icons are linked in other css stylesheets in the same manner I did, with the version numbers. Should I keep it for the sake of consistency, since you said we may not often upgrade leaflet anyway?

mightynimble:
@jennywagg yeah, keep the version number in the url then. Let's deal with the broken links later if we did upgrade leaflet. Copying over the images is not the best resolution anyways. Don't forget to push your css changes before merge.

app/assets/javascripts/modules/help.js.coffee
```diff
@@ -61,30 +61,61 @@ ns = @edsc.models.page
       showNext: true
       },{
       title: "Facets"
-      content: "<p>Refine your search further with available facets, such as:
-                <div style='margin-left: 15px;'><ul style='list-style-type: disc;'>
-                  <li>Features - subsettable, near-real-time or has imagery</li>
+      content: '<p>Refine your search further with available facets, such as:
+                <div class="tour-indent"><ul class="bullet-list">
```
mreese84: Let's change this class from "tour-indent" to "tour-list". Remember, class names should describe content, not styles, as much as possible.

app/assets/javascripts/modules/help.js.coffee
```diff
@@ -61,30 +61,61 @@ ns = @edsc.models.page
...
       placement: 'right'
       showNext: true
       top: null
     }, {
+
       title: "Map Tools"
-      content: '<p>Use these standard map tools to configure and position the map as well as enable certain spatial search tools.</p>'
+      content: '<p>Use these map tools to configure and position the map as well as enable certain spatial tools.
+                  <table><tr>
+                    <td width="40%">
```
mreese84: You can kill both width rules.

app/assets/javascripts/modules/help.js.coffee
mreese84: Can you also move this `ul` down to a new line? I don't really like having block level elements on the same line. So I think the markup should look like this:
```
<div class="tour-list">
  <ul class="bullet-list">
    <li>blah blah</li>
  </ul>
</div>
````
I think that makes things easier to read and see the structure.

File: app/assets/javascripts/modules/help.js.coffee [hunk 3]
```diff
@@ -61,30 +61,61 @@ ns = @edsc.models.page
...
     }, {
       title: 'Toolbar'
-      content: '<p>Use the options available (upon logging in) in the application toolbar to view recent downloads, saved projects, and profile
-      information. You can provide feedback using our feedback module.</p>'
+      content: '<p>This is the end of the tour. Using this toolbar, you can:
+        <div class="tour-indent">
+          <ul>
+            <li class="toolbar-draw-list"><i class="fa fa-bullhorn"></i> Provide feedback</li>
+            <li class="toolbar-draw-list"><i class="fa fa-lock"></i> Login to Earthdata for more user tools</li>\
+          </ul></div></p>
```
mreese84: Let's move these closing element tags down to their own lines (and make sure they are indented the same as their opening tags.

app/assets/javascripts/modules/help.js.coffee
```diff
@@ -61,30 +61,61 @@ ns = @edsc.models.page
...
+        <p>When logged in, you can also:
+        <div class="tour-indent"><ul class="bullet-list">
```
mreese84: Move this `ul` to it's own line and make the closing tags match as well.

PR #482, opened 2017-08-09
Added close button to 'Choose your temporal constraint' modal

jennywagg:


mightynimble:
Do we need a new test to make sure that the modal can be closed by clicking the 'x'?

mreese84:
Yes, please add a test for this ticket. @mightynimble, you may do a screenshare session to walk Jenny through our testing protocol and how we write tests.

Also, @jennywagg, please format your commit messages like this "EDSC-XXXX: Your message here" where XXXX is the ticket number. Doing that enables some linking between various systems so that we can tie tickets to builds, releases, and commits. Thanks!

jennywagg:
@mreese84 Got it, will do!

PR #492, opened 2017-08-23
EDSC-1290: Update Metadata Dropdown HTML View Text

jsiarto:
- @mightynimble fixed the `metadata_url()` function to look for a `data-` attribute instead of the link text.
- Updated text to "Web View"


jennywagg:
@mreese84 I've updated the tests and they are passing now.

jboukes:
Code looks quite reasonable - I'd approve too however am not on the approver list :)

PR #493, opened 2017-08-23
EDSC-1194 Clear Filters button at top of page clears granule filters

jennywagg:


mreese84:
@jennywagg, this functionality looks good. However, we do need to add a test for this. We need a test that proves that when a user clicks clear filters, all filters (including granule filters) are removed. You can see there's a test that proves the clear filters button clears temporal filters here: `earthdata-search/spec/features/address_bar_spec.rb`

Once that test is added and the build is green this should be good to go.

mreese84:
@jboukes @mightynimble, please review the spec test that Jenny added. I'm approving from a functionality standpoint. @jennywagg, wait until you hear from at least John or Bo before merging. Thanks!

spec/features/address_bar_spec.rb
```diff
@@ -132,6 +132,48 @@ def query_string
         expect(page.current_url).not_to have_content("sb=0%2C0%2C10%2C10")
       end
```
mightynimble: we should add an ```after :all do``` block to re-set the bounding_box for use in your new test. If the clearing filter test (line 125) is executed before the new test, the new one will not have a collection-level filter bounding box to begin with.

```ruby
after :all do
  visit '/search'
  wait_for_xhr
  create_bounding_box(0, 0, 10, 10)
  wait_for_xhr
end
```

spec/features/address_bar_spec.rb
```diff
@@ -132,6 +132,48 @@ def query_string
...
     end
+
+    context 'setting granule filters' do
+      before(:all) do
+        visit '/search?q=C14758250-LPDAAC_ECS'
+        wait_for_xhr
+        dismiss_banner
```
mightynimble: Don't use ```visit``` again here - this will remove the bounding box filter in the url. ```q``` will be the only filter left when the page loads.

Replace line 138 to 140 with this: (run it locally to make sure it works)
```
expect(page.current_url).to have_content("sb=0%2C0%2C10%2C10")
fill_in "keywords", with: 'C14758250-LPDAAC_ECS'
wait_for_xhr
```

PR #496, opened 2017-08-28
EDSC 1557 - Added null checks to prevent script generation crash

jboukes:
Collection 'C1000000044-EDF_TS2' (which is only on SIT, so use cmr_env=sit to test) has a quirk wherein the script generation is crashing due to some unexpected null values. 

To test, do a data retrieve with 'C1000000044-EDF_TS2' on SIT, choose direct download, then click 'Download Access Script'.  The page with the script should render without hitting an exception.

Please note that the script created is somewhat questionable as the URL's given within the collection do not appear to be valid; that said, that is nothing our side of the code can change, and so correcting those URL's is (I believe) not part of this ticket.


mightynimble:
Take a look at the granule page of this collection. You will see that some granules are downloadable. The script examines the first granule of the collection, which in this case is not downloadable. With the nil check, users won't be able to download the granules using the script.

Instead of looking at only the first granule, we should look at 'x' number of granules to determine if a download script can be generated. I suggest 'x' to be the number of loaded granules on the page. @mreese84?

Or a simple fix will be that if the first granule of a collection is not downloadable, we don't show the download script button at all.

app/views/granules/prepare_script.html.erb
```diff
@@ -55,6 +55,7 @@ netrc=$(mktemp netrc.XXXXXXXXXX)
 chmod 0600 "$cookiejar" "$netrc"
 function finish {
   rm -rf "$cookiejar" "$netrc"
+  <% if @first_link %>
   <% if @first_link.start_with? 'ftp' %>
```
mreese84: I'll leave it to @mightynimble to comment on whether or not this is the best way to add a check... but if it is, I think we should indent the nested `if` statement for readability. So just indent lines 59-67 by one (and also the other nested ifs added by this PR) and this is good to go.

mreese84:
I think what @mightynimble is suggesting is fair. I think that this use case is pretty edge case, but it is possible. So yeah, checking for download links against loaded granules would be better.

Also... we need to add a test showing that we are checking for download links. A test showing what happens when we find links and a test for what happens when no links are found.

jboukes:
@mightynimble, @mreese84, is this close to what you two were suggesting?  I am not at all confident that I got it right here - take a look at let me know what you think.

app/views/granules/_generic_download.sh.erb
```diff
@@ -11,7 +11,7 @@ setup_auth_curl() {
     status=$(curl -s -z "$(date)" -w %{http_code} <%= raw @first_link %> | tail -1)
     if [[ "$status" -ne "200" && "$status" -ne "304" ]]; then
         <% # Check if this is an FTP retrieval %>
-        <% if @first_link.start_with? 'ftp' %>
+        <% if (@first_link) && (@first_link.start_with? 'ftp') %>
```
mightynimble: ```if @first_link && @first_link.start_with?('ftp')```
 - https://github.com/bbatsov/ruby-style-guide

mightynimble:
Looks good, just some styling recommendations.

app/views/granules/_generic_download.sh.erb
jboukes: Ahhh - tried that before, but I didn't have parenthesis around the 'ftp' and it threw an exception - the way you did it there works well.  Updated ;)

mightynimble:
Looks good!

PR #499, opened 2017-08-31
EDSC-1553 Scrub services.yml from git history

mightynimble:


mightynimble:
@mreese84 this is a remake of EDSC-1553. It should remove services.yml from git history. Again, there is nothing to 'review' since this should have been merged directly to master and do a 'forced' push.

PR #501, opened 2017-09-04
EDSC-1593 Make tests ready for 'trusty'

mightynimble:
TravisCI is switching the default ubuntu to 14.04 'trusty'. We need to make changes accordingly.
However, due to the (very) old versions of test-related gems (capybara, rspec, capybara-webkit, etc.), the tests fails with broken pipe error.

Broken pipe error was fixed by making changes to ```.travis.yml``` to install gstreamer and related libs explicitly.

Then the build stalls due to possible capybara-webkit/qt issues. Travis seems to have some issues as well - they blogged a new option ```group: deprecated-2017Q3``` with sudo enabled. Also split the tests into more groups and increased the default wait time to 15 sec per suggestion by TravisCI.

Other approaches tried but didn't work out:
 - use poltergeist driver
 - use osx build platform (with qt@5.5)
 - update test gems to higher/latest versions

TODO: we need to update our gems to the latest version. A few issues between capybara-webkit and qt have been fixed in later releases (that is why googling hardly returned any results)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ script:
   - bundle exec rake travis:ci
 env:
   global:
-    - CI_NODE_TOTAL=9
+    - CI_NODE_TOTAL=7
   matrix:
     - CI_NODE_INDEX=0
     - CI_NODE_INDEX=1
@@ -25,8 +25,6 @@ env:
     - CI_NODE_INDEX=4
     - CI_NODE_INDEX=5
     - CI_NODE_INDEX=6
-    - CI_NODE_INDEX=7
-    - CI_NODE_INDEX=8
     - JASMINE=true
 branches:
   only: # Only build master. Pull requests to master also get built.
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,25 +1,30 @@
...
   - bundle exec bin/setup
   - bundle exec rake travis:ci
-#after_failure:
-#  - cat ./log/test.log
-before_install:
-  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
 env:
   global:
-    - CI_NODE_TOTAL=4
+    - CI_NODE_TOTAL=7
```
mreese84: Curious what bumping this up to 7 does.

mightynimble: The build will stall due to unknown reasons. It could be Travis who killed the worker due to system resource issues (memory, disk, etc). Or it might be that we are using very old versions test related gems.

mightynimble: Set it to 7 shortens each job but makes the total build time longer since we can run 5 jobs concurrently for free.

PR #504, opened 2017-09-14
use portal org for site_name

eniad:
Action: Updating site_name to use portal org defined in config/portals.yml instead of portal title.

Reasoning: The link at the bottom of portal search results allows a user to escape the portal to the full earthdata-search. This text must get it's wording from the portal title. The text is "Looking for more collections? Leave the Search Portal". This is ambiguous. Preferred is 'Leave the ORNL DAAC Portal'.

To consider: Other uses of 'site_name' seem to make more sense with the updated site_name, although I may have missed something. Here's the list of 'site_name' locations I found:
./node_modules/jodid25519/jsdoc.json
./app/views/data_access/retrieval.html.erb
./app/views/search/index.html.erb
./app/helpers/portal_helper.rb (changed file)

### DIFF
--- a/app/helpers/portal_helper.rb
+++ b/app/helpers/portal_helper.rb
@@ -22,7 +22,7 @@ module PortalHelper
 
   def site_name
     if portal?
-      "#{portal['title'] || portal_id.titleize} Portal"
+      "#{portal['org'] || portal_id.titleize} Portal"
     else
       "Earthdata Search"
     end
--- a/app/helpers/portal_helper.rb
+++ b/app/helpers/portal_helper.rb
@@ -20,6 +20,14 @@ module PortalHelper
     end
   end
 
+  def site_org
+    if portal?
+      "#{portal['org'] || portal_id.titleize}"
+    else
+      "Earthdata Search"
+    end
+  end
+
   def site_name
     if portal?
       "#{portal['org'] || portal_id.titleize} Portal"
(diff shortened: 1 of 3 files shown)


mreese84:
@eniad, I've been thinking about this. I think there are some cases where org makes sense and some cases where a more specific "title" makes sense. How about we split the difference and use both? So in ORNL's case, your portals would show the following text...

- Looking for more collections? Leave ORNL DAAC's CARVE Portal
- Looking for more collections? Leave ORNL DAAC's AirMOSS Portal
- Looking for more collections? Leave ABoVE's Search Portal
- Looking for more collections? Leave ORNL DAAC's Search Portal

Thoughts? If you're on board I can push an update to your branch for you to check out.

eniad:
@mreese84 I'm happy with that. Using both gets rid of the ambiguity. Both will be great unless someone uses really long strings for them.

mreese84:
@eniad, ok... I just pushed up a change. Take a look at let me know what you think.

Also, in the future, when submitting a pull request that changes functionality or even something as simple as link text, you'll want to do a quick scan to see if any tests are impact. You can run `rspec` to run the full test suite to see if any tests were broken by your change. You can see in my push that I needed to update a test as EDSC's test suite was looking for a specific line of text. Changing that text broke the test, which would block us from merging it into our master branch.

Anyways... the new push is currently building. Once I get the "looks good" from you and a green build from TravisCI I'll get it merged in.

mreese84:
@eniad, just aheads up that I'm looping back around on this. I think that your fork is out of date as this PR had tons of test failures. Either way, #512 is green and just needs to be reviewed by someone on the team. Once that's done it'll be merged to SIT and we'll be all set.

PR #505, opened 2017-09-17
EDSC-1634: Draft of initial modal and reverb retirement logic

jboukes:


app/views/search/_reverb_retirement_modal.html.erb
```diff
@@ -0,0 +1,17 @@
+<div class="modal fade sitetour" id="reverbRetirementModal" tabindex="-1" role="dialog">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <a href="#" class="close modal-close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></a>
+        <h2 class="modal-title" id="reverbRetirementModalLabel">Reverb is Retiring - Welcome to Earthdata Search</h2>
+      </div>
+      <div class="modal-body modal-body-welcome">
+        <p>Reverb is retiring.  Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. </p>
```
mreese84: Can you change this text to...

> You have been redirected to Earthdata Search because Reverb is being retired on January 1st, 2018. Earthdata Search is replacing Reverb. If you would like to go return to Reverb for the time being, you may do so by clicking "Return to Reverb" below. Or you can choose to use Earthdata Search for now. At any point you may opt to return back to Reverb. We will remind you periodically that Reverb is retiring as its retirement date draws closer. If you would like to take a tour of Earthdata Search's main features, you may do so after choosing to stay on Earthdata Search. If you would like a more in-depth demonstration of Earthdata Search, [watch this recorded webinar](https://www.youtube.com/watch?v=eJFHMestpCo).

app/views/search/_reverb_retirement_modal.html.erb
```diff
@@ -0,0 +1,17 @@
+<div class="modal fade sitetour" id="reverbRetirementModal" tabindex="-1" role="dialog">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <a href="#" class="close modal-close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></a>
+        <h2 class="modal-title" id="reverbRetirementModalLabel">Reverb is Retiring - Welcome to Earthdata Search</h2>
+      </div>
+      <div class="modal-body modal-body-welcome">
```
mreese84: After this `div` and before the first `p` in the modal content, we should place a heading that says "Welcome to Earthdata Search"

app/views/search/_reverb_retirement_modal.html.erb
jboukes: @mreese84 Done :)

mightynimble:
Overall a few things and ideas:
1) I personally would like this redirect to be implemented like a hat on top of EDSC which can be taken off after Reverb retires.
 - there is no need to add new file(class)
 - move everything as close as possible (to prevent the possibility that later we may want to insert some piece of code in between which will make the revert harder)
2) Even though we can test this locally, it doesn't mean that it will work 100% in sit/uat/prod. We need to prepare for that to the maximum that minimum code change should happen if it does happen to fail.
 - inject ENVs
3) Add missing features

* I will comment details in the PR on the line.

app/assets/javascripts/models/page/search_page.js.coffee
```diff
@@ -12,6 +12,8 @@
 #= require models/ui/state_manager
 #= require models/ui/feedback
 #= require models/ui/sitetour
+#= require models/ui/reverb_retirement
```
mightynimble: It can be moved into this file

app/assets/javascripts/models/page/search_page.js.coffee
```diff
@@ -12,6 +12,8 @@
...
+#= require util/js.cookies
```
mightynimble: Require the js.cookies right after jquery loads which means a different file location for js.cookies is desired.

app/assets/javascripts/models/page/search_page.js.coffee
```diff
@@ -31,12 +33,14 @@ ns.SearchPage = do (ko
...
                     SiteTourModel = ui.SiteTour
+                    ReverbRetirementModel = ui.ReverbRetirement
                     url = @edsc.util.url
                     StateManager = ui.StateManager) ->
   current = null
 
   preferences = new PreferencesModel()
-  sitetour = new SiteTourModel();
+  sitetour = new SiteTourModel()
+  reverbRetirement = new ReverbRetirementModel()
```
mightynimble: this shouldn't be a new class and don't initialize it here.

app/assets/javascripts/models/page/search_page.js.coffee
```diff
@@ -50,7 +54,11 @@ ns.SearchPage = do (ko
     $('.launch-customize-modal').click ->
       $('#customizeDataModal').modal('show')
 
-    preferences.ready.done(-> initModal()) if !window.edscportal
+    reverb = ["http://echo-reverb-rails.dev", "https://testbed.echo.nasa.gov", "https://api-test.echo.nasa.gov", "https://testbed.echo.nasa.gov", "https://reverb.echo.nasa.gov"]
```
mightynimble: Move everything related to showing modal, setting the cookie and writing back the metrics here inside ```$(document).ready``` so that they are performed only after user sees the page.

mightynimble: Set the reverb URLs in config file, not here. Also leave the option to ops so that they can inject the URL from bamboo

app/assets/javascripts/models/page/search_page.js.coffee
```diff
@@ -50,7 +54,11 @@ ns.SearchPage = do (ko
...
+    if $.inArray(document.referrer, reverb) != -1 && Cookies.get('ReadyForReverbRetirement') != 'true'
```
mightynimble: Match the URLs fetched from the services config.

app/assets/javascripts/models/page/search_page.js.coffee
```diff
@@ -50,7 +54,11 @@ ns.SearchPage = do (ko
...
+      $('#reverbRetirementModal').modal('show')
+    else
```
mightynimble: Use a separate ```if``` to separate the redirect from the tour (or anything).

app/assets/javascripts/models/page/search_page.js.coffee
```diff
@@ -66,6 +74,7 @@ ns.SearchPage = do (ko
         projectList: new ProjectListModel(@project, @collections)
         feedback: new FeedbackModel()
         sitetour: new SiteTourModel()
+        reverbRetirement: new ReverbRetirementModel()
```
mightynimble: Again, remove the class and put everything into ```$(document).ready```

app/assets/javascripts/models/ui/reverb_retirement.coffee
```diff
@@ -0,0 +1,19 @@
+#= require util/js.cookies
```
mightynimble: This file is not needed.

app/assets/javascripts/models/ui/reverb_retirement.coffee
```diff
@@ -0,0 +1,19 @@
...
+
+ns = @edsc.models.ui
+data = @edsc.models.data
+
+ns.ReverbRetirement = do (ko) ->
+  class ReverbRetirement
+    constructor: ->
+
+    returnToReverb: () =>
+      Cookies.set('ReadyForReverbRetirement', 'false', { expires: 7 })
```
mightynimble: The cookie expiration should be longer than 7 days. It should expire after Reverb retires.

app/assets/javascripts/models/ui/reverb_retirement.coffee
```diff
@@ -0,0 +1,19 @@
...
+      $('#reverbRetirementModal').modal('hide')
```
mightynimble: I might be wrong but I think the modal will close when the button is clicked. No need to make it close

app/assets/javascripts/models/ui/reverb_retirement.coffee
```diff
@@ -0,0 +1,19 @@
...
+      window.location.replace(document.referrer)
+    
+    stayWithEDSC: () =>
+      Cookies.set('ReadyForReverbRetirement', 'true', { expires: 7 })
```
mightynimble: set longer expiration.

app/assets/javascripts/util/js.cookies.js
```diff
@@ -0,0 +1,165 @@
+/*!
```
mightynimble: Require the file in search.js.coffee and make it load right after jquery, which may require it to be moved to app/assets/javascripts

app/views/search/_reverb_retirement_modal.html.erb
```diff
@@ -0,0 +1,17 @@
...
+        <a href="#" class="close modal-close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></a>
+        <h2 class="modal-title" id="reverbRetirementModalLabel">Welcome to Earthdata Search</h2>
+      </div>
+      <div class="modal-body modal-body-welcome">
+        <p>You have been redirected to Earthdata Search because Reverb is being retired on January 1st, 2018. Earthdata Search is replacing Reverb. If you would like to go return to Reverb for the time being, you may do so by clicking "Return to Reverb" below. Or you can choose to use Earthdata Search for now. At any point you may opt to return back to Reverb. We will remind you periodically that Reverb is retiring as its retirement date draws closer. If you would like to take a tour of Earthdata Search's main features, you may do so after choosing to stay on Earthdata Search. If you would like a more in-depth demonstration of Earthdata Search, watch this recorded webinar.</p>
+      </div>
+      <div class="modal-footer">
+        <button class="button" data-bind="click: $root.ui.reverbRetirement.stayWithEDSC">Stay with EDSC <i class="fa fa-arrow-circle-left"></i></button>
```
mightynimble: @mreese84 's question - should these two buttons use different styles?

jboukes:
I agree with your second point and it refers to the first - the new class *is* the hat!  This PR will be easy to back out of once we'd done with the migration; just delete the class, restore the search_page, and then snipe the 'includes' and you're good!  Plus, an OO approach seems reasonable given that this app is OO all over the place.

It has been my hope that we can test this on SIT.  This depends on the Reverb code though - can you structure it so that the Reverb test beds send their redirects to SIT?  If we could do that, then we can test from SIT which I believe is valid ( @mreese84 ?).

I have not said this before but I feel there is a minor risk in whether or not the referrer has been stripped from the sender on either app; some apps are done that way for security reasons.  Since our apps are very security conscious, it almost wouldn't surprise me.  Getting this tested on SIT would be a good way to confirm that document.referrer is populated with both apps, as well as testing functionality.

I'll re-read #508; however, if you could point out in the code where features are lacking that would be a big help.

Thank you, Bo.

mightynimble:
@jboukes Yeah, the stand alone file sounds good to me. But you need to address the issues in the file.

app/assets/javascripts/models/page/search_page.js.coffee
jboukes: Mmmmmm...but wouldn't that interfere with the 'hat' philosophy?  Storing the URL's in a config file seems - embedded.  We won't need them any longer after this migration, so why keep them stored so deeply?

app/views/search/_reverb_retirement_modal.html.erb
```diff
@@ -0,0 +1,17 @@
...
+        <button class="button" data-bind="click: $root.ui.reverbRetirement.returnToReverb">Return to Reverb <i class="fa fa-arrow-circle-right"></i></button>
```
mightynimble: Make these <a> instead of <button> and set ```href``` with the reverb url in config.

mightynimble:
Missing features and how to add them back: (see PR #508 )
1. ```......models/page/search_page.js.coffee``` was renamed to ```models/page/search_page.js.coffee.erb``` to use ENV values.
2. The ```models/page/search_page.js.coffee.erb``` also shows how to make calls to the rails app to write metrics
3. If there are changes needed for the modal button, new css should be put in ```app/assets/stylesheets/modules/modal.css.scss```
4. Add actions to ```app/controllers/search_controller.rb```
5. Set ```href``` to the modal buttons and simplify on the js side
6. Add 'back to reverb' button (```app/views/search/index.html.erb``` and ```app/views/shared/_back_to_reverb.html.erb```)
7. Add new routes in rails: ```config/routes.rb```
8. New configs in ```onfig/services.yml.erb```

app/assets/javascripts/models/page/search_page.js.coffee
jboukes: My concern is that the two modals would compete in some unhealthy way.  What if the site tour modal shows up on *top* of the reverb modal?  By doing an if/then, the site tour modal is effectively momentarily suppressed in the case of a Reverb referral.

mightynimble:
When everything is in place and working, I can test it locally.

app/assets/javascripts/models/page/search_page.js.coffee
mightynimble: What I am worried about is between now and early next year, there will be changes inserted among the lines that you did today which, when we do git revert or the manual eyeball revert, will have some risks of modifying existing features and/or not to be removed completely. Add new values in services makes the changes scattered, yes, but it won't hurt us when we do the revert.

app/views/search/_reverb_retirement_modal.html.erb
jboukes: An href is a direct link; at this step in the logic, the cookie has not been set.  The data-bind sends the logic to the 'hat' to do its work, which then forwards the user appropriately.

mightynimble: A link set for an <a> tag is fine. Cookie should be set 'onclick', meaning it should happen only after user actually determines that he/she wants to stay/leave the site. (see 5) in the 'missing features' comment below)

mightynimble: also we want to redirect the user to the reverb home page where it has a "convert the query to edsc" that user can use.

mightynimble:
The redirect modal will show up before the tour modal since it is loaded right after dom loads. At this time, none of ajax calls has been fired yet.

app/assets/javascripts/models/page/search_page.js.coffee
mightynimble: The redirect modal will show up before the tour modal since it is loaded right after dom loads. At this time, none of ajax calls has been fired yet.

app/assets/javascripts/models/page/search_page.js.coffee
jboukes: You're right - doesn't need it after all!  Done in next push.

app/assets/javascripts/models/ui/reverb_retirement.coffee
jboukes: I think you're right - I mean, even if it doesn't disappear, the next line is the redirect back to Reverb which refreshes the page.

I'll delete it and we'll see how it looks when its on SIT.

jboukes:
@mightynimble - I made some updates and I believe we are closer, however there are a some requests that don't seem appropriate given the difference in techniques.  We took two different approaches to solving this problem - which is neat, to me - and this approach simply doesn't require some of the changes that yours did.  We don't need ENV values, for example, as the Reverb links are stored in a simple easy-to-delete array in the class object.  

I am hoping you find this reasonable - it's just a different approach.  I think the ENV approach is neat; though I'd argue that using an object is more plug-and-play (easier to remove later).

1. ......models/page/search_page.js.coffee was renamed to models/page/search_page.js.coffee.erb to use ENV values.  -->  **Since we're not using ENV values, not needed**
2. The models/page/search_page.js.coffee.erb also shows how to make calls to the rails app to write metrics --> **There doesn't appear to be any metric collections mentioned in the acceptance criteria, though I'm open to gilding it with metrics after we've settled the required stuff**
3. If there are changes needed for the modal button, new css should be put in app/assets/stylesheets/modules/modal.css.scss  **Noted**
4. Add actions to app/controllers/search_controller.rb  **This approach uses an object with knockout bindings instead of routes and services**
5. Set href to the modal buttons and simplify on the js side **This approach uses an object with knockout bindings that handle the href functionality**
6. Add 'back to reverb' button (app/views/search/index.html.erb and app/views/shared/_back_to_reverb.html.erb)  **Done!**
7. Add new routes in rails: config/routes.rb **This approach uses an object with knockout bindings instead of routes and services**
8. New configs in config/services.yml.erb **Again, not needed with this approach**

mightynimble:
ENV is needed for the ops team to change the URL value in bamboo without a code change on our side.

app/assets/javascripts/models/ui/reverb_retirement.coffee
jboukes: It kind of is now - I removed it from the search_page, so now it's only included upon this object instantiation (which of course happens in the ready event).

mightynimble:
Metrics are an after-thought and should be captured either using splunk or google analytics.

mreese84:
@jboukes, the metrics was something I thought of when talking through the Reverb side with Bo. Basically, we need to know how many people are saying "thanks, but no thanks" to EDSC. I'd like to know how many say "Take me back to Reverb" from the modal and how many try EDSC for a bit and _then_ hit the "take me back" button.

app/views/search/_reverb_retirement_modal.html.erb
jboukes: @mightynimble Okay - now have the redirect to reverb to its home page instead of the full referrer URL

jboukes:
>>"ENV is needed for the ops team to change the URL value in bamboo without a code change on our side."

Unless I'm missing something - this approach doesn't require the OPS team to change any values at all.  It's good from the gate.

app/views/search/_reverb_retirement_modal.html.erb
mreese84: Updated.

app/views/shared/_back_to_reverb.html.erb
```diff
@@ -0,0 +1,8 @@
+<a
+  id="back-to-reverb-link"
+  style="display: none;"
+  class="toolbar-button"
+  data-bind="visible: $root.ui.reverbRetirement.referrerIsReverb(), click: function() {$root.ui.reverbRetirement.returnToReverb('toolbar link')};"
```
jboukes: Just to proactively say 'why' here - the click event is wrapped in an anonymous function, as this prevents that function from being executed on page load.  Without it, the click event fires as soon as the page is generated, which is not desirable.

mreese84:
@jboukes, @mightynimble , what needs to be done to finish this ticket up? I know we need to add metrics. Anything else?

jboukes:
@mreese84 I believe the only thing remaining is the metrics.  I spent some time last night testing the routine (and did find a couple of bugs) and I do believe it's working.  I added a couple of temporary lines that will assist in testing.  In search_page.js.coffee, find:

# These lines for testing only - delete before merging!
# Object.defineProperty(document, "referrer", {get : => return "https://testbed.echo.nasa.gov"; });
# Cookies.set('ReadyForReverbRetirement', 'false')

The 'Object.define...' line can be used to set the referrer - this allows testing 'as if' the referrer came in from one of the Reverb instances.  The second line will clear your cookie from true to false, so that you can reset your test.  Using these two tools, I was able to run through the logic a few times and see it working.

As far as metrics - the issue is our metrics functions are on the back end as this logic is all on the front.  I've been working on setting up a new post route to the back end so that we can post an AJAX call from the front to set metrics.  I'm sure I can figure it out, however given that we want to move quickly, if @mightynimble might want to help with that part, that would be awesome.

config/routes.rb
```diff
@@ -60,12 +60,13 @@
   get 'data/status' => 'data_access#status'
   post 'data/remove' => 'data_access#remove', format: 'json'
 
+  post 'metrics' => 'search#metrics_event'
```
mightynimble: Route should be changed to something meaningful.

app/controllers/search_controller.rb
```diff
@@ -11,4 +11,13 @@ def extract_filters
...
   end
 
+  def metrics_event
+    Rails.logger.info "We made it!"
+    type = params[:type]
+    data = params[:data]
+    other_data = params[:other_data]
+    metrics_event(type, data, other_data)
+    # this response makes no sense I'm sure - just trying to get something that doesn't error out
+    respond_with(Rails.logger.info params.inspect, status: :ok)
```
mightynimble: respond_with :nothing

app/controllers/search_controller.rb
```diff
@@ -11,4 +11,13 @@ def extract_filters
     respond_with TextSearchClient.parse_text(params[:q])
   end
 
+  def metrics_event
+    Rails.logger.info "We made it!"
+    type = params[:type]
```
mightynimble: All params should be validated to be safe.

app/assets/javascripts/models/ui/reverb_retirement.coffee
```diff
@@ -16,14 +17,32 @@ ns.ReverbRetirement = do (ko) ->
...
       window.location.replace("https://" + document.referrer.match(/:\/\/(.[^/]+)/)[1])
     
     stayWithEDSC: () =>
       Cookies.set('ReadyForReverbRetirement', 'true', { expires: 90 })
-      # metrics go here once it's figured out...
-      # metrics_event('reverb_redirect', 'stay_in_edsc') %>
+
+      data['type'] = 'reverb_redirect'
+      data['data'] = 'stay_in_edsc'
+      data['other_data'] = ""
```
mightynimble: Add source here?

jboukes: This one doesn't need it - it's the "stay in edsc" choice which has only a single source (the modal).  It's only the routes back to reverb (toolbar versus modal) that need a source.

config/services.yml.erb
```diff
@@ -43,19 +43,25 @@ cwic_root: "http://cwic.wgiss.ceos.org/"
...
     enable_esi_order_chunking: true
+    reverb_root: "api-test.echo.nasa.gov"
   lab:
     enable_esi_order_chunking: true
+    reverb_root: "testbed.echo.nasa.gov"
   production:
     limited_collections: C14758250-LPDAAC_ECS, C190733714-LPDAAC_ECS, C1299783607-LPDAAC_ECS, C1299783800-LPDAAC_ECS, C1299783608-LPDAAC_ECS, C1299783630-LPDAAC_ECS, C1299783574-LPDAAC_ECS, C1299783609-LPDAAC_ECS, C1299783631-LPDAAC_ECS, C1299783579-LPDAAC_ECS, C1299783651-LPDAAC_ECS, C1299783684-LPDAAC_ECS
     enable_esi_order_chunking: true
+    reverb_root: "reverb.echo.nasa.gov"
```
mightynimble: *I think* these should be full URLs starting with ```https?:\\```

app/assets/javascripts/models/ui/reverb_retirement.coffee
```diff
@@ -5,14 +5,16 @@ ns = @edsc.models.ui
...
+                          ajax = @edsc.util.xhr.ajax,
+                          config=@edsc.models.data.config) ->
   class ReverbRetirement
     constructor: ->
 
     referrerIsReverb: () =>
       console.log "Checking referrer: " + document.referrer
       referrer = if document.referrer then document.referrer.match(/:\/\/(.[^/]+)/)[1] else false
-      reverb = ["echo-reverb-rails.dev", "testbed.echo.nasa.gov", "api-test.echo.nasa.gov", "testbed.echo.nasa.gov", "reverb.echo.nasa.gov"]
+      reverb = [edsc.config.reverb_url]
```
mightynimble: ```reverb``` will no longer be an array. For each env, only one URL is valid. For example, if current env is SIT, reverb url will start with https://testbed.echo.nasa.gov

app/assets/javascripts/config.js.coffee.erb
```diff
@@ -13,6 +13,9 @@
   match = /cmr_env=(\w+)/.exec(query)
   cmr_env = if match then match[1] else '<%= Rails.configuration.cmr_env %>'
 
+  
+  reverb_url = "<%= ENV['reverb_url']%>"
```
mightynimble: Grab a value from services.yml.erb as the default. You can refer to how borwse_scaler_url and other variables are set up below.

app/assets/javascripts/config.js.coffee.erb
```diff
@@ -13,6 +13,9 @@
...
+
   if cmr_env == 'uat'
     browse_scaler_url = "<%= ENV['browse_scaler_url_uat'] ||  Rails.configuration.services['earthdata']['uat']['browse_scaler_root'] %>"
```
mightynimble: If you start rails server like this: ```browse_scaler_url_uat=https://uat.url rails s```
The line above tries to set ```browse_scaler_url``` using the ENV set in command line being ```https://uat.url```.

If you start the rails server ```rails s```, ```ENV['browse_scaler_url_uat']``` will be nil. So the code continues to take a look at the second part ```Rails.configuration.services['earthdata']['uat']['browse_scaler_root']``` to try to get the default value from there.

config/services.yml.erb
jboukes: I had stripped them as the referrer is regex'd so as to be missing the leading https:// - I updated the code so that the ENV values are also regex'd before comparison and added the leading http?:\\ back - 'should' work.

app/assets/javascripts/config.js.coffee.erb
mightynimble: You need to ```export``` the variable too. See below in "exports =" lines.

config/services.yml.erb
mightynimble: It will work for sure. I can't remember 100% but I think if you put the link in href, it may treat the link as a relative url without the leading http.

app/assets/javascripts/config.js.coffee.erb
```diff
@@ -13,15 +13,21 @@
   match = /cmr_env=(\w+)/.exec(query)
   cmr_env = if match then match[1] else '<%= Rails.configuration.cmr_env %>'
 
+
```
mightynimble: Add this to line#16 
```reverb_url = "<%=ENV['reverb_root'] || Rails.configuration.services['edsc'][Rails.env]['reverb_root']%>"```
and remove line#18, 23, 27, 28, and 30, since ```cmr_env``` changes based on query (see line#12 and 13) and we don't want it to point to a different reverb if cmr_env is different than rails env.

File: app/assets/javascripts/models/page/search_page.js.coffee [hunk 27]
```diff
@@ -50,6 +52,14 @@ ns.SearchPage = do (ko
     $('.launch-customize-modal').click ->
       $('#customizeDataModal').modal('show')
 
+    # These lines for testing only - delete before merging!
```
mightynimble: cleanup the debug lines

app/assets/javascripts/models/ui/reverb_retirement.coffee
mightynimble: Remove line 1. The lib should be included earlier right after jquery loads.

app/assets/javascripts/models/ui/reverb_retirement.coffee
```diff
@@ -0,0 +1,49 @@
+#= require util/js.cookies
+#= require util/metrics
```
mightynimble: Why include ```util/metrics```?

app/assets/javascripts/models/ui/reverb_retirement.coffee
```diff
@@ -0,0 +1,49 @@
...
+                          config=@edsc.models.data.config) ->
+  class ReverbRetirement
+    constructor: ->
+
+    referrerIsReverb: () =>
+      console.log "Checking referrer: " + document.referrer
+      console.log "Checking referrer against ENV value: " + edsc.config.reverb_url
+      referrer = if document.referrer then document.referrer.match(/:\/\/(.[^/]+)/)[1] else false
+      reverb = if edsc.config.reverb_url then edsc.config.reverb_url.match(/:\/\/(.[^/]+)/)[1] else false
+      return referrer == reverb && referrer
```
mightynimble: If ```document.referrer``` is not from reverb but ```edsc.config.reverb_url``` is, will the function return ```true```? In this case, the referrer is apparently not reverb, then why should it return true?

app/assets/javascripts/models/ui/reverb_retirement.coffee
```diff
@@ -0,0 +1,49 @@
...
+      data['data'] = 'back_to_reverb'
+      data['other_data'] = "{source: '" + source + "'}"
+      ajax
+        data: JSON.parse(JSON.stringify(data))
+        dataType: 'json'
+        url: "/metrics"
+        method: 'post'
+        success: (data) ->
+          console.log data
+      window.location.replace("https://" + document.referrer.match(/:\/\/(.[^/]+)/)[1])
```
mightynimble: Rename this file to ```app/assets/javascripts/models/ui/reverb_retirement.coffee.erb``` and use ```edsc.config.reverb_url``` explicitly. (We talked about this in gotomeeting. It is _that_ issue.)

app/assets/javascripts/models/ui/reverb_retirement.coffee
```diff
@@ -0,0 +1,49 @@
...
+      Cookies.set('ReadyForReverbRetirement', 'false', { expires: 90 })
+      data['type'] = 'reverb_redirect'
+      data['data'] = 'back_to_reverb'
+      data['other_data'] = "{source: '" + source + "'}"
+      ajax
+        data: JSON.parse(JSON.stringify(data))
+        dataType: 'json'
+        url: "/metrics"
+        method: 'post'
+        success: (data) ->
```
mightynimble: Just a heads-up, if the controller returns ```:nothing```, the data will be nil here.

app/assets/javascripts/models/ui/reverb_retirement.coffee
```diff
@@ -0,0 +1,49 @@
...
+          console.log data
+      window.location.replace("https://" + document.referrer.match(/:\/\/(.[^/]+)/)[1])
+    
+    stayWithEDSC: () =>
+      Cookies.set('ReadyForReverbRetirement', 'true', { expires: 90 })
```
mightynimble: 90 days probably is still not long enough. Reverb retires on Jan 2nd or 3rd.

app/assets/javascripts/models/ui/reverb_retirement.coffee
```diff
@@ -0,0 +1,49 @@
...
+
+    referrerIsReverb: () =>
+      console.log "Checking referrer: " + document.referrer
+      console.log "Checking referrer against ENV value: " + edsc.config.reverb_url
+      referrer = if document.referrer then document.referrer.match(/:\/\/(.[^/]+)/)[1] else false
+      reverb = if edsc.config.reverb_url then edsc.config.reverb_url.match(/:\/\/(.[^/]+)/)[1] else false
+      return referrer == reverb && referrer
+
+    returnToReverb: (source) =>
+      Cookies.set('ReadyForReverbRetirement', 'false', { expires: 90 })
```
mightynimble: 90 days probably is still not long enough. Reverb retires on Jan 2nd or 3rd.

app/controllers/search_controller.rb
```diff
@@ -11,4 +11,14 @@ def extract_filters
     respond_with TextSearchClient.parse_text(params[:q])
   end
 
+  def log_metrics_event
+    if params[:type] && params[:data] then
```
mightynimble: Validate all the params here. (the gotomeeting issue)

app/controllers/search_controller.rb
```diff
@@ -11,4 +11,14 @@ def extract_filters
...
+      if params[:other_data] && params[:other_data] != "" then
+        metrics_event(params[:type], params[:data], eval(params[:other_data]))
```
mightynimble: why ```eval```?

app/controllers/search_controller.rb
```diff
@@ -11,4 +11,14 @@ def extract_filters
...
 
+  def log_metrics_event
+    if params[:type] && params[:data] then
+      if params[:other_data] && params[:other_data] != "" then
+        metrics_event(params[:type], params[:data], eval(params[:other_data])) 
+      else 
+        metrics_event(params[:type], params[:data])
+      end
+    end
+    render nothing: true, status: :ok, content_type: "text/html"
```
mightynimble: ```respond_with :nothing``` should be good enough. And content_type is not needed for ':nothing'.

app/assets/javascripts/models/ui/reverb_retirement.coffee
jboukes: >>"If document.referrer is not from reverb but edsc.config.reverb_url is, will the function return true?"

Nope!  ;)

app/controllers/search_controller.rb
jboukes: Beyond checking that they exist, what sort of validation are you looking for?

app/controllers/search_controller.rb
jboukes: It's sent in as a string; however the function in application_controller defines it as an hash.  Eval converts it so that the parameter is acceptable by metrics_event().

app/controllers/search_controller.rb
jboukes: It wasn't good enough - it's the first thing I tried.  :\  It throws this error:

NoMethodError (undefined method `nothing_url' for #<SearchController:0x007ffa413f44e8>):
  app/controllers/search_controller.rb:22:in `log_metrics_event'

It put me on the right track, though; it led me to the syntax you see here.  The syntax above, works - no exceptions and behaves.

app/assets/javascripts/models/page/search_page.js.coffee
jboukes: I'll need to know where that is :\

app/assets/javascripts/models/ui/reverb_retirement.coffee
jboukes: That's okay; only used for a console log entry.

app/assets/javascripts/models/ui/reverb_retirement.coffee
jboukes: Which line are you referring to in regards to using edsc.config.reverb_url explictely?  From what I can tell it already does that on line 16.

I am not sure why the file needs to be renamed as well.  :\

app/assets/javascripts/util/js.cookies.js
jboukes: Just moving the file to that directory loses it; where is it to be loaded right after jquery?

app/assets/javascripts/models/ui/reverb_retirement.coffee
mightynimble: what if there is no referrer? so ```referrer``` will be ```false``` and ```reverb``` will be an url. Then ```reverb && referrer``` will be ```false```. So that ```referrer == reverb && referrer``` will be equivalent to ```false == false``` which is ```true```. If my deduction is right, which is probably not the case, the function will think that ```referrerIsReverb``` when there is no referrer at all??

app/assets/javascripts/models/ui/reverb_retirement.coffee
mightynimble: This line. Depends on how you wanna make the change. It could be something _**like**_ ```window.location.replace("<%=ENV['reverb_root'] || Rails.configuration.services['edsc'][Rails.env]['reverb_root']%>")``` and rename it as an ```.erb``` file. Or you could say ```window.location.replace(edsc.config.reverb_url)```

app/assets/javascripts/util/js.cookies.js
mightynimble: There are three places that ```//= require jquery``` - ```search.js``` is the one that should require the cookie lib. Require the file after ```//= require jquery.color-2.1.2``` (line 4)

app/controllers/search_controller.rb
mightynimble: No need to check the existence, but validate _all_ the possible values that will be passed in to this param, using ```==``` or ```.eql?```. For example, if params[:type] can have values of 'a', 'b', or 'c'. Then in this controller action, you should have something in place like this:
```if params[:type] == 'a'
  # code
elsif params[:type] == 'b'
  # code
elsif params[:type] == 'c'
  # code
else
  # code```

Comment (mightynimble) [reply app/controllers/search_controller.rb]: We can talk about this in another gotomeeting.

File: app/controllers/search_controller.rb (hunk 35 above)
Comment (mightynimble) [reply app/controllers/search_controller.rb]: ```eval(params[:other_data])``` => ```{other_data: params[:other_data]}```

File: app/controllers/search_controller.rb (hunk 36 above)
Comment (mightynimble) [reply app/controllers/search_controller.rb]: You sure it is ```respond_with :nothing```, not ```render :nothing```?

**Comment by mightynimble** [comment]
@jboukes I added a couple of more inline comments to the PR. But really, there are only two major issues that should be updated. For the rest of the stuff, I just posted there for you to take a look.
1. Set up the ENV values
2. Security issues
3. The question on ```referrer == reverb && referrer```

File: app/assets/javascripts/models/ui/reverb_retirement.coffee (hunk 30 above)
Comment (jboukes) [reply app/assets/javascripts/models/ui/reverb_retirement.coffee]: In the gotomeeting, you had said that this line was fine as it was.  It was one of the first concerns we talked about.  You had a concern that it was using the referrer and were worried about a security concern; when I pointed out that the referrer was filtered by a regex, you said (paraphrased) "Oh - that's okay then." and we moved on.

What changed?  :\

File: app/assets/javascripts/models/ui/reverb_retirement.coffee (hunk 29 above)
Comment (jboukes) [reply app/assets/javascripts/models/ui/reverb_retirement.coffee]: If referrer is false then it will return false - same if they are both false - take a closer look at the return conditional (adding parenthesis for emphasis):

return (referrer == reverb) && (referrer)
evaluates to
return (false == false) && (false)
evaluates to
return (true) && (false)
evaluates to - false!  Both sides of the && must be true for it to be true.

File: app/assets/javascripts/util/js.cookies.js (hunk 14 above)
Comment (jboukes) [reply app/assets/javascripts/util/js.cookies.js]: Got it - I believe that'll work.

File: app/controllers/search_controller.rb (hunk 34 above)
Comment (jboukes) [reply app/controllers/search_controller.rb]: I'll have it check that "if params[:type] == 'reverb_redirect'" = right now there's only one 'type' coming from the frontend and that is it.

I think this function might be useful in the future for logging other front end events; when doing so though any new 'types' will have to be added to the conditional (minor programmatic overhead).

File: app/controllers/search_controller.rb (hunk 35 above)
Comment (jboukes) [reply app/controllers/search_controller.rb]: That didn't work as well.  Metrics output using that syntax:
[metrics] {"event":"reverb_redirect","data":"back_to_reverb","session":"780241f7be3da5cad2f823b7a094c805","timestamp":1506531924415,"{:source=\u003e\"modal link\"}":{"other_data":"{source: 'modal link'}"}}

Note the "\u003e\" stuff.

The line I used doesn't have that and is easier to read.  It works - why fix it?

File: app/controllers/search_controller.rb (hunk 36 above)
Comment (jboukes) [reply app/controllers/search_controller.rb]: Yes.  I tried it multiple times.  The code I have here works; the respond_with :nothing does not.

Comment (jboukes) [reply app/controllers/search_controller.rb]: I'll add - I'm aware that "respond_with :nothing" *should be* acceptable - I really don't know why it's throwing an exception.  However, the line used above absolutely does work and my feeling is that is what is important.

**Comment by jboukes** [comment]
@mightynimble We're even closer :)

I left a question on the ENV values for you - in the gotomeeting, we decided that using the referrer that is filtered through a regex was acceptable for URL forwarding in the case of 'return to reverb'.  Since the regex is applied, all security concerns are handled.  I still think that should be fine and I am not sure why we should do something other than what we agreed in that meeting.  

I added a 'type' check for the log metrics event which will filter out anything that isn't reverb related (future usages of this function will have to add their own 'types' to make them legitimate).

I moved that pesky js.cookies function to load after jQuery.

I also answered other various questions (and you need more info just let me know).

File: app/assets/javascripts/models/ui/reverb_retirement.coffee (hunk 30 above)
Comment (jboukes) [reply app/assets/javascripts/models/ui/reverb_retirement.coffee]: To add - I'll give in on this one and say 'uncle' if you really want it - it's just that we agreed to do it this way in the gotomeeting (since it has a regex filter it was deemed 'safe').  If you reaaaally want it though, I'll do it - we need to wrap this all up.  Let me know.  :)

File: app/assets/javascripts/search.js [hunk 37]
```diff
@@ -2,6 +2,7 @@
 //= require jquery
 //= require jquery_ujs
 //= require jquery.color-2.1.2
+//= require util/js.cookies
```
mightynimble: js.cookies is a vendor lib and therefore should be in the ```vendor``` dir. (take a look at that directory, we have all the js dependencies stored there.)

app/assets/javascripts/models/ui/reverb_retirement.coffee
mightynimble: I thought it was something else - like you set the referrer using the ENV first then use it in the redirect. In fact, there is no code to set this ```document.referrer``` so it is subject to others to set it to something else. That's the reason that I want what inside of ```window.location.replace``` to be a hard-coded ENV value.

app/controllers/search_controller.rb
mightynimble: I am not asking you to fix it. See my overall comment. We currently only need to address 1. security issues and 2. the ENV values which will fix some of the issues in 1.

app/controllers/search_controller.rb
```diff
@@ -13,10 +13,13 @@ def extract_filters
 
   def log_metrics_event
     if params[:type] && params[:data] then
-      if params[:other_data] && params[:other_data] != "" then
-        metrics_event(params[:type], params[:data], eval(params[:other_data])) 
-      else 
-        metrics_event(params[:type], params[:data])
+      if params[:type] == 'reverb_redirect'
+        if params[:other_data] && params[:other_data] != "" then
```
mightynimble: like what you did for ```:type```, do the same thing for ```:other_data```; find out all the possible values and whitelist them.

app/controllers/search_controller.rb
```diff
@@ -13,10 +13,13 @@ def extract_filters
...
+          
+          metrics_event(params[:type], params[:data], eval(params[:other_data])) 
+        else
```
mightynimble: scrub all the data in ```:data```, ```:other_data``` in this "default" block.

mightynimble:
Two issues and we are good to go.
1. Use the ENV value in window.href.replace
2. Sanitize params in the metrics controller.

app/assets/javascripts/models/ui/reverb_retirement.coffee
jboukes: I will update it, however - it's harmless as it is.  The regex makes it so.  Even if someone spoofs a referrer in, what could they do with it?  The best they could do is trigger the soft redirect link when it shouldn't come up.  This wasn't really a security concern.

However, I'll make the change, incoming in 5, 4, 3, 2...

app/controllers/search_controller.rb
jboukes: I will, though I think that's validation too far - metrics are nothing but key/value string pairs.  If we add other metrics handling on the front end, this strict validation could easily become unwieldy.

jboukes:
@mightynimble - those two items are now complete.

mreese84:
@mightynimble are you going to be able to review this today?

mightynimble:
Yeah, I will take a look this morning.

mreese84:
Awesome. Thanks man.

app/controllers/search_controller.rb
mightynimble: This is a public facing endpoint. If later we wanna use this at all (note all of these will be reverted), we should do the same. The data passed to default ```else``` should always be scrubbed. Plus, we mainly use google analytics for js metrics. Adding this endpoint is for quicker retrieval of metrics from splunk using reports.

mightynimble:
Everything looks good now. Only place that John should change is scrubbing all the data in ```else``` to prevent parameter injections.

app/controllers/search_controller.rb
jboukes: @mightynimble The data passed to the default else *is* scrubbed - the default else uses the type and data, please look at the conditional.  Both of those have been validated before it hits that else.

jboukes:
@mightynimble left a comment on that at the line - that data is *already* scrubbed.  The default else uses type and data, and both of those are validated before the else.

jboukes:
@mightynimble 
Please see the full conditional below.  The 'else' is an else of the other_data check - to hit it, the logic must have passed the validation of type and data.  The data you are concerned about has already been scrubbed by this logic and there's nothing more I can add.

if params[:type] && params[:data] then
      **if params[:type] == 'reverb_redirect'**
        **if params[:data] == 'back_to_reverb' || params[:data] == 'stay_in_edsc' then**
          if params[:other_data] && params[:other_data] != "" then
            if params[:other_data] == "{source: 'modal link'}" || params[:other_data] == "{source: 'toolbar link'}"
              metrics_event(params[:type], params[:data], eval(params[:other_data])) 
            end
          else # **other_data is empty, however data and type have been validated**
            metrics_event(params[:type], params[:data])
          end
        end
      end
    end

mightynimble:
Ship it!

PR #513, opened 2017-09-29
EDSC-1586-2: Card text overflow fix

jsiarto:


mreese84:
@jsiarto I'm not a fan of the current email key/value setup. What if we did something like this...

screen shot 2017-09-29 at 12 18 50 pm

PR #514, opened 2017-09-29
EDSC-1626 Create Dockerfile for EDSC

jennywagg:


Procfile
```diff
@@ -1,2 +1,2 @@
-web: env RAILS_ENV=$RAILS_ENV env PORT=$PORT bundle exec unicorn -E $RAILS_ENV -c config/unicorn.rb
-jobs: env RAILS_ENV=$RAILS_ENV bundle exec bin/delayed_job run
```
jonvel: How does the `delayed_job` figure out which `RAILS_ENV` to run in?  Or is that handled by the `.env` file?

Procfile
```diff
@@ -1,2 +1,2 @@
...
+web: bundle exec unicorn -E $RAILS_ENV -c config/unicorn.rb
+jobs: bundle exec bin/delayed_job run
```
mreese84: @jennywagg @mightynimble, I was reviewing this with Jon V and he made a comment about rails_env not being included with delayed jobs.

> Interesting - was the RAILS_ENV not needed for the `delayed_job`?

He was surprised by that. Is this intentional?

jennywagg: I got rid of this after discussing with Nathan Clark. My understanding is that the RAILS_ENV variable is handled by NGAP after my conversation with him, and he said these lines were not necessary for either web or jobs. I am open to the possibility we overlooked something if I don't fully understand the delayed_job.

PR #516, opened 2017-09-29
EDSC-1634 hotfix for the redirect issue in sit.

mightynimble:


mightynimble:
Tested in SIT. Works now.

PR #518, opened 2017-10-02
EDSC-1507 Group links by URLContentType, add type and subtype and oso…

mightynimble:
…rt URLs in same URLContentType, remove EDSC and Reverb links. Add tests and fixtures.

### DIFF
--- a/app/assets/stylesheets/modules/modal.css.scss
+++ b/app/assets/stylesheets/modules/modal.css.scss
@@ -31,7 +31,7 @@
   overflow-y: scroll;
 
   ul {
-    list-style: circle;
+    list-style: none;
     padding: 0 30px;
   }
 }
--- a/app/assets/stylesheets/modules/modal.css.scss
+++ b/app/assets/stylesheets/modules/modal.css.scss
@@ -46,6 +46,25 @@
   }
 }
 
+.modal-body-related-urls {
+
+  h4 {
+    margin: 20px 0 10px 0;
+    padding-bottom: 5px;
+    border-bottom: 1px solid $light-grey;
+  }
+
+  ul {
+    padding: 0;
+    margin: 5px 0 0 0;
+  }
+
+  p {
+    margin: 0 0 10px 0;
+  }
+
+}
+
 .modal-dialog {
   width: 800px;
   margin-top: 120px;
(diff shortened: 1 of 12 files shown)


mightynimble:
See the ticket and the linked ```EPR-181``` ticket for detail.
Use ```C1200230663-MMT_1``` (in sit) or ```C1238517344-GES_DISC```(in prod)

mreese84:
Functionality looks good. @jsiarto, when you are free, can you take a spin through this modal to update styles?

@jennywagg @jboukes... you can go ahead and review the rest of the PR.

app/assets/stylesheets/modules/modal.css.scss
```diff
@@ -31,7 +31,7 @@
   overflow-y: scroll;
 
   ul {
-    list-style: circle;
+    list-style: none;
```
mreese84: Are you sure you want to make this change here? Won't this affect all modal lists? In some cases we will probably want these markers.

mightynimble: No - I am not sure about any css changes. It is up to you and @jsiarto. But I know that the dot is used only in cwic 'learn more' modal.

mreese84: Ok. So @jsiarto, when you start styling these related url links, just make sure that this `list-style` rule is in a place that doesn't bork other things.

mreese84:
@jsiarto, I can't tell if it's just me, but I find it a little confusing that the URL type is listed above the URL itself. When I'm looking at a list of related urls I found myself having to think about which label goes with which URL. So it's either a matter of the order in which the information is presented or a matter of the spacing between the type/url (the space between each `arrow-tag-group-list` and `p`) being too close to the space between each url/type listing (the space between each `li`).

What do you think?

jsiarto:
@mreese84 I could increase the spacing -- I didn't like the URL above, because it wasn't the strongest part of the design (the type was bolder). The URL itself really doesn't matter--people are just clicking a link. The most important information here is what _type_ of URL it is.

mreese84:
@jsiarto let's not worry about the spacing for now then. Let's see how this performs and then go from there.

PR #519, opened 2017-10-03
EDSC-1645: Cleaned up DOI badge

jboukes:
Added code to strip out instances of 'http://' and 'https://', and also added a conditional to check for presence of base url 'dx.doi.org' so that it doesn't contain that string twice.

### DIFF
--- a/app/presenters/collection_details_presenter_umm_json.rb
+++ b/app/presenters/collection_details_presenter_umm_json.rb
@@ -14,8 +14,7 @@ class CollectionDetailsPresenterUmmJson < DetailsPresenterUmmJson
     @collection[:temporal] = temporal(collection['TemporalExtents'])
     @collection[:science_keywords] = science_keywords(collection['ScienceKeywords']) if collection['ScienceKeywords']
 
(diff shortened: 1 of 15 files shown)


app/presenters/collection_details_presenter_umm_json.rb
```diff
@@ -48,10 +48,16 @@ def initialize(collection, collection_id = nil, token = nil, env = 'prod')
   def doi(doi)
     if doi && doi['DOI']
       doi = doi['DOI']
+      doi = doi.gsub('https://', '')
+      doi = doi.gsub('http://', '')
```
mightynimble: - combine two lines to something like ```doi.gsub(/https?:\/\//, '')```
 - Add a test using the collection in the ticket.
Everything else looks good.

mreese84:
Same thing here. Please add a test showing that these new values are being parsed out are actually being parsed out.

jboukes:
@mreese84 & @mightynimble - comments addressed, should be good so long as this turns green.  :)

PR #521, opened 2017-10-03
EDSC-1634: Updated method of redirect to Reverb

jboukes:


mreese84:
@mightynimble I'm merging this. In the morning, can you create a release for 1.54.13? That gets this code into what is going into UAT tomorrow for normal PM. We'll worry about back porting to 1.53.x once we confirm everything works in PT/UAT.

mightynimble:
@mreese84 I am a bit confused - isn't it true that 1.54.13 will be automatically created and deployed to sit?

mreese84:
No, 1.55.0 will be. You could just rename the release and put that into sit if you want

PR #522, opened 2017-10-03
EDSC-1640: Added default 'image unavailable' to use in case browse scaler is down

jboukes:
Edit: This initial write up is semi void!  It only describes part of the issue.

First - it is Firefox only, so load that up.  Then, to see the bug, visit Prod and search for "MODIS/Aqua Aerosol 5-Min L2 Swath 10km V5.1".  [Here is a direct link.](https://search.earthdata.nasa.gov/search?q=MODIS%2FAqua+Aerosol+5-Min+L2+Swath+10km+V5.1&ok=MODIS%2FAqua+Aerosol+5-Min+L2+Swath+10km+V5.1).  In the collections pane, after the two items found, you should see the bug - a broken list of semi-random seeming words.  Again - you'll only see this bug in Firefox.

The problem is this - the thumbnail being called from CMR is returning a 400 'bad request'.  It also returns a stub image that says 'No image available'.  Chrome sees this and says 'well, too bad about the 400 - I'll just show this stub image CMR was so kind to provide, instead'.  Sadly, Firefox does not behave the same way.  Firefox sees this and says "Ahhh!  A bad request!  Ignore the stub image and show the 'alt' field instead!  Ahhh!".

However, the alt field, when used, is so long that it sticks out into the text of the collection pane and also cascades down the form.  It looks terrible.  Firefox is being bad, here.

The 'best' fix would be to figure out how to get Firefox to show the stub 'No image available' thumbnail despite the 400; however, Firefox is very stubborn and I haven't figured out how to do that.  

A heavy-handed fix would be to delete the 'alt' value in the image entirely; then Firefox would just show a black square.  That's a bit extreme though.

The middle road fix is to 'escape' the alt value and also truncate it, so that what is shown in the square image, stays in the square image, and doesn't cascade down the page in a mess.  This is what this PR does.  Given the minor nature of this bug and its rarity, I would argue this suffices.


mreese84:
@jboukes thanks for the write up. I've been chewing on this and now that you've laid out what is causing the problem (well, one of them...) I have a clearer picture of how I'd like to handle it.

Instead of truncating the alt text, I'd rather just use a different field. How about instead of using `dataset_id` for the alt text, which will typically always be really long, we use the collection short name, which will typically be something really... well, short. That way we get a full value that should still convey meaning. There's something about just chopping off the text (which leaves you with something like "MODIS/Terr") that doesn't seem right. Using the full short name should get us around all of that.

Thoughts?

And finally... don't forget that we also need to decouple EDSC's dependency on Browse Scaler being available in this ticket as well.

jboukes:
@mreese84 - new fix is in!  This is better than the other fix as it covers both the Firefox issue as well as a general 'browse scaler is down' solution.  

I moved the 'image unavailable' svg file to our side of the fence to use in the case that browse scaler is non-responsive - prior the app was dependent upon browse scaler to provide that image (and still does to an extent) however that dependency means even the 'image unavailable' svg is gone if browse scaler goes down.

The 'onerror' attribute automatically executes setting the image src to the local image unavailable svg if it errors out.  Seems to work!

mreese84:
Thanks @jboukes. Please make sure to add a test that proves that the no image available image will still display in the case of 1) browse scaler being down and 2) the requested image not being available by browse scaler. Remember, with new functionality or bug fixes we will always want tests to prove the functionality works or the bug has been squashed.

jboukes:
@mreese84 I am uncertain of how to create such a test - I've been trying all morning.  The issue is, how to disable browse scaler for a test?

I've tried using one of the collections that would fail in Firefox; however sadly it does not fail from webkit.  It doesn't have the same error as Firefox and so the 'onerror' isn't being triggered.

@mightynimble - would you have any suggestions on how to disable browse scaler via rspec in order to test this fix?  :\

jboukes:
@mreese84 - just FYI, we do have a test already in place that checks for the 'image unavailable' svg in the case that the thumbnail doesn't exist, so we're good there - the one thing that I'm not sure how to test is the case that browse scaler is completely down.  

Here's the test we do have - for checking for image unavailable (collection_results_spec.rb):

it "displays  for collections which have no thumbnail URLs" do
    fill_in "keywords", with: 'C179003030-ORNL_DAAC'
    wait_for_xhr
    expect(find('img.panel-list-thumbnail')['src']).to have_content("image-unavailable.svg")
end

mightynimble:
This looks promising - https://robots.thoughtbot.com/how-to-stub-external-services-in-tests#disable-all-remote-connections

jboukes:
@mightynimble @mreese84  I am having a lot of trouble getting webmock to work and this is probably going to take a long time.  :(

jboukes:
@mreese84 @mightynimble This is my attempt to use webmock - it does not work and I've no idea why it doesn't work.  I've tried oodles of permutations as well.  :\

```
it "displays 'image unavailable' when browse scaler is down" do
    stub_request(:any, "cmr.earthdata.nasa.gov/browse-scaler/").to_return(status: [500, "Internal Server Error"])
    fill_in "keywords", with: 'C138500-PODAAC'
    wait_for_xhr
    expect(find('img.panel-list-thumbnail')['src']).to have_content("/images/image-unavailable.svg")
end
```

mightynimble:
If it is too much effort to use the gem, I don't think it is worth it to add the test. Maybe a manual test (with steps in jira ticket) wll be sufficient?

PR #524, opened 2017-10-10
EDSC-1645: Stripped out occurrences of dx.doi.org stuff in order to make string consistent

jboukes:
The regex based conditionals that originally controlled this string cleansing became unwieldy, and so this patch provides a more direct and simpler solution.  Based upon these values (provided on the ticket):

blank 1 record
doi: 645 records
http://dx.doi.org 122 records
https://doi.org 1 record
10. 2368 records ('10.' is part of the URL and so is 'okay')

The code has been updated to 'clean' the string of these substrings (i.e., 'doi:', 'http://dx.doi.org', 'https://doi.org', ~~and '10.'~~), and then concatenate the remainder with a consistent 'https://dx.doi.org'. 

This has been tested with  'C1200237617-MMT_1' as well as 'C1200230663-MMT_1' (both on SIT).

### DIFF
--- a/app/presenters/collection_details_presenter_umm_json.rb
+++ b/app/presenters/collection_details_presenter_umm_json.rb
@@ -49,18 +49,14 @@ class CollectionDetailsPresenterUmmJson < DetailsPresenterUmmJson
       doi = doi['DOI']
       # EDSC-1645: This string varies a *lot* - clean up permutations so that we start from the same place...
       doi = doi.gsub(/^https?\:\/\//, '')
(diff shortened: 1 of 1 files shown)


mreese84:
Are you stripping out "10."? If so, that is part of the valid DOI string and should be left in tact

jboukes:
@mreese84 - nope!  If you look at the summary at the top I mention that - I did strip it out for one commit and then backed out of it when I realized it needs to be there.

PR #525, opened 2017-10-12
EDSC-1636 EDSC-1615 backend and js support for project page. Show pro…

mightynimble:
…ject title on project page. Refactor, fix tests and add new tests.


mreese84:
Just discussed some things that need improving/addressing with Bo.

- [ ] Moving save project icon next to the name
- [ ] Change button colors to match approved comp in EDSC-1433
- [ ] Always surface a "Back to Search/Collections" link instead of the back to projects listing link
- [ ] Support existing project URLs and shortened path URLs

mreese84:
@mightynimble, I think I put my comment into Slack a little late. I see that you have hidden the "my project" button unless you are in the project context. Since the button is in the "global toolbar" we should show it at all times (if in the project context). So we should see it on the main search UI, when in the granules view, and when in the collection details view.

mreese84:
Functionality is spot on. Everything is good from my perspective except for the failing tests. Get this green (Monday) and we're ready to roll.

mightynimble:
Fixing them now. Should be green in my next push.

PR #527, opened 2017-10-13
EDSC-1652: Direct Download GA Event

jsiarto:
I'd like someone to review this to see if I'm on the right track. I added a click event to the download button and had that trigger a `dataaccessevent` with a new Download method type called "direct_download." This should still populate the data access metric, but now we'll have a breakdown under download for this type.

I'm about 90% sure I've used the right methods from the other metrics calls, but I'd like someone with more experience on the GA side to take a look.


mightynimble:
@jsiarto there are a couple of examples in ```app/assets/javascripts/util/metrics.js.coffee``` that you can take a look.

PR #528, opened 2017-10-17
EDSC-1648: Fixed spacing error with script generation; added 'how-to' information

jboukes:


app/views/granules/prepare_script.html.erb
```diff
@@ -126,7 +131,7 @@ prompt_credentials
 }
 
 fetch_urls <<'EDSCEOF'
-  </pre>
```
jboukes: The two space indentation here is what was causing the first link in the list in the script to also be indented - we don't want that.  The problem is 'pre' keeps things literal, so it also adds the spaces to the script it creates.

app/views/granules/prepare_script.html.erb
```diff
@@ -191,9 +196,9 @@ fetch_urls <<'EDSCEOF'
       var links = doc.links;
       preContent = document.getElementById('script');
       preContent.innerHTML += links.join("\n");
-      preContent.innerHTML += '\n'
       var next = getNextPageUrl();
       if (next) {
+        preContent.innerHTML += '\n'
```
jboukes: Now it only adds the carriage return if there *is* a 'next' - if there isn't, then it doesn't add a line.  This stops it from adding that extra carriage return at the end of the script file.

spec/features/collections/collection_results_spec.rb
```diff
@@ -103,7 +103,7 @@
   it "doesn't' show version_id for collections that don't have one" do
     fill_in "keywords", with: 'C1214605943-SCIOPS'
     wait_for_xhr
-    expect(page).to have_no_content("vNot provided")
+    expect(page).to have_no_content("Not provided")
```
jboukes: This change was a random typo I found and decided to clean up as part of the ticket.

app/views/granules/prepare_script.html.erb
```diff
@@ -43,7 +43,12 @@
 <body>
 <div id="errors"></div>
 <div id="loading">Preparing download script. Fetching downloadable granule links...</div>
-<div id="download-description" style="display: none;">Script preparation completed. Please click the button to download the script. <a href="#" id="download-button">Download Script File</a></div>
+<div id="download-description" style="display: none;">
+  <h2>How to use this script</h2>
+  <p>Linux: You must first make the script an executable by running the line 'chmod 777 download.sh' from the command line.  After that is complete, the file can be executed by typing './download.sh'.</p>
```
mreese84: Let's add a line to the effect of "For a detailed walk through of this process, read this How To guide." and make that link (and open in a new window/tab) to this wiki page - https://wiki.earthdata.nasa.gov/display/EDSC/How+To%3A+Use+the+Download+Access+Script

app/views/granules/prepare_script.html.erb
```diff
@@ -43,7 +43,10 @@
 <body>
 <div id="errors"></div>
 <div id="loading">Preparing download script. Fetching downloadable granule links...</div>
-<div id="download-description" style="display: none;">Script preparation completed. Please click the button to download the script. <a href="#" id="download-button">Download Script File</a></div>
+<div id="download-description" style="display: none;">
+  <p>Script preparation completed. Please click the button to download the script. <a href="#" id="download-button">Download Script File</a></p>
+  <p>For a detailed walk through of this process, please reference this <a target="_blank" href='https://wiki.earthdata.nasa.gov/display/EDSC/How+To%3A+Use+the+Download+Access+Script'>How To guide</a>.</p>
```
mreese84: @jboukes, I don't think I was very clear in my first comment. I thought that this line of text should be added in addition to the text you had before. I would add this after the Linux instructions, before the Windows instructions.

jboukes: Woops!  Yes, that's definitely different - I'll update the PR in 5, 4, 3...

app/views/granules/prepare_script.html.erb
```diff
@@ -43,7 +43,13 @@
 <body>
 <div id="errors"></div>
 <div id="loading">Preparing download script. Fetching downloadable granule links...</div>
-<div id="download-description" style="display: none;">Script preparation completed. Please click the button to download the script. <a href="#" id="download-button">Download Script File</a></div>
+<div id="download-description" style="display: none;">
+  <h2>How to use this script</h2>
+  <p>Linux: You must first make the script an executable by running the line 'chmod 777 download.sh' from the command line.  After that is complete, the file can be executed by typing './download.sh'.</p>
+  <p>Windows: The file can be executed within Windows by first installing a Unix-like command line utility such as <a href='https://www.cygwin.com/'>Cygwin</a>. After installing Cygwin (or a similar utility), run the line 'chmod 777 download.sh' from the utility's command line, and then execute by typing './download.sh'.</p>
+  <p>For a detailed walk through of this process, please reference this <a target="_blank" href='https://wiki.earthdata.nasa.gov/display/EDSC/How+To%3A+Use+the+Download+Access+Script'>How To guide</a>.</p>
```
mreese84: Sorry to nitpick... but I'd prefer this paragraph to be right after the Linux instructions. As it is, it could be read that the detailed instructions are for how to accomplish this using Windows... which it's not.

jboukes: @mreese84 done - happy to keep fiddling with it if it needs more tweaking!  Just let me know.

PR #532, opened 2017-10-23
EDSC-1641 Download all #data links in a zip.

mightynimble:
The example collection in the ticket times out when retrieving its links (firewall issue??). Plus, you may not even be able to see the collection due to ACL setups.

As an alternative, ```C1200237736-MMT_1``` was created in SIT with one granule that has two #data links.

Note that the feature doesn't have a test since we don't have any tests that verify the downloaded content in the past (capybara-webkit won't allow us to. See here: https://forum.shakacode.com/t/how-to-test-file-downloads-with-capybara/347). The existing test that verifies the download link, however, has been updated.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -92,6 +92,7 @@ gem 'daemons'
 
 gem 'nokogiri'
 gem 'responders', '~> 2.0'
+gem 'rubyzip', '>= 1.0.0'
 
 # Eventually we'll need these, but there's version conflict when installing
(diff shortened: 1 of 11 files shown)


mreese84:
I tried this with the first granule listed under `C194001241-LPDAAC_ECS` and got an Action Controller Exception (401 Unauthorized). When I hit the link listed in the metadata it works just fine. Any idea what's going on there?

PR #539, opened 2017-10-25
EDSC-1641 Introduce a dropdown if a granule has more than one downloadable links.

mightynimble:
Keep the 'click-to-download' feature if there is only one link.

### DIFF
--- a/app/views/search/_granule_list.html.erb
+++ b/app/views/search/_granule_list.html.erb
@@ -100,9 +100,19 @@ Delimiters: Separate multiple granule IDs by space, comma, or new line."/>
                 <a href="#" class="button icon-button master-overlay-forward" data-master-overlay-forward="Granules" data-bind="click: $parent.showGranuleDetails" title="View granule details">
                   <i class="fa fa-info-circle"></i>
                 </a>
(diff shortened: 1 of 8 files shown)


app/views/search/_granule_list.html.erb
```diff
@@ -100,9 +100,19 @@ Delimiters: Separate multiple granule IDs by space, comma, or new line."/>
...
-                <a class="button icon-button direct-download" title="Download single granule data" target="_blank" data-bind="visible: !$parent.collection.cwic() && online_access_flag, attr: {href: download_now_url()}">
-                  <i class="fa fa-download"></i>
-                </a>
+                <!--ko if: dataLinks().length == 1-->
+                  <a class="button icon-button direct-download" title="Download single granule data" target="_blank" data-bind="visible: !$parent.collection.cwic() && online_access_flag, attr: {href: dataLinks()[0].href}">
+                    <i class="fa fa-download"></i>
+                  </a>
+                <!--/ko-->
+                <!--ko if: dataLinks().length > 1-->
+                  <a href="#" data-toggle="dropdown" class="button icon-button direct-download" aria-expanded="false"><i class="fa fa-download"></i></a>
```
mreese84: I'm pretty sure our GA event is looking for a click on `a.direct-download`. That would fire a download event in GA if the dropdown is clicked. What we really want is for the event to be fired when one of the nested child `a` links has been clicked.

PR #543, opened 2017-10-28
EDSC-1613: Created minimap for new project page

jboukes:


jboukes:
@mreese84 - this should be good to go!

app/assets/javascripts/modules/map/map.js.coffee
```diff
@@ -72,43 +73,77 @@ ns.Map = do (window,
...
       @setProjection(projection)
       @setBaseMap("Blue Marble")
       @setOverlays([OVERLAYS[0], OVERLAYS[2]])
 
-      @time = ko.computed(@_computeTime, this)
-
-      map.fire('edsc.visiblecollectionschange', collections: page.project.visibleCollections())
-      @_granuleVisualizationSubscription = page.project.visibleCollections.subscribe (collections) ->
-        map.fire('edsc.visiblecollectionschange', collections: collections)
+      if !@isMinimap
```
abbottry: A couple of well placed comments could go a long way here. I *think* its just centering the map around the selection spatial param, but that's given little context.

app/assets/javascripts/models/page/project_page.js.coffee
```diff
@@ -45,6 +49,7 @@ ns.ProjectPage = do (ko,
       @ui =
         temporal: new TemporalModel(@query)
         projectList: projectList
+        projectPage: this
```
abbottry: This seems redundant -- is the reason for this the code below `$root.ui.projectPage.hasType()`?  If so, can that just be `$root.hasType()`?

app/assets/javascripts/modules/map/map.js.coffee
jboukes: I agree with you - though I'm not quite sure of the best comment as this isn't 'new' code (it's only 'new' because it was moved into a conditional).  This code was written quite some time ago.

app/assets/javascripts/models/page/project_page.js.coffee
jboukes: You have it right - setting 'projectPage' to equal itself exposes it to the view - without that line, the call to $root.ui.projectPage.hasType will fail. 

You're also right in that $root.hasType() does work!  However, isn't using the full path more descriptive?  I'm a little torn - could go either way.

jboukes: Ahhhhh - I see what you mean.  If you use $root.hasType, then the assignment to @ui isn't necessary.

I'll go with that - nice catch.  :)

app/assets/javascripts/models/page/project_page.js.coffee
```diff
@@ -56,6 +60,43 @@ ns.ProjectPage = do (ko,
...
+      type = ""
+      if @query.serialize().bounding_box then type = "Rectangle"
+      if @query.serialize().polygon then type = "Polygon"
+      if @query.serialize().point then type = "Point"
+      type
+  
+    hasType: =>
+      @query.serialize().bounding_box || @query.serialize().polygon || @query.serialize().point
+      
+    showTemporal: =>
```
mreese84: @jboukes can you remove all of this temporal code as well. This code should be introduced with the appropriate label and stuff as well. And secondary to that, this seems like a lot of code that is deconstructing and reconstructing date formats. Surely there's a better way to reformat a date and show only the bits we want? Is there something else going on here that I'm not seeing? Either way... we should take this conversation over to the appropriate branch/ticket.

jboukes: @mreese84 Back when that code was written, we'd talked about just commenting it out until we get to the temporal ticket (which, uncalled code kind of is 'commented out' since it's not executed).

However - not a prob.  :)  Will update in 5, 4, 3, 2,...

jboukes: Done!

As far as the other question, the code is a bit customized to the date format in order to get it closer to what was shown in the mock ups.  The way it is written is this:

If the year is the same between the dates, then show it as this:
Jan 12 - Mar 5, 2017

If the year *and* the month is the same between the dates, then show it as this:
Jan 12-15, 2017

Otherwise, show the dates in the more standard:
Jan 12, 2017 - Jan 12, 2018

We can talk more about that when we get to that ticket.

app/assets/javascripts/modules/map/map.js.coffee
abbottry: Understandable. I'm seeing the addition of `urlUtil`at the top of the document and then it's used within the code below to. Was this copied from somewhere else that already included that tool?

jboukes: It was - though from where precisely, I do not recall.  If you look at the changes for project_page.js.coffee, you'll see there it uses the @query class to access the URL parameters, which is preferable to using urlUtil.  

However, the map class does not seem to have visibility with the @query object; I attempted to introduce it, and it always comes up empty (i.e., calls to @query.serialize returns an empty structure).  I am not sure of the reason why.  It didn't seem like a huge deal, so I went with urlUtil in this case.  If you might have insight on what the @query class might need in order for it to self-populate (as it is in project_page.js.coffee), I would be happy to work on it further.

Also, if you think the urlUtil stuff could use additional comments to explain things, I'd be happy to do that as well - just let me know ;)

app/assets/javascripts/models/page/project_page.js.coffee
```diff
@@ -56,6 +60,16 @@ ns.ProjectPage = do (ko,
       setTimeout((=>
         @_loadFromUrl()
         $(window).on 'edsc.pagechange', @_loadFromUrl), 0)
+   
+    showType: =>
+      type = ""
```
abbottry: This is an odd method. I'd suggest setting a variable equal to `@query.serialize()` and using a standard if/else if/else statement.

app/assets/javascripts/modules/map/map.js.coffee
abbottry: Well, that's not a super reassuring response 😬 . Generally I'd ask someone before pushing something that I personally deem 'not the preferred way'. I will admit though, I don't know what the best practices are quite yet so until I cant say much. You are your own quality measure :).

jboukes: Please allow me to attempt more reassurance, then.   :blush:

I think urlUtil is fine, here.  If you do a text search through our files for 'urlUtil', you really will see it used throughout our code.  It's a dependable, structured way to grab parameters from the URL, which is exactly what was needed.

When I first implemented it, I also felt it was fine - however, Bo pointed out early in the PR that it would be preferable (at least to himself) to use the query object.  From that feedback, I updated project_page.js.coffee to use the query object (it had previously also been using urlUtil).  However, the map class doesn't seem to interact with the query object very well, so I've let it using urlUtil.

Bo sometimes was a little hard to read; however, the feeling I got from him was that the use of urlUtil wasn't so bad for him, just more of a preference.  Personally, again - I think it's fine.  I'm absolutely confident in the most important part which is - the code will work and do what it's supposed to.  :)

app/assets/javascripts/models/page/project_page.js.coffee
jboukes: How about a compromise and I just split it up in a more eye-friendly conditional.  :)

The previous code was due to a mild aversion to having multiple returns in a function - however, I agree that it 'looked funny'.  I did keep @query.serialize (seems unnecessary to tuck into into a variable when it's only used a few times), however, the conditional structure should be easier on the eyes.

PR #547, opened 2017-11-18
EDSC-1613: Updated minimap implementation

jboukes:
This update integrates the minimap in a manner that is more inline with the overall architecture.  The 'urlUtil' approach previously taken is no longer necessary; further, the issues found in test - where the spatial coordinates would not load from the saved project - are now working.

This is done by using the spatial selection object to draw the map instead of relying on urlUtil.


jboukes:
@abbottry - I agree with that, a subclass sounds good.  I just spoke with Mark and we will be doing that in a separate PR for logistical reasons.

PR #549, opened 2017-12-01
EDSC-1614: Adding Temporal Label

jboukes:
Take a look and let me know what you all think - this label uses moment.js as well as some simple conditional logic to make a custom temporal label.


app/assets/javascripts/models/page/project_page.js.coffee
```diff
@@ -88,6 +88,45 @@ ns.ProjectPage = do (ko,
     hasType: =>
       @query.serialize().bounding_box || @query.serialize().polygon || @query.serialize().point
 
+    showTemporal: =>
```
abbottry: I would much rather see this use `momentjs` or something similar to format these dates. This is a lot of code to understand and maintain for such a small piece of functionality.

jennywagg: Since @jboukes found that `momentjs` on its own doesn't do anything for ranges, would something like `moment-range` be helpful? https://www.npmjs.com/package/moment-range

jboukes: I can take a look at moment-range - as Jenny said, moment.js doesn't really do this 'special' formatting so much (it might make a single line here or there 'look' better).

Nice find, Jenny :)

jboukes:
My own comment - tests!  It needs tests!  I'll see about having tests added to this PR by Monday morning.

app/assets/javascripts/models/page/project_page.js.coffee
jboukes: Okay I do think moment.js actually did make this a good bit better so far as simplicity.  Creating the 'range' between the two dates using moment is actually pretty easy.  Take a look - should be better!

spec/features/projects/viewing_project_spec.rb
```diff
@@ -54,8 +54,9 @@
         click_link "Temporal"
         fill_in "Start", with: "2010-01-01 00:00:00\t\t"
         fill_in "End", with: "2014-02-01 00:00:00\t\t"
-        js_click_apply ".temporal-dropdown"
         wait_for_xhr
+        click_button('Apply Filter')
+        sleep 1
```
jboukes: I know 'sleep' isn't a good option - however, I'm just trying to find *something* that works here.  'wait_for_xhr' is throwing an odd error.

abbottry: Yea, this isn't something we should consider a solution though. This test will fail just as 'reliably' locally when run. This particular test failure is an actual bug, there are things going on in the background of this page (more specifically viewing the project page for a collection that the user doesn't have access to view) which prevents the page from loading. At least this was my initial thought -- so, 

1. Let's try to get tests to pass reliably locally instead of pushing changes to GitHub in hopes that they pass, if we can't trust the tests then they are useless.

2. We should always run tests locally before pushing, a PR isn't a holding area for our code, we should use PRs as a place to review completed code.

fixtures/cassettes/collections_responses.yml
```diff
@@ -80656,6 +80656,32 @@ a31c03cb72c4be7f2e5973194936bdf6369055ee:
...
+      - nginx
+      strict-transport-security:
+      - max-age=31536000; includeSubDomains;
+      content-length:
+      - '47'
+      connection:
+      - Close
+    body:
+      encoding: UTF-8
+      string: '{"errors":["Token edsc-access does not exist"]}'
```
abbottry: This seems like a bad/undesirable recording. (There are two more below)

spec/features/projects/viewing_project_spec.rb
```diff
@@ -45,6 +45,61 @@
...
+      before :all do
+        Capybara.reset_sessions!
+        load_page :search, project: ['C14758250-LPDAAC_ECS', 'C1000000000-LANCEAMSR2']
+        login
+        click_link "Temporal"
+        fill_in "Start", with: "2010-01-01 00:00:00\t\t"
+        fill_in "End", with: "2014-02-01 00:00:00\t\t"
+        wait_for_xhr
+        click_button('Apply Filter')
+        sleep 1
```
abbottry: We should take care of the underlying issue here, we can't let `sleeps` make it into the baseline.

spec/features/projects/viewing_project_spec.rb
```diff
@@ -45,6 +45,61 @@
...
+      end
+    end
+
+    context 'when a start date has been selected without an end date' do
+      before :all do
+        Capybara.reset_sessions!
+        load_page :search, project: ['C14758250-LPDAAC_ECS', 'C1000000000-LANCEAMSR2']
+        login
+        click_link "Temporal"
+        fill_in "Start", with: "2010-01-01 00:00:00\t\t"
```
abbottry: Are the tab characters necessary?

app/assets/javascripts/models/page/project_page.js.coffee
```diff
@@ -88,6 +88,31 @@ ns.ProjectPage = do (ko,
...
+      if @query.serialize().temporal
+
+        m_names = new Array("Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec")
+        label = "" 
+        dates = @query.serialize().temporal.split(",")
+        
+        dateStart = moment(dates[0].split("T")[0].split("-")).subtract(1, 'months') # months start at '0', so we need to subtract one
```
abbottry: `momentjs` should prevent the need for splitting on `T` and dealing with all of this.

app/assets/javascripts/models/page/project_page.js.coffee
```diff
@@ -88,6 +88,31 @@ ns.ProjectPage = do (ko,
     hasType: =>
       @query.serialize().bounding_box || @query.serialize().polygon || @query.serialize().point
 
+    showTemporal: =>
+      if @query.serialize().temporal
+
+        m_names = new Array("Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec")
```
abbottry: Is this still necessary?

app/assets/javascripts/models/page/project_page.js.coffee
```diff
@@ -88,6 +88,31 @@ ns.ProjectPage = do (ko,
...
+          else
+            # format only showing the year a single time (e.g., Jan 1 - Feb 10, 2011)
+            label = dateStart.format("MMM DD") + " - " + dateEnd.format("MMM DD, YYYY")
+        else
+          # Otherwise, show the full date range (e.g., Jan 1, 2010 - Feb 10, 2011)
+          label = if dateStart.isValid() then dateStart.format("MMM DD, YYYY") else "Beginning of time "
+          label += if dateEnd.isValid() then " - " + dateEnd.format("MMM DD, YYYY") else " - End of time"
+        label
+      else
+        false
```
abbottry: in the `if` portion of this method the return value is a string, yet in the `else` portion here its a boolean. 

We should ensure that the responses of our methods are consistent types so that when we call them we know what to expect.

spec/features/projects/viewing_project_spec.rb
abbottry: This particular failure is also plaguing the other open PR, so I'd save investigation of that for last on the ticket.. something screwy going on :-/

fixtures/cassettes/collections_responses.yml
jboukes: I wonder if that's related to the wait_for_xhr bug?  Hrm.  I'll rerecord using e2e-services as a baseline, *after* the sleep/wait_for_xhr issue is fixed.

spec/features/projects/viewing_project_spec.rb
jboukes: Might not be - I can try it without however that particular code was a copy and paste from similar tests. Usage of tabs when setting temporal ranges is in multiple tests.

app/assets/javascripts/models/page/project_page.js.coffee
jboukes: Confirmed - nice catch.

jboukes: Actually, sadly, misconfirmed - without the splits it reads the dates incorrectly :(  It appears they are programmatically necessary.

spec/features/projects/viewing_project_spec.rb
jboukes: @abbottry - I found a way around this failure by doing an 'expect' on the address bar URL to ensure that the temporal range setting had 'taken'.  No more 'sleep'!   

Not sure if this solution helps the problem with your PR or not, but you might want to review to consider it.

mreese84:
I'm not entirely sold on "Beginning of time" and "End of time" as the labels in the case of no start/end times. I've reached out to Chris L to see if he has a preference.

jboukes:
@mreese84 - totally fair, it was my 'best guess' for text at the time.  Let me know what Chris says and I'll update accordingly.

mreese84:
@jboukes Chris would like the labels changed to "Any start time" and "Any end time".

app/assets/javascripts/models/page/project_page.js.coffee
```diff
@@ -88,6 +88,29 @@ ns.ProjectPage = do (ko,
     hasType: =>
       @query.serialize().bounding_box || @query.serialize().polygon || @query.serialize().point
 
+    showTemporal: =>
+      label = ""
+      if @query.serialize().temporal
+        dates = @query.serialize().temporal.split(",")
+        
+        dateStart = moment(dates[0].split("T")[0])
```
abbottry: What was determined here regarding splitting out the date on the 'T'? Moment is capable of parsing the date without stripping information. I'm afraid to do this type of thing because it's specific to the date format currently being used, but it doesn't have to be.

jboukes: @abbottry - I wasn't able to split it all out.  If that split 'T' doesn't happen, it has a warped date (months or years off).  The other splits were 'okay' to remove.

jboukes:
@mreese84 - 'Any start time' and 'Any end time' has been added ;)

app/assets/javascripts/models/page/project_page.js.coffee
abbottry: I'm not able to reproduce that http://jsfiddle.net/remus/rLjQx/ Seems to work fine..

jboukes: I invite you to try it within the application - or, I could try it and post screenshots of it not working.  :\

jboukes: @abbottry - gah, let me toy with it just a bit more.  At a minimum I'll  see if I can find an explanation.

abbottry: Working for me. Moment takes into account the Timezone, so maybe you were seeing the timezone adjustment that moment does on display but moment has a very very good parser and this is a standard date format so I'd expect it to handle it just fine. 

screen shot 2017-12-06 at 2 00 49 pm

PR #561, opened 2018-01-11
EDSC-1641: Updated download links to only show those that are not inherited

jboukes:
The issue is that inappropriate links were being included into the 'Download' button of granules with multiple downloads.  The links are provided as a bundled array which contains both the links to the data (which we want) as well as links to collection related data (which we do not want).  

This array does not differentiate between the two _except_ that the collection related data is all set to "inherited=true".  Therefore, we can leverage this in order to filter them out, and only show the links to the granule specific data.

### DIFF
--- a/app/assets/javascripts/models/data/granule.coffee
+++ b/app/assets/javascripts/models/data/granule.coffee
@@ -27,7 +27,7 @@
     download_now_urls: ->
       links = []
       if @links? && @links.length > 0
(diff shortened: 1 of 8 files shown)


mreese84:
@jboukes can you address the test failure here: https://travis-ci.org/nasa/earthdata-search/jobs/327833101#L2370

jboukes:
@mreese84 - working on that, which is why I hadn't added any reviewers yet ;)

PR #563, opened 2018-01-16
EDSC-1641: Bug fix, update to test and fixtures

jboukes:
The issue is that inappropriate links were being included into the 'Download' button of granules with multiple downloads. The links are provided as a bundled array which contains both the links to the data (which we want) as well as links to collection related data (which we do not want).

This array does not differentiate between the two except that the collection related data is all set to "inherited=true". Therefore, we can leverage this in order to filter them out, and only show the links to the granule specific data.

### DIFF
--- a/app/assets/javascripts/models/data/granule.coffee
+++ b/app/assets/javascripts/models/data/granule.coffee
@@ -27,7 +27,7 @@
     download_now_urls: ->
       links = []
       if @links? && @links.length > 0
-        links.push(link) for link in @links when link.rel.indexOf('/data#') != -1
+        links.push(link) for link in @links when link.rel.indexOf('/data#') != -1 && link.inherited != true
       @dataLinks(links)
 
     onThumbError: (granule) ->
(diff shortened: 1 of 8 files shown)


abbottry:
The diff in `fixtures/cassettes/services_requests.yml` really concerns me. Generally the amount of changes to recordings concerns me. It seems like, from reading the PR description, that there should only be a code change and a few tests, but there are quite a few new and modified requests. Services though, changes a bunch of digests -- that is 1. odd, and therefore scary and 2. is not going to merge into `e2e-services`.

Do you know the reason for the recording updates, is it expected?

jboukes:
@abbottry @mreese84 - the issue is there appears to be a lack of recordings for some tests and so they are doing new 'Faraday' calls when executed.  If I run the suite locally, there are fixture updates that have nothing to do with this code - this is true even if you just run the suite on master locally.

I do not know why this has happened - a simple explanation is that there were instances where fixture updates were not included in PR's for some past tickets, and this caught up with us.  

It seems the solution would be - update the fixtures - all of them - and get them up to date.  

Thoughts?

mreese84:
@jennywagg, I'd venture a guess it would pass if you connect to the VPN. Performing a search locally requires VPN. Give that a shot and report back

jboukes:
@mreese84 I believe Jenny double checked that (I asked her as well) and verified she was on the VPN.  However, I do tend to believe it is a localized bug and not an actual issue with the test.

PR #566, opened 2018-01-23
EDSC-1754 - added CWIC portal

littleidiot40:


config/portals.yml
```diff
@@ -50,6 +50,13 @@ portals: &portals
     scripts:
       - edsc-portal.ornl.min.js
 
+  cwic:
+      org: CWIC
+      title: Search
+      params:
+        tag-key:
+          - edsc.portal.cwic.collections
```
abbottry: @littleidiot40 This tag doesn't match the tag in the ticket, or a tag I'm familiar with -- are you sure this is the correct tag?

littleidiot40: As per the ERB today we decided to use a new rather than existing tag as Chris Lynnes doesn't want to see 4K collections in this portal. He agreed to start with the 20 MOSDAC (ISRO) collections which are represented in that tag in PROD right now

abbottry: Got it. (If the ticket doesn't say it, I can't ensure its correct)

jboukes:
The change does seem to match the ticket now - I think the big remaining question is why it has broken so many tests?  Not sure if those are Travis 'glitches' or not (they might be...).  :\

PR #568, opened 2018-01-23
EDSC-1724: Created New Mechanism for Download Data Links

jboukes:
A user has three options when doing a 'direct download' of links, set in three buttons on the data access page - 'View Download Links', 'Download Data Links File', and 'Download Access Script'.  The first and the third are working fine; however, this ticket addresses and issue with the 2nd - 'Download Data Links File', which is to start a download of a textfile with all of the URLs of the related granules within it.

The problem is that this text file had a limit to 2000 granules (again, the other two mechanisms are fine).  This was really by design, in this line of code:

```send_data fetch_links, filename: "#{collection_id}_data_urls.txt"```

This line calls the function 'fetch_links' and downloads it - the issue is, 'fetch_links' is a function to be _iterated over_, and not just called once.  Calling it just once will guarantee you only get the first 2000 granules.

Further, it appears that the API is structured so that we can only really get 2000 granule URLs at once - that's how the others operate.

This solution creates a new view for the Download Data Links File mechanism - it opens a new page rather similar to the 'Download Access Script'.  The user is greeted with a page that shows them the progress of the links as they are parsed and retrieved; this is one of the reasons this new view is necessary.  Downloading all of the links to give to the user in a text file takes time - if we did it with just one click at the front, then the user would have to wait a pretty long time - 30 seconds or more maybe - without doing anything.  That's not good.

This way, with a view, the user will see the links progress as they are downloaded.  Just as with the 'Download Access Script', once the links have all been retrieved, the user is then given a button to obtain a text file with all that data.

Here's a localized link for your own testing - this particular collection is the one grumped in the ticket, and has 50K+ granules which makes it good for testing (though it does take a moment).

http://localhost:3000/data/configure?cmr_env=prod&p=C1231627223-GES_DISC&m=23.625!-102.568359375!4!1!0!0%2C2&tl=1500931499!4!!&q=TRMM_3B42RT&ok=TRMM_3B42RT&back=%2Fsearch%2Fgranules

Hit that, then try the Download Data Links Button, and - let me know.

### CODE REVIEW

File: app/views/granules/download_links.html.erb [hunk 1]
```diff
@@ -0,0 +1,149 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
+  <title><%= site_page_title('Download Granule Links') %></title>
+  <link rel="shortcut icon" href="/favicon.ico?v=1.1">
+  <style>
```
mreese84: You need to separate the style of this page out to it's own SCSS file. We always try to avoid straight CSS in a `style` element.

mreese84:
@jboukes I buy your logic in your explanation. With that being said, have you considered doing something with the button? For instance, when the user clicks the button to download the text file, change the button text to "Generating Text File" with a spinner indicating that the app is working. Could save the user a click, which is nice.

app/views/granules/download_links.html.erb
jboukes: @mreese84 - Please take a look at the other erb files within the same directory.  There is app/views/granules/download.html.erb and app/views/granules/prepare_script.html.erb. If you look at those, then you'll see why I did it this way - inline CSS is actually the established method within these little download mechanism erb files.   

This solution was also picked as it is consistent - this is how the other download mechanisms are designed.  Both of those other files have the same inline CSS structure; actually, prepare_script.html.erb was used as a template to generate this new file.  With this mechanism the new download_links has consistent behavior with the other download methods.

This kind of answers your other question as well - having the button on the front page turn into a spinner while it is all working on the backend would be a new design strategy.  My issues with it are that - first, my gut tells me that a user is more likely to feel impatient with that strategy (clicking a button and see it turn disabled while it spins for 60 seconds is oddly frustrating - a page refresh would be less so, to me).  Second, it would be a new design approach differentiating from the established method of the erb files, and so would take some thought to do (days, at least).  So, to me, that would be little to no gain for a lot of work.

mreese84: Ok. I buy it as that's the established pattern. I don't necessarily like the pattern, but it is what it is and to establish/redesign would've been a larger ticket altogether. So... approved on my end. Please wait for one other approval before merging.

spec/features/data_access/data_download_spec.rb
```diff
@@ -272,6 +272,34 @@
...
+          within_last_window do
+            synchronize do
+              expect(page).to have_link('Download Links File')
+            end
+            click_link 'Download Links File'
+          end
+        end
+
+        it 'downloads a list of links' do
+          within_last_window do
```
abbottry: Is this link beyond the first page? I think it be valuable to ensure that the loading of more than the first page

jboukes: @abbottry I looked up the collection and it appears to only have 15 granules, so no - no second page (2000 granules per page).  Further the collection seems to be another one that is 'gone' and can't be found live - seems to be a recording, only (I hate that).

I will look around for a larger collection to use - one that's maybe just a bit over 2000 would be nice.

jboukes: @abbottry As stated in an email I'm having some trouble finding a collection that is > 2000 granules that won't throw errors, since rspec does not seem to be able to login.  :(

jboukes: @abbottry test updated - I left the original test above, plus added one with a 2000+ granule collection.

app/views/granules/download_links.html.erb
```diff
@@ -0,0 +1,159 @@
+<!DOCTYPE html>
```
abbottry: Is this the newly created and used file or is this one no longer necessary now that the pages are combined?

spec/features/data_access/data_download_spec.rb
```diff
@@ -187,6 +189,34 @@
...
+    context 'upon clicking a "View/Download Data Links" button' do
+      before :all do
+        click_link "View/Download Data Links"
+        wait_for_xhr
+      end
+
+      it "displays a page containing direct download hyperlinks for the collection's granules in a new window" do
+        within_last_window do
+          expect(page).to have_text("Collection granule links have been retrieved")
+          expect(page).to have_link("ftp://n5eil01u.ecs.nsidc.org/DP5/AMSA/AE_SI6.003/2011.08.03/AMSR_E_L3_SeaIce6km_V15_20110803.hdf")
```
abbottry: @jboukes This test is likely to fail because of the previously merged PR that removes the FTP links. But also, its only the 65th link in the list so it doesn't actually test that paging is working and more than 1999 links appear. 

I would update this test and instead of looking for a specific link I would ensure in some way that there are in fact more than 1999 link present.

app/views/granules/download_links.html.erb
jboukes: @abbottry - this is the newly created and used file - another that is no longer necessary was deleted. ;)

spec/features/data_access/data_download_spec.rb
jboukes: @abbottry You're right - I checked and live it is correctly doing 'https' links - the only reason this test was working is that the links were 'ftp' in the fixtures and the test was matching.

I deleted the references to this collection in the fixtures, and updated the test to include a count of links (3385).  I reran them to regenerate the granule fixtures file (so that the links are correctly https) and just pushed it all up.

PR #569, opened 2018-01-24
EDSC-1550 Valid links in collection summary detail are clickable

jennywagg:


app/assets/javascripts/models/knockout_extensions.js.coffee
```diff
@@ -207,3 +207,19 @@ do (ko, $=jQuery) ->
...
     return
+
+  ko.bindingHandlers.descriptionLinks =
+    init: (element, valueAccessor, allBindings, viewModel, bindingContext) ->
+      description = $($.parseHTML(bindingContext.$data.description))
+      newDescription = ''
+      for el in description
+        if el.nodeName == 'A' && el.host != window.location.host
+          $(el).attr('target', '_blank')
+          icon = "<i class=\"fa fa-external-link\" aria-hidden=\"true\"></i>"
```
abbottry: Rather than including new markup here the EUI (even though we don't use the actual library, we still try and follow the best practices) prefers using a class on the `<a>` tag. 

See [Typography](https://cdn.earthdata.nasa.gov/eui/latest/docs/eui/index.html#type) from the EUI documentation.

You can see the css used in EUI [here](https://cdn.earthdata.nasa.gov/eui/latest/stylesheets/application.css)

app/assets/javascripts/models/knockout_extensions.js.coffee
```diff
@@ -207,3 +207,19 @@ do (ko, $=jQuery) ->
     ko.bindingHandlers.text.update element, ->
       truncatedText
     return
+
+  ko.bindingHandlers.descriptionLinks =
+    init: (element, valueAccessor, allBindings, viewModel, bindingContext) ->
+      description = $($.parseHTML(bindingContext.$data.description))
+      newDescription = ''
```
abbottry: Do we need to initialize this? It looks like we avoid returning a nil value given the if/else structure.

app/views/search/_collection_details.html.erb
```diff
@@ -133,11 +133,11 @@
         </div>
         <!-- end of first row -->
 
-        <!-- second row: metadata links | api endpionts | relative urls-->
+        <!-- second row: metadata links | api endpoints | relative urls-->
         <div class="row">
           <div class="flex-column">
             <!-- description -->
-            <div class="long-paragraph collapsed" data-bind="text: $data.description"></div>
+            <div class="long-paragraph collapsed" data-bind="descriptionLinks: $data.description "></div>
```
abbottry: Nice and clean -- I like this! :) Since there is another change here anyway, can we remove the extra space at the end of this data-binding?

app/assets/javascripts/models/knockout_extensions.js.coffee
jennywagg: Just realized you're probably referring to line 214 and not 212...

What's a better way that you recommend building this string from empty? There's probably some clean and simple way that doesn't involve initializing it like this, but I'm drawing a blank here

abbottry: Yea, `:214` -- well you are creating an empty string and then `+=` it. You could just set it in each of the if/else blocks instead of concatenating.

spec/features/collections/collection_details_spec.rb
```diff
@@ -202,6 +202,22 @@
...
   end
 
+  context 'when selecting a collection with a description with links' do
+    before :all do
+      load_page '/search/collection-details', focus: 'C1200230663-MMT_1', env: :sit, ac: true
+    end
+
+    it 'displays the external link' do
+      expect(page).to have_content('Greenland Ice Mapping Project (GIMP)')
+      # expect(page).to have_link("Greenland Ice Mapping Project (GIMP)", href: "http://nsidc.org/data/measures/gimp")
```
abbottry: Let's keep new PRs as clean as possible by removing debug or unused code.

spec/features/collections/collection_details_spec.rb
```diff
@@ -202,6 +202,22 @@
...
+      # find_link("Greenland Ice Mapping Project (GIMP)")[:target].should == 'blank'
+    end
+
+    it 'does not display a link for relative paths' do
```
abbottry: Remove empty test, or supply a check as described.

app/assets/javascripts/models/knockout_extensions.js.coffee
```diff
@@ -207,3 +207,18 @@ do (ko, $=jQuery) ->
...
+        if el.nodeName == 'A' && el.host != window.location.host
+          $(el).attr('target', '_blank')
+          $(el).attr('class', 'ext')
+          description[i] = el.outerHTML
+        else if el.nodeName == 'A' && el.host == window.location.host
+          description[i] = $(el.outerHTML).text()
+        else
+          description[i] = $(el).text()
+
+      ko.applyBindingsToNode(element, { 'html': (el for el in description).join('')}, bindingContext)
```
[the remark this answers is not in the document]
jennywagg: That would have been a good option, but `.join('')` doesn't work because `description` isn't really an Array - it's a DOM Nodelist so `.join()` isn't a valid function.

spec/features/collections/collection_details_spec.rb
jennywagg: I pushed this so that you guys could pull my branch and help debug my test in reference to my email to the team - in the future should I not push commits unless I know they work? Do I not need to push my commits in order for others to pull them locally?

spec/features/collections/collection_details_spec.rb
jennywagg: Same comment/questions as above - I noted this commit as "WIP" and pushed it so other devs could pull my changes and debug the failing tests. Can other devs pull my changes if I just commit and don't push? Rookie question

app/assets/javascripts/models/knockout_extensions.js.coffee
```diff
@@ -207,3 +207,18 @@ do (ko, $=jQuery) ->
...
+
+  ko.bindingHandlers.descriptionLinks =
+    init: (element, valueAccessor, allBindings, viewModel, bindingContext) ->
+      description = $($.parseHTML(bindingContext.$data.description))
+      for el, i in description
+        if el.nodeName == 'A' && el.host != window.location.host
+          $(el).attr('target', '_blank')
+          $(el).attr('class', 'ext')
+          description[i] = el.outerHTML
+        else if el.nodeName == 'A' && el.host == window.location.host
```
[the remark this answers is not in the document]
jennywagg: I followed the explanation provided in the second answer to this question:

https://stackoverflow.com/questions/2910946/test-if-links-are-external-with-jquery-javascript

jboukes:
Other than responding to Ryan's comments - looks fine to me.  👍

app/assets/javascripts/models/knockout_extensions.js.coffee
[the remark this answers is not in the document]
abbottry: 👍

app/assets/javascripts/models/knockout_extensions.js.coffee
```diff
@@ -207,3 +207,18 @@ do (ko, $=jQuery) ->
...
       truncatedText
     return
+
+  ko.bindingHandlers.descriptionLinks =
+    init: (element, valueAccessor, allBindings, viewModel, bindingContext) ->
+      description = $($.parseHTML(bindingContext.$data.description))
+      for el, i in description
+        if el.nodeName == 'A' && el.host != window.location.host
+          $(el).attr('target', '_blank')
+          $(el).attr('class', 'ext')
```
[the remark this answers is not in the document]
mreese84: I don't think it is likely that we will run across any links with classes in this case. I think `.addClass` is fine.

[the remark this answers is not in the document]
abbottry: Ok so @jennywagg can you make one last modification please -- change `$(el).attr('class', 'ext')` to `$(el).addClass('ext')` this will keep any existing class the user has specified remains intact.

PR #577, opened 2018-02-08
EDSC 1733: Created New Modal for Collections that Might Cause Delays

jboukes:
This new modal will appear based upon the following tagging system:

```
"edsc.collection_alerts": {
 "data": {
 "limit": 10,
 "message": "Optional message to be displayed in the modal"
 }
}
```

When the 'Download Data' button is clicked from the granules page of a collection, a check is made to see if this tag is set for the targeted collection, and if so it checks whether the number of granules is greater than the 'limit'.  If it is - it shows the modal warning the user of the possible delay as well as shows the optional message, if present.

Collection 'C24931-LAADS' on SIT has this flag set and is suitable for testing.

### DIFF
--- a/app/assets/javascripts/models/ui/project_list.js.coffee
+++ b/app/assets/javascripts/models/ui/project_list.js.coffee
@@ -95,6 +95,8 @@ ns.ProjectList = do (ko
       limit = false
       limit = collection.tags()['edsc.collection_alerts']['data']['limit'] if collection.tags() && collection.tags()['edsc.collection_alerts']
       if limit && collection.granuleCount() > limit
+        message = collection.tags()['edsc.collection_alerts']['data']['message']
+        if message then $("#delayOptionalMessage").text(message) else $("#delayOptionalMessage").text("")
         $("#delayWarningModal").modal('show')
       else
         @project.focus(collection)
--- a/app/assets/javascripts/models/ui/project_list.js.coffee
+++ b/app/assets/javascripts/models/ui/project_list.js.coffee
@@ -96,7 +96,7 @@ ns.ProjectList = do (ko
       limit = collection.tags()['edsc.collection_alerts']['data']['limit'] if collection.tags() && collection.tags()['edsc.collection_alerts']
       if limit && collection.granuleCount() > limit
         message = collection.tags()['edsc.collection_alerts']['data']['message']
-        if message then $("#delayOptionalMessage").text(message) else $("#delayOptionalMessage").text("")
+        if message then $("#delayOptionalMessage").text("Message from data provider: " + message) else $("#delayOptionalMessage").text("")
         $("#delayWarningModal").modal('show')
       else
         @project.focus(collection)
--- a/app/assets/javascripts/models/ui/project_list.js.coffee
+++ b/app/assets/javascripts/models/ui/project_list.js.coffee
@@ -85,22 +85,29 @@ ns.ProjectList = do (ko
         collections.splice(endIndex, 0, collection)
         @project.collections(collections)
 
+    _launchDownload: (collection) =>
+      @project.focus(collection)
+      @configureProject()
+
     awaitingStatus: =>
       @collectionsToDownload().length == 0 && @collectionOnly().length == 0 && @submittedOrders().length == 0 && @submittedServiceOrders().length == 0 && @collectionLinks().length == 0
 
     loginAndDownloadCollection: (collection) =>
       $('#delayOk').on 'click', =>
-        @project.focus(collection)
-        @configureProject()
+        @_launchDownload(collection)
       limit = false
-      limit = collection.tags()['edsc.collection_alerts']['data']['limit'] if collection.tags() && collection.tags()['edsc.collection_alerts']
+      if collection.tags()
+        if collection.tags()['edsc.collection_alerts']
+          limit = collection.tags()['edsc.collection_alerts']['data']['limit']
       if limit && collection.granuleCount() > limit
         message = collection.tags()['edsc.collection_alerts']['data']['message']
-        if message && message.length > 0 then $("#delayOptionalMessage").text("Message from data provider: " + message) else $("#delayOptionalMessage").text("")
+        if message && message.length > 0
+          $("#delayOptionalMessage").text("Message from data provider: " + message) 
+        else 
+          $("#delayOptionalMessage").text("")
         $("#delayWarningModal").modal('show')
       else
-        @project.focus(collection)
-        @configureProject()
+        @_launchDownload(collection)
 
     loginAndDownloadGranule: (collection, granule) =>
       @project.focus(collection)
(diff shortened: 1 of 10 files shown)


app/views/search/_delay_modal.html.erb
```diff
@@ -0,0 +1,24 @@
...
+    <div class="modal-content">
+      <div class="modal-header">
+        <a href="#" class="close modal-close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></a>
+        <h4 class="modal-title" id="delayWarningModalLabel">Possible Delay Ahead</h4>
+      </div>
+      <div class="modal-body">
+        <p>
+          This collection is known to cause delays and may take longer to download than expected.
+        </p>
+        <p>
```
mreese84: Obviously I've not seen this in context of the app/workflow... but I think we should style this optional message from the provider a bit differently or give the user some context that the message comes from the provider. Maybe a "Message from data provider" label/heading?

jboukes: @mreese84 I took a shot at it - added some leading text to make it more obvious where the optional message was coming from.

Let me know what you think ;)

app/views/search/_delay_modal.html.erb
```diff
@@ -0,0 +1,24 @@
...
+          <span id='delayOptionalMessage'></span>
+        </p>
+      </div>
+      <div class="modal-footer">
+        <a class="button text-button" id="delayOk" data-dismiss="modal" href="#">Okay</a>
```
jennywagg: I'm not sure if there's a UX precedent for this already in EDSC, but I think there may be a more effective way to word this button than "Okay" - I clicked on this button and my instinctual expectation for the "Okay"/"OK" button is that the modal would just disappear, as if I was just "dismissing" the modal or saying "okay." 

Of course, my expectation as a user should have been that this button brings me on to the download - a more "action specific" button might feel more natural (maybe "Continue" or "Download"). A counterargument I could understand is that any regular user may naturally understand "Okay" to continue them through the workflow (because what good would hitting "Okay" be otherwise?).

It's possible I'm alone in my initial perception of this UX flow, so please feel free to disagree with me since I can understand a solid argument for the "Okay" button!

jboukes: @jennywagg Seems  like a valid point to me!  'Okay' is also a bit informal.

Perhaps the button could say 'Proceed' - would that be more appropriate?  Open to other suggestions as well.

mreese84: "Continue" is better

app/views/search/_delay_modal.html.erb
```diff
@@ -0,0 +1,24 @@
...
+<div class="modal" id="delayWarningModal" tabindex="-1" role="dialog" aria-labelledby="delayWarningModal">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <a href="#" class="close modal-close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></a>
+        <h4 class="modal-title" id="delayWarningModalLabel">Possible Delay Ahead</h4>
+      </div>
+      <div class="modal-body">
+        <p>
+          This collection is known to cause delays and may take longer to download than expected.
```
mreese84: Let's change this text to...
"The amount of granules requested for download may result in a significant delay. You may wish to further refine your search to limit the size of your download. Are you sure you want to continue?"

app/views/search/_delay_modal.html.erb
```diff
@@ -0,0 +1,24 @@
...
+        </p>
+        <p>
+          <span id='delayOptionalMessage'></span>
```
mreese84: What happens if there is no optional message from the provider?

app/views/search/_delay_modal.html.erb
```diff
@@ -0,0 +1,24 @@
...
+        </p>
+      </div>
+      <div class="modal-footer">
+        <a class="button text-button" id="delayOk" data-dismiss="modal" href="#">Okay</a>
+        <a class="button text-button" id="delayCancel" data-dismiss="modal" href="#">Cancel</a>
```
mreese84: Given the updated text I just provided, let's change this cancel button to "Refine Search"

app/views/search/_delay_modal.html.erb
jboukes: @mreese84 - in that case, then span would be empty of all text and invisible to the user.  It only becomes populated with text in the event that there is an optional message to present.

app/views/search/_delay_modal.html.erb
jboukes: You got it :)

app/assets/javascripts/models/ui/project_list.js.coffee
```diff
@@ -89,8 +89,18 @@ ns.ProjectList = do (ko
...
+        @project.focus(collection)
+        @configureProject()
+      limit = false
+      limit = collection.tags()['edsc.collection_alerts']['data']['limit'] if collection.tags() && collection.tags()['edsc.collection_alerts']
+      if limit && collection.granuleCount() > limit
+        message = collection.tags()['edsc.collection_alerts']['data']['message']
+        if message && message.length > 0 then $("#delayOptionalMessage").text("Message from data provider: " + message) else $("#delayOptionalMessage").text("")
+        $("#delayWarningModal").modal('show')
+      else
+        @project.focus(collection)
```
abbottry: I think it'd be best to pull our the 'successful' functionality into a method instead of copying and pasting it, in the future if we add work to this process we'd need to duplicate it which can easily be overlooked.

app/assets/javascripts/models/ui/project_list.js.coffee
```diff
@@ -89,8 +89,18 @@ ns.ProjectList = do (ko
       @collectionsToDownload().length == 0 && @collectionOnly().length == 0 && @submittedOrders().length == 0 && @submittedServiceOrders().length == 0 && @collectionLinks().length == 0
 
     loginAndDownloadCollection: (collection) =>
-      @project.focus(collection)
-      @configureProject()
+      $('#delayOk').on 'click', =>
```
abbottry: Knockout has functionality that we've used previously for handling modals, I think? I know we have an extension it. I remember using it for the OPeNDAP stuff that I did, is there a reason that we can't use it here? (My use case didn't have multiple buttons so that may prove different)

jboukes: @abbottry - this is the only comment that I didn't make a change to address in the upcoming push (which will happen about 30 seconds after I post this comment).  

I did some searches through the code and I'm not seeing other instances to mimic; I'm not so sure there is a smoother method to attach this event handler.  I'm happy to adjust it if we can find a better way.

PR #578, opened 2018-02-14
EDSC-1722: Fixed tags bug and changed auto-select and submit button behavior

jboukes:
The first issue in EDSC-1722 involved collection 'C1424491172-LARC_ASDC' - within the data download page, if the user hits 'continue', the page does nothing and a console error is present if one looks.  

The issue is that the code checks "collection.tags()["edsc.limited_collections"]"; however, with this collection, tags() returns null and therefore has no index.  This automatically throws a JS exception; the fix is simply to check to make sure tags() returns a value before checking what the value of the index might be.

The code was also updated such that if there is only a single download mechanism available, it is auto selected.  Further, the 'submit' button will not be active until at least one item is chosen; it remains in an 'inactive' state until the user makes their choice.


jboukes:
@jennywagg - hover tip added :)

app/assets/javascripts/models/data/service_options.js.coffee
```diff
@@ -61,6 +61,7 @@ ns.ServiceOptions = do (ko, edsc = @edsc, KnockoutModel = @edsc.models.KnockoutM
           @subsetOptions(null)
         result
 
+
```
jennywagg: Just being picky, there's an extra whitespace that's been added here!

app/assets/javascripts/models/data/service_options.js.coffee
```diff
@@ -176,10 +176,14 @@ ns.ServiceOptions = do (ko, edsc = @edsc, KnockoutModel = @edsc.models.KnockoutM
...
       
       result = false
+      selected = false
       for m in @accessMethod()
+        if @granuleAccessOptions().methods?.length == 1
+          m.method(m.availableMethods[0].name)
         result = true if m.method()? && m.isValid()
         result = true if !m.loadForm()
-      result
+        selected = true if m.serialize().method != null
```
abbottry: Here you're using `m.serialize().method != null` but just above the previous code is using `m.method()?` -- do these two methods of checking for null differ?

app/assets/javascripts/models/data/service_options.js.coffee
```diff
@@ -176,10 +176,14 @@ ns.ServiceOptions = do (ko, edsc = @edsc, KnockoutModel = @edsc.models.KnockoutM
       return true if @granuleAccessOptions().methods?.length == 0
       
       result = false
+      selected = false
```
abbottry: - `result = true if m.method()? && m.isValid()`
- `selected = true if m.method()?`

and the value we return `result && selected`

My comment on redundancy was the above, I don't think you need the `selected` value at all. It's check is just a subset of the check that was already there.

app/assets/javascripts/models/data/service_options.js.coffee
```diff
@@ -176,14 +176,13 @@ ns.ServiceOptions = do (ko, edsc = @edsc, KnockoutModel = @edsc.models.KnockoutM
       return true if @granuleAccessOptions().methods?.length == 0
       
       result = false
-      selected = false
       for m in @accessMethod()
         if @granuleAccessOptions().methods?.length == 1
           m.method(m.availableMethods[0].name)
-        result = true if m.method()? && m.isValid()
+        result = true if m.isValid()
```
abbottry: This changes the logic for determining whether or not result is true. Previously it was `result = true if m.method()? && m.isValid()` which requires both of those conditions to be true. This change has modified this logic to be `OR` by setting them separately. Your update shouldnt require any change to this logic, you're just ensuring that if there is only one data access method, that its pre-checked for the user. 

Again, I don't think any of this is necessary. Line 180 and 181 are the only updated needed to accomplish your goal unless I'm completely missing something.

jboukes: @abbottry - There are two goals :)

Lines 180 and 181 do the first goal - which is preselect a method if there is only one method - the first goal.

The second goal is whether or not the 'submit' button is enabled.  Changing the logic to effectively being an 'or' is required for this to work - the submit button should not be enabled if a method is not selected.  Therefore, result = false if !m.method()? makes sure that it returns false - thus disabling the 'submit' button - in the case that the method has not been chosen.

PR #583, opened 2018-02-27
EDSC-1678: Display format of start/end dates has consistent placement on granule card

jennywagg:
https://bugs.earthdata.nasa.gov/browse/EDSC-1678

The collection used as an example for this bug is https://search.sit.earthdata.nasa.gov/search/granules?cmr_env=uat&p=C1000-LPDAAC_TS2&m=-57.375!24.46875!1!1!0!0%2C2&tl=1494414918!4!!&sb=-22.21875%2C9.5625%2C29.25%2C43.59375

By putting the "start" and "end" labels in table cells with fixed height, the spacing is consistent when the labels are not (in the case of no end label) and allows us to place the start label at the bottom of the card.


abbottry:
I think the design team is going to request that we use flexbox here instead of tables. The specific height/width properties will not render across all browsers the same. I've attached a screenshot of what it looks like on my machine and it is not desirable. 

@mreese84 or @jsiarto can help with specifics.

screen shot 2018-02-27 at 11 08 48 am

app/views/search/_granule_list.html.erb
```diff
@@ -84,15 +84,15 @@ Delimiters: Separate multiple granule IDs by space, comma, or new line."/>
...
-                  <!--/ko-->
-                  <!--ko if: ($data[1] && !$data[0])-->
-                  <div class="temporal-end"><h5>END</h5><p data-bind="text: $data[1]"></p></div>
-                  <!--/ko-->
-                  <!--ko if: ($data[0] && $data[1])-->
-                  <div class="temporal-start"><h5>START</h5><p data-bind="text: $data[0]"></p></div><div class="temporal-end"><h5>END</h5><p data-bind="text: $data[1]"></p></div>
-                  <!-- /ko -->
+                  <div class="temporal-end"><h5>END</h5>
+                    <!--ko if: $data[1]-->
+                    <p data-bind="text: $data[1]"></p>
```
abbottry: This can be much simpler just using `<p data-bind="text: $data[1] || Not Provided"></p>` or some variation using a method.

PR #584, opened 2018-02-27
EDSC-1599: Added additional logging during data retrieval

jboukes:
This additional logging will output the data request object when the 'Submit' button is clicked as per the following:

The following task is being submitted to delayed_job: {"query":"p=C179003620-ORNL_DAAC","collections":[{"id":"C179003620-ORNL_DAAC","params":"echo_collection_id=C179003620-ORNL_DAAC&sort_key%5B%5D=-start_date&page_size=20","serviceOptions":{"accessMethod":[{"method":"Order","type":"order","id":null}]},"form_hashes":[{"id":"order","form_hash":null}]}],"source":"p=C179003620-ORNL_DAAC&tl=1503944607!4!!&back=%2Fsearch%2Fgranules"}

It's worth noting that a test to check for this logging isn't available - I do not believe our test suites check logging (or are even able to).

### DIFF
--- a/app/controllers/data_access_controller.rb
+++ b/app/controllers/data_access_controller.rb
@@ -51,8 +51,8 @@ class DataAccessController < ApplicationController
     retrieval.project = project
     retrieval.save!
 
-    Rails.logger.info("The following task is being submitted to delayed_job: " + params[:project]) 
-    Retrieval.delay.process(retrieval.id, token, cmr_env, edsc_path(request.base_url + '/'), session[:access_token])
+    new_job = Retrieval.delay.process(retrieval.id, token, cmr_env, edsc_path(request.base_url + '/'), session[:access_token])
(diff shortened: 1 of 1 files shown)


andyfekete:
John -- can you also include the order id, if it is available?

jboukes:
@andyfekete - you got it, just found a way to do that :)

Will merge when green.

jboukes:
@andyfekete - it is the ID of the order within the delayed jobs queue.

I do not believe the ID you are looking for is available at this point in the work flow.  I believe (could be wrong?) that the hex coded string you're referring to isn't available until after the job has been submitted to CMR (in other words, the task is being actively worked by delayed_job).  

The logging here is occurring with the 'Submit' button being clicked (as per the ticket instructions) - which is earlier in the process.

PR #586, opened 2018-03-01
EDSC-1534: Implemented delayed_job queues and added additional workers

jboukes:
The issue this ticket addresses is that job retrieval can become hung up if the single delayed_job worker becomes 'enthralled' with a long standing task, such as those coming from NSIDC.  The current implementation of data retrieval uses only a single lowly worker and one queue.

Edit: An earlier version of this PR required a modification to config/application.yml - this is no longer necessary.

My largest concern is with the instruction:

jobs: bundle exec bin/delayed_job --pool=Default:2 --pool=NSIDC,LPDAAC:2 --pool=*:2 run

I *believe* this will work, however my ability to test this line locally is limited.  It works well enough using 'start' instead of 'run' - when I use 'run', it keeps the process in the foreground (which it is supposed to do) however doesn't seem to start all the workers.  I am very interested to see how this line will behave on SIT.

To explain that line - what that does, is it creates two workers for the 'Default' pool, two workers for the 'NSIDC,LPDAAC' pool, and then two extra workers who will do anything that comes up that needs doing in either queue.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,4 +1,4 @@
 web: ./start.sh
-jobs: bundle exec bin/delayed_job --pool=Default:2 --pool=NSIDC,LPDAAC:2 --pool=*:2 run
+jobs: bundle exec bin/delayed_job --pool=Default:2 --pool=NSIDC,LPDAAC:2 --pool=*:2 start
 cron: ./cron.sh
 on_build: bundle exec rake deploy:pre
--- a/Procfile
+++ b/Procfile
@@ -1,4 +1,5 @@
 web: ./start.sh
-jobs: bundle exec bin/delayed_job --pool=Default:2 --pool=NSIDC,LPDAAC:2 --pool=*:2 start
+jobs: bundle exec bin/delayed_job --pool=Default:2 --pool=NSIDC,LPDAAC:2 --pool=*:2 run
+worker: bundle exec rake jobs:work
 cron: ./cron.sh
 on_build: bundle exec rake deploy:pre
--- a/Procfile
+++ b/Procfile
@@ -1,6 +1,6 @@
 web: ./start.sh
 normal_jobs: bundle exec bin/delayed_job --pool=Default:2 run
-priority_jobs: bundle exec bin/delayed_job --pool=NSIDC,LPDAAC:2 run
+priority_jobs: bundle exec bin/delayed_job --pool=NSIDC --pool=LPDAAC run
 worker: bundle exec rake jobs:work
 cron: ./cron.sh
 on_build: bundle exec rake deploy:pre
--- a/Procfile
+++ b/Procfile
@@ -1,4 +1,4 @@
 web: ./start.sh
-jobs: bundle exec bin/delayed_job run
+jobs: bundle exec bin/delayed_job --pool=Default:2 --pool=NSIDC,LPDAAC:2 --pool=*:2 run
 cron: ./cron.sh
 on_build: bundle exec rake deploy:pre
--- a/Procfile
+++ b/Procfile
@@ -1,5 +1,6 @@
 web: ./start.sh
-jobs: bundle exec bin/delayed_job --pool=Default:2 --pool=NSIDC,LPDAAC:2 --pool=*:2 run
+normal_jobs: bundle exec bin/delayed_job --pool=Default:2 run
+priority_jobs: bundle exec bin/delayed_job --pool=NSIDC,LPDAAC:2 run
 worker: bundle exec rake jobs:work
 cron: ./cron.sh
 on_build: bundle exec rake deploy:pre
--- a/app/controllers/data_access_controller.rb
+++ b/app/controllers/data_access_controller.rb
@@ -66,7 +66,7 @@ class DataAccessController < ApplicationController
       end
     end
     new_job = Retrieval.delay(:queue => queue).process(retrieval.id, token, cmr_env, edsc_path(request.base_url + '/'), session[:access_token])
-    Rails.logger.info("Delayed job " + new_job.id.to_s + " has been sent into queue " + new_job.queue.to_s + " with:" + params[:project]) 
+    Rails.logger.info("Delayed Job " + new_job.id.to_s + " has been sent into queue " + new_job.queue.to_s + " with:" + params[:project]) 
     redirect_to edsc_path("/data/retrieve/#{retrieval.to_param}")
   end
 
--- a/app/controllers/data_access_controller.rb
+++ b/app/controllers/data_access_controller.rb
@@ -66,7 +66,15 @@ class DataAccessController < ApplicationController
       end
     end
     new_job = Retrieval.delay(:queue => queue).process(retrieval.id, token, cmr_env, edsc_path(request.base_url + '/'), session[:access_token])
-    Rails.logger.info("Delayed job " + new_job.id.to_s + " has been sent into queue " + new_job.queue.to_s + " with:" + params[:project]) 
+    if new_job
+      if new_job.queue
+        Rails.logger.info("Delayed job " + new_job.id.to_s + " has been sent into queue " + new_job.queue.to_s + " with:" + params[:project]) 
+      end
+      Rails.logger.info("There was an issue reading the queue from the newest job.")
+    else
+      Rails.logger.info("Retrieval returned an unexpected object when attempting to create a new job.")
+    end
+
     redirect_to edsc_path("/data/retrieve/#{retrieval.to_param}")
   end
 
--- a/app/controllers/data_access_controller.rb
+++ b/app/controllers/data_access_controller.rb
@@ -52,7 +52,7 @@ class DataAccessController < ApplicationController
     retrieval.save!
     
     queue = "Default"
-    daacs = ENV["priority_daacs"].split(",")
+    daacs = ENV["priority_daacs"] ? ENV["priority_daacs"].split(",") : []
     collections = JSON.parse(params[:project])['collections']
     daacs.each do |daac|
       collections.each do |collection|
--- a/app/controllers/data_access_controller.rb
+++ b/app/controllers/data_access_controller.rb
@@ -65,7 +65,8 @@ class DataAccessController < ApplicationController
         break
       end
     end
-    Retrieval.delay(:queue => queue).process(retrieval.id, token, cmr_env, edsc_path(request.base_url + '/'), session[:access_token])
+    new_job = Retrieval.delay(:queue => queue).process(retrieval.id, token, cmr_env, edsc_path(request.base_url + '/'), session[:access_token])
+    Rails.logger.info("Delayed job " + new_job.id.to_s + " has been sent into queue " + new_job.queue.to_s + " with:" + params[:project]) 
     redirect_to edsc_path("/data/retrieve/#{retrieval.to_param}")
   end
 
--- a/app/controllers/data_access_controller.rb
+++ b/app/controllers/data_access_controller.rb
@@ -66,15 +66,7 @@ class DataAccessController < ApplicationController
       end
     end
     new_job = Retrieval.delay(:queue => queue).process(retrieval.id, token, cmr_env, edsc_path(request.base_url + '/'), session[:access_token])
-    if new_job
-      if new_job.queue
-        Rails.logger.info("Delayed job " + new_job.id.to_s + " has been sent into queue " + new_job.queue.to_s + " with:" + params[:project]) 
-      end
-      Rails.logger.info("There was an issue reading the queue from the newest job.")
-    else
-      Rails.logger.info("Retrieval returned an unexpected object when attempting to create a new job.")
-    end
-
+    Rails.logger.info("Delayed job " + new_job.id.to_s + " has been sent into queue " + new_job.queue.to_s + " with:" + params[:project]) 
     redirect_to edsc_path("/data/retrieve/#{retrieval.to_param}")
   end
 
--- a/app/controllers/data_access_controller.rb
+++ b/app/controllers/data_access_controller.rb
@@ -52,7 +52,7 @@ class DataAccessController < ApplicationController
     retrieval.save!
     
     queue = "Default"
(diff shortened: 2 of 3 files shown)


Procfile
```diff
@@ -1,4 +1,4 @@
 web: ./start.sh
-jobs: bundle exec bin/delayed_job run
+jobs: bundle exec bin/delayed_job --pool=Default:2 --pool=NSIDC,LPDAAC:2 --pool=*:2 run
```
tgoff: I dont see any reason this wouldnt work (from an NGAP pserspective), but there are a few ways to do this which are worth considering:
- In the approach you have now, 6 delayed_job workers will be running in the same ECS task (AKA in its own docker container).  I do not know what the memory/cpu implications of adding a worker are, and we can scale up RAM/CPU if we need to, but the point is that the workers will be sharing resources
- You could also scale up the number of 'jobs' tasks running in NGAP, meaning that n separate docker containers will be running the same delayed_job command (which does not seem to be what you want since you seem to have different types of queues).
- Finally, you could take sort of a hybrid approach.  basically, create multiple jobs tasks in the Procfile with unique names (e.g. nsidc_jobs, lpdaac_jobs...) and have each of those run some subset of the workers.  This way, you have container separation, but each is doing something different.  Unless you are confident that there will be no resource contention in the current approach, I would suggest trying this.  

As always, try this out in searchlab or sit before merging to make sure everything works as expected.

jboukes: Totally agree with testing on SIT - it's definitely required here.

It is true that the additional workers will take up additional memory, however I can't fathom it's all that much.  As you said as well - memory is also pretty cheap and can be added.  From all the documentation and chatter, using multiple workers as well as queues with delayed_job is rather common.

mreese84:
@jboukes, my biggest question is can we prove that this will work before merging. So, I know that we can deploy this branch directly to SIT for testing. However, can we reliably generate the conditions by which an order(s) get stuck in the creating state? @andyfekete, can we do that? If so, then I'd like to do that on SIT (before deploying this branch). And record how long it took the queue to be worked off. And then deploy this branch and do the same exact thing. I'd suspect that after doing the same thing we should never see the queue get gummed up.

Does that make sense? Basically I'm asking for some sort of demonstration that this will work before merging.

Procfile
andyfekete: Tim -- I'm wary of the additional complexity, though I see the merit of your hybrid approach.  

How about we start simple -- i.e. all workers running in the same ESC task / docker container -- see how that works, and reach for the hybrid solution if we have to?

tgoff: That is fine with me, just throwing out the options :-) as far as the memory/CPU load, it may be hard to determine that in a non-prod env, so make sure you/ngap are keeping an eye on it when it goes to prod so we know if we need to bump something up.

app/controllers/data_access_controller.rb
```diff
@@ -50,10 +50,22 @@ def retrieve
     retrieval.user = user
     retrieval.project = project
     retrieval.save!
-
-    new_job = Retrieval.delay.process(retrieval.id, token, cmr_env, edsc_path(request.base_url + '/'), session[:access_token])
-    Rails.logger.info("A new delayed job with ID " + new_job.id.to_s + " has been created: " + params[:project]) 
-
+    
+    queue = "Default"
+    daacs = ENV["priority_daacs"] ? ENV["priority_daacs"].split(",") : []
```
abbottry: Good catch on this, I was going to comment on it. That being said I think a better way to handle this would be to make it a config in the application, its already hardcoded in the Procfile so its not going to provide us any flexibility by being an ENV var. Additionally we'll have to set this value in all of our local environments, on Travis, and on Bamboo, just a lot of overhead for something that isn't going to change often and isn't a secret key.

abbottry:
Testing it will be the challenge.. I have no real knowledge of the issue here but I would suspect that if @andyfekete where to look at logs on the worker node for EDSC, the thing that takes the longest is our  code that is being called when this is executed. We ping for all the collections, and then submit the order to CMR, that call could take a while but I'm not convinced its a memory hog. Testing in SIT seems like an unreliable test though I think generally having more workers is a good start.

app/controllers/data_access_controller.rb
jboukes: @abbottry - considering the names are already in the Procfile, I *could* just include them in this code as well - just set daacs = ['NSIDC', 'LPDAAC'].  Other daacs could be added down the line as needed.  

What do you think?

abbottry: Sure, thats fine -- if it gets more complicated in the future we can pull them out.

jboukes:
@mreese84 - here's how you can run it locally and at least see the jobs are being added to the correct queues.

First, modify your application.yml so that it includes the line:

priority_daacs: "NSIDC,LPDAAC"

Then, start your local instance up, find an NSIDC collection, and then submit a data retrieval.  You can them drop back to your console and start up a rails command line with 'rails c'.  From that command line, you can type:

Delayed::Job.last
or
Delayed::Job.all

So long as you don't have a running worker, the newly submitted job should be shown and it should also show that it is in the 'NSIDC' queue.  I've done this locally and witnessed the work being added to the correct queue.

To really test this though, it needs to be on SIT, I believe.  :\

app/controllers/data_access_controller.rb
```diff
@@ -50,10 +50,22 @@ def retrieve
...
-
-    new_job = Retrieval.delay.process(retrieval.id, token, cmr_env, edsc_path(request.base_url + '/'), session[:access_token])
-    Rails.logger.info("A new delayed job with ID " + new_job.id.to_s + " has been created: " + params[:project]) 
-
+    
+    queue = "Default"
+    daacs = ['NSIDC', 'LPDAAC']
+    collections = JSON.parse(params[:project])['collections']
+    daacs.each do |daac|
+      collections.each do |collection|
```
abbottry: @mreese84 how common is it for people to order collections from different providers? This makes the assumption that all collections ordered are from the same provider (at least from a queue perspective) -- I think.

jboukes: @abbottry - you're precisely right.  NSIDC and LPDAAC get a special line since their collections have been having issues (causing a bottleneck, I believe), and all the rest - which are assumed to have good behavior - are all in the 'Default' line.

Other DAACs might get their own lines eventually as well.

mreese84: It's entirely possible that a user will use a project to request data from multiple collections and that those collections could be from different DAACs. How often that happens? I don't know. But I guess I've always assumed that each collection would be its own order. So in this case, if there are 3 collections (NSIDC, LP, and any other DAAC), would they each be sent to their own queue? Or would EDSC handle everything in one queue?

This is probably an edge case, but it is realistic.

jboukes: @mreese84 I experimented with that and found that a project with multiple collections is - surprisingly - treated as a single job.  

I wrote the logic under the philosophy that priority gets the priority line even if it brings friends; therefore, if the logic finds any of the collections are from one of the priority DAACs, then it gets put into that DAACs queue.

This is also true if there is the real edge case of a project with both an NSIDC and an LPDAAC collection - that case would get sent to the first priority queue encountered as it iterates, which would be NSIDC.  The order of the priority DAACs in that array matters - slightly.

abbottry: You guys are responsive, I just came here to write `actually, i dont think that matters does it?` :)

andyfekete: Good catch -- the queuing logic should check all daacs in the order, and if the order includes a priority daac put the order in that daac's queue.

mreese84: @andyfekete, that is what happens unless an order includes collections from multiple priority DAACs. Since it is a single order, it can only go to one queue.

mreese84:
@jboukes, once Francell is finished testing tickets in SIT, please get this deployed to SIT before merging. You can use SIT to test the functionality. Once we have some data on what happens on SIT @abbottry can approve. After he approves you're free to merge this.

app/controllers/data_access_controller.rb
jboukes: @mreese84 - exactly, if it is a project with a collection from more than one priority daac, then the whole thing gets put into the queue of only one of those priority daacs.

This doesn't feel like a problem to me though since projects have always been a 'single job'.  It's just a - 'feature'?  Let's go with 'feature'.

Procfile
andyfekete: Tim -  In the light of what we're just now learning about EDSC restarts, ooms and load capacity, before we deploy this I'd like to re-ask your opinion here.  What is the safest way  allocate workers in the EDSC cluster?  Knowing what we know now, it unwise to run all workers in the same container?

mreese84: @tgoff, can you please respond to @andyfekete's comment here? We need to get this deployed if everything checks out. And we need to come up with an alternative if there's more to be done. Either way, this is a _huge_ priority for NSIDC, so we need to close the loop quickly.

tgoff: I think we could go either way.  separate containers means more isolation, if one queue goes down or is overwhelmed, it does not affect the others.  However, there is some overhead associated with an additional docker container.  we have like 4 extra GB of RAM to work with (though it is possible we can go a bit higher depending on how deployments are done).  We could easily spin up 2 more jobs tasks for those queues if we want.  Ultimately that probably is the most resilient solution

tgoff:
Replying to : @tgoff is there a way we can stop and then restart the workers?
yes, we can scale the jobs down to 0

Procfile
mreese84: @tgoff, I'm not sure I follow 100%. Are you saying it's okay to move forward with things as is and we bump up RAM on the NGAP side? Or something else? I don't know how all of the specifics work, but I just want a clear "yes" or "no" from someone that what we are proposing is okay or not.

andyfekete: So two jobs tasks total, per the hybrid approach you mentioned above (ie - one for hi priority daacs, and the other for everybody else)?

tgoff: well, we actually have 2 jobs currently: 
```
=== jobs
Desired: 2
Running: 2
RAM: 1024
CPU: 256
```
we could make one of those high priority and one normal, or add a third if current capacity is not sufficient.  TO do this, you would replace the single `jobs` line in the Procfile with two lines e.g. `normal_jobs` and `priority_jobs`.

tgoff: yeah, that is all (with each line specifically calling out the queues to work).  Then we can work on making sure the jobs are properly scaled and resourced.  But the more I think about it, the more it seems smart to separate the queues into different instances

jboukes: @mreese84, @tgoff, just to be explicit, is this what we're aiming for now? 

normal_jobs: bundle exec bin/delayed_job --pool=Default run
priority_jobs: bundle exec bin/delayed_job --pool=NSIDC,LPDAAC run

Realize that's only a single worker for 'priority', so if there is a long running task in there, I think it may 'bottleneck' following priority jobs.

tgoff: that looks good.  you can also set priority to 2 runners (and increase memory if needed) and/or scale priority to 2 tasks

andyfekete: Wait -- so if we go to two tasks, we lose the capability of multiple workers per queue?

jboukes: @andyfekete - not at all!  I would think this would be better:

normal_jobs: bundle exec bin/delayed_job --pool=Default:2 run
priority_jobs: bundle exec bin/delayed_job --pool=NSIDC,LPDAAC:2 run

This would allow two workers per task/queue set.  Would this work for everyone?  Let me know and I'll update the PR if so.

mreese84: Works for me if it works for @andyfekete

tgoff: looks good.  we will need to watch RAM and make sure we scale it accordingly.  is there any way to produce an ops like load in sit or UAT?

andyfekete: John, in this scheme is it possible to dedicate one worker to each priority daac? That way one couldn't starve the other.

jboukes: @andyfekete - I believe this should do it:

normal_jobs: bundle exec bin/delayed_job --pool=Default:2 run
priority_jobs: bundle exec bin/delayed_job --pool=NSIDC --pool=LPDAAC run

Procfile updated and pushed up.  :)

andyfekete: Looks good.  Engage!

jboukes: @andyfekete @mreese84 excellent - I'll merge here shortly :)

Procfile
```diff
@@ -1,4 +1,6 @@
 web: ./start.sh
-jobs: bundle exec bin/delayed_job run
+normal_jobs: bundle exec bin/delayed_job --pool=Default:2 run
+priority_jobs: bundle exec bin/delayed_job --pool=NSIDC --pool=LPDAAC run
+worker: bundle exec rake jobs:work
```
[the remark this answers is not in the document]
jboukes: I'm not 100% sure it's not necessary at all - I probably added it as it wasn't working otherwise.  In that same thread is a comment "rake jobs:work is no less production worthy than bin/delayed_job. Use what works for your use case."

If you want to try it without, test it first - though this does feel like a case of "if it's not broken..."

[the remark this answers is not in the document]
andyfekete: That line causes 2 additional docker instances to be spun up, which consume additional resources.  We are already over allocated in ram, so that isn't good.

More concerning is that instead of a clean design with two queues, we now have a solution that we don't quite understand.  

Does "worker: bundle exec rake jobs:work" create a third queue, and receive a combination of high and default priority daacs?  If so, that defeats the priority mechanism we've implemented.

Do log messages clearly indicate when a worker is working on the third "worker" queue vs the "priority jobs" and "normal_jobs" queues?  If not, triage will unnecessarily complicated.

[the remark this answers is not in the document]
andyfekete: https://bugs.earthdata.nasa.gov/browse/EDSC-1826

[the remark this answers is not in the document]
jboukes: I do not believe it "creates" a queue at all, and the log messages should always report which queue they are in (regardless of how a worker is spawned).

The memory consumption is indeed concerning; if this appears to be the culprit, then by all means - let's test without it and see how it does.

PR #588, opened 2018-03-05
EDSC-1618: Implements UMM-Var subsetting

abbottry:


jboukes:
@mreese84 @abbottry - what might be a good collection to use to test?  So far, all collections I've used hit 'This collection doesn't have any associated variables.'

mreese84:
@jboukes we have a wiki page that keeps a lot of this stuff together. https://wiki.earthdata.nasa.gov/display/EDSC/E2E+Services+Details

You can search for C1200187767-EDF_OPS

PR #589, opened 2018-03-12
EDSC-1717: Sets related_urls to empty if there are no Related URLs

jennywagg:
https://bugs.earthdata.nasa.gov/browse/EDSC-1717

There already existed a check (in _collection_details.html.erb) intended to make sure `related_urls` was not empty before displaying "Related URLs: View All Related URLs" on the collection details page. However, even if there were no Related URLs present in a collection, `related_urls` would never itself be an empty array - it was made up of objects to represent Collection URLs, Distribution URLs, Publication URLs, etc...  

The solution in this PR is to leave the check for an empty `related_urls` the same, and to set `related_urls` to actually be empty in the case where there are no Related URLs.

### DIFF
--- a/app/presenters/collection_details_presenter_umm_json.rb
+++ b/app/presenters/collection_details_presenter_umm_json.rb
@@ -207,10 +207,14 @@ class CollectionDetailsPresenterUmmJson < DetailsPresenterUmmJson
       end
     end
 
-    # URLs hsould be listed alphabetically and grouped by Type
+    # URLs should be listed alphabetically and grouped by Type
     related_urls.each do |url_category|
       url_category[:urls].sort_by! {|url| [url['Type'], url['Subtype'], url['URL']]}
     end
+
+    # related_urls should be empty if there are no Related URLs
+    related_urls.each {|x| if !(x[:urls].empty?) then return related_urls end}
+    related_urls = []
 
     related_urls
   end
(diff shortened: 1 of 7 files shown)


app/presenters/collection_details_presenter_umm_json.rb
```diff
@@ -207,11 +207,15 @@ def related_urls(collection)
...
+    # URLs should be listed alphabetically and grouped by Type
     related_urls.each do |url_category|
       url_category[:urls].sort_by! {|url| [url['Type'], url['Subtype'], url['URL']]}
     end
 
+    # related_urls should be empty if there are no Related URLs
+    related_urls.each {|x| if !(x[:urls].empty?) then return related_urls end}
+    related_urls = []
+
     related_urls
```
jboukes: I like this strategy!

Since you're returning an empty array when related_urls is empty, I wonder if the assignment 'related_urls = []' is even necessary.  Line 216 appears to return related_urls in the case that *any* of the sub arrays are *not* empty - so our 'success' condition is covered.  So, you should be able to delete line 219 (and 218) entirely and change 217 to just return an empty array, such as this:

related_urls.each {|x| if !(x[:urls].empty?) then return related_urls end}
[]

jennywagg: This totally makes sense, I agree it should be simplified.

PR #590, opened 2018-03-20
EDSC-1796: Updates Gems

abbottry:


Gemfile.lock
```diff
@@ -148,7 +148,7 @@ GEM
     launchy (2.4.2)
       addressable (~> 2.3)
     libv8 (3.16.14.15)
-    loofah (2.1.1)
+    loofah (2.2.1)
```
littleidiot40: Shouldn't you enforce this in the Gemfile with a 

gem 'loofah', '>= 2.2.1'

?

abbottry: It's a dependency, not a gem we've installed.

littleidiot40: Yeah but our rule of thumb used to be 'never manually update Gemfile.lock'

PR #601, opened 2018-04-09
EDSC-1828: Updated facets feature 'Subsetting Services' to 'Customizable'

jboukes:
This PR updates the facets feature label as well as associated tests and fixtures.


andyfekete:
Tooltip suggestion -- instead of  "Supports custom service options to be performed. (e.g. subsetting, etc.),"  how about  "Supports data customizations (e.g. subsetting, reformatting, etc.)" ?

reasoning: "Service" doesn't really mean anything to the user.

mreese84:
@andyfekete, I am using text from another portion of the app that was approved by ESDIS. For consistency, and to keep this ticket within scope, I'd prefer to use previously approved copy.

mreese84:
@jboukes the badge itself still says "subsetting" and should say "Customizable"

screen shot 2018-04-11 at 9 07 49 am

PR #602, opened 2018-04-19
EDSC-1809: Temporarily Disable status page polling

abbottry:


tgoff:
should we add some text telling the user that they will have to refresh to get updates?

PR #604, opened 2018-04-25
EDSC-1807: Improved Granule Panel CTA

jsiarto:


mreese84:
@abbottry what's the easiest way to get these fixture conflicts resolved?

abbottry:
Probably easiest and fastest to just re-record.

app/assets/stylesheets/modules/master_overlay.css.scss
```diff
@@ -333,21 +354,48 @@
...
 
   .retrieve {
     background: #27ae60;
     border: none;
     color: #fff;
-    padding-left: 10px;
-    padding-right: 10px;
 
     &:hover {
       background: #219853;
```
mreese84: I realize you didn't touch this, but other buttons throughout the app have a transition between normal and hover states. Pretty sure we have a mixin for it somewhere. Would be nice to throw that on here.

app/views/search/_granule_details.html.erb
```diff
@@ -3,7 +3,7 @@
 <section id="granule-details" class="details master-overlay-no-min">
   <nav class="master-overlay-nav">
     <a href="#" class="master-overlay-back" data-bind="click: ui.collectionsList.focused().hideGranuleDetails">
-      <i class="fa fa-arrow-circle-left"></i> <span class="master-overlay-breadcrumb">Back to Granules</span>
+      <i class="fa fa-chevron-circle-left"></i> <span class="master-overlay-breadcrumb">Back to Granules</span>
```
mreese84: Missed removing the "report a metadata problem" link. Look at lines 19-26 of this file.

app/views/search/_granule_list.html.erb
```diff
@@ -1,76 +1,68 @@
...
+            <i class="fa fa-chevron-circle-left"></i> Back to Collections
           </a>
         </nav>
         <!-- ko if: ui.collectionsList.focused() && ui.collectionsList.focused().granules()  -->
         <!-- ko with: ui.collectionsList.focused() -->
         <header class="with-options">
-          <h2 data-bind="text: collection.title"></h2>
-          <!-- ko with: collection -->
+          <div class="collection-title-filters">
+            <h2><span data-bind="text: collection.title"></span> <!-- ko with: collection --><a href="#" class="master-overlay-forward collection-title-link" data-master-overlay-forward="Granules" data-bind="click: $root.ui.collectionsList.showCollectionDetails" title="View collection details">View details</a><!-- /ko --></h2>
```
mreese84: Any thoughts on including `fa-info-circle` with the "View details" link? That may help bridge the gap between old and new. I could see some people not catching the new way. Up to you, but worth asking.

app/views/search/_granule_list.html.erb
```diff
@@ -1,76 +1,68 @@
...
+                  Granule Search: <input id="granule-ids" type="text" placeholder="Search Single or Multiple Granule IDs..." data-bind="value: $data.collection.granuleIds, valueUpdate: 'keyup'" data-tooltip="true" title="Wildcards: * (asterisk) matches any number of characters, and ? (question mark) matches exactly one character.
+    Delimiters: Separate multiple granule IDs by space, comma, or new line."/>
+                </p>
+              <!--/ko-->
+
+              <!-- ko with: collection -->
+                <!-- ko if: has_granules -->
+                <a href="#" data-bind="visible: !$root.project.isSearchingGranules($data), click: $root.ui.projectList.showFilters, css: {'button-highlighted': granuleFiltersApplied()}" class="master-overlay-hide-parent master-overlay-show-secondary granule-filter-link" title="Filter granules">Granule filters</a>
+                <a href="#" data-bind="visible: $root.project.isSearchingGranules($data), click: $root.ui.projectList.hideFilters, css: {'button-highlighted': granuleFiltersApplied()}" class="master-overlay-hide-parent master-overlay-hide-secondary granule-filter-link" title="Hide granule filters">Granule filters</a>
```
mreese84: If we add the `fa-info-circle` to the view details link, then we should also add the filter icon on the Granule Filters link as well.

mreese84:
This looks really good. Most of my comments were pretty small. The only major thing is to consider how to handle a user adding lots (I mean LOTS) of granules. Check out what happens when I add a couple million...

Download button wraps
screen shot 2018-04-25 at 8 23 30 am

"Remove from project" gets pushed to the next line down
screen shot 2018-04-25 at 8 23 37 am

jsiarto:
@mreese84 Agree with all points above, I'll push a small update. I didn't test the button state with millions of granules--I think I only looked to 6 digits, so nice find. Not sure what I'm going to do about the "remove from project" link -- it's just wordy and I don't have an alternate solution that won't cause a UX problem.

mreese84:
@jsiarto, we could decide to start rounding and give approximate granule counts once a certain threshold is reached. Like, instead of 2,393,411 granules we could say 2.4M Granules. That would help for sure.

abbottry:
I wish when we displayed large numbers, we had commas. It's hard at first glance to determine if I'm looking at `100,000` or `1,000,000` -- this seems like an easy thing to do with not a ton of effort. I suppose we can move it to another ticket but real estate already seems to be minimal. I like @mreese84 idea of rounding up in certain places. `100,000` would look much different than `1M`.

abbottry:
Looks like there is collateral damage with your updates. Clicking through to your project after adding a collection results in the attached screenshots.

screen shot 2018-04-26 at 8 58 18 am

If I click on `Visualize all Collections` (which I don't know what this is even supposed to do) it toggles the button to 👇 

screen shot 2018-04-26 at 8 58 26 am

PR #605, opened 2018-04-25
EDSC-1729: Support ordering AE_L2A through new project flow

abbottry:


jboukes:
No review required to merge - is that because it is e2e services?  It might be good to add this safe guard (this comment has nothing to do with this specific PR, obviously).

app/assets/javascripts/models/page/project_page.js.coffee
```diff
@@ -139,24 +134,25 @@ ns.ProjectPage = do (ko,
...
-            else
-              collection.total_size(@_convertSize(totalSize)['size'])
-              collection.unit(@_convertSize(totalSize)['unit'])
-              @sizeProvided(true)
+          if collection.granuleDatasource()
+            granules = collection.granuleDatasource().data()
+            if granules.isLoaded()
+              loadedCollectionNum += 1
+              _size = 0
+              _size += parseFloat(granule.granule_size ? 0) for granule in granules.results()
```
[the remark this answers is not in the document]
abbottry: potentially -- but I'm going to wave the 'scope' flag here, I didnt actually modify this, I just added `if collection.granuleDatasource()` to ensure that it was loaded on the project page.

app/assets/javascripts/models/page/project_page.js.coffee
```diff
@@ -139,24 +134,25 @@ ns.ProjectPage = do (ko,
...
+              totalSize = _size / granules.results().length * granules.hits()
+              totalSize = 0 if isNaN(totalSize)
+              projectGranules += granules.hits()
+              collection.granule_hits(granules.hits())
+              projectSize += totalSize
+              if isNaN(totalSize) || (totalSize == 0 && granules.hits() > 0)
```
jboukes: isNaN(totalSize) will never be true here, as totalSize is set to 0 in the case that it is NaN on previous line 144.  Since that part of the conditional will never be true, this could probably be simplified to:

```
if (totalSize == 0 && granules.hits() > 0)
```

app/controllers/data_access_controller.rb
```diff
@@ -300,59 +348,88 @@ def get_downloadable_access_methods(collection_id, granules, granule_params, hit
...
+        access_methods[option_id][:count] += 1
       end
     end
 
-    defs = defs.map do |option_id, config|
-      option_def = echo_client.get_option_definition(option_id, token).body['option_definition']
-      if option_def['deprecated']
+    # Generally this will end up being a single record
+    access_methods = access_methods.map do |option_id, config|
+      Rails.logger.debug option_id
```
jboukes: Is this intended to commit?  I only ask as it is 'debug' - wondering if it snuck in there

app/controllers/projects_controller.rb
```diff
@@ -1,4 +1,5 @@
 class ProjectsController < ApplicationController
+  # before_filter :set_env_session
```
jboukes: Commented out code - do we want this in the commit as well?  It's the only thing updated in this file so it seems odd.

mreese84:
@abbottry, as we just discussed, the links in the script aren't getting the spatial parameters set properly. We've got some question marks at the end of them. I can provide you with the script I got if needed.

PR #608, opened 2018-04-30
EDSC-1729: Fixes back button

abbottry:


mreese84:
Broken test is broken.

PR #620, opened 2018-05-23
added js to ORNLDAAC portal

eniad:
portal now includes collections without granules by default
ORNL DAAC has several service collections (MODIS subsetting)
  that have no ganules


eniad:
failed, redoing request

PR #632, opened 2018-06-27
EDSC-1897: Adds ability to stage data through new project flow

abbottry:


abbottry:
@mreese84 before this is merged I'd like to cherry-pick EDSC-1884 once its approved and merged into this PR.

PR #642, opened 2018-07-17
EDSC-1916: Enables Google Tag Manager tag

trevorlang:


trevorlang:
This now contains work from EDSC-1916 and EDSC-1919 in order to make deployment a little easier

PR #647, opened 2018-07-24
EDSC-1922: Don't send nlp request if query is blank

macrouch:


app/assets/javascripts/modules/nlp_parsing.js.coffee
```diff
@@ -37,11 +37,12 @@ do ($=jQuery, currentPage = window.edsc.models.page.current, ajax=@edsc.util.xhr
 
       query = $nlpInput.val()
       currentPage.query.originalKeywords(query)
-      ajax
-        dataType: 'json'
-        url: "/extract_filters?q=#{query}&rerun=#{immediateReenter}&previous_q=#{previousKeyword}"
-        success: (data) =>
-          _applyParsedText(data)
+      unless query == ''
```
abbottry: Can we strip whitespace on this? I'm afraid that user entering '   ' would still send a request.

PR #648, opened 2018-07-25
EDSC-1922: Reset the keyword query without sending an nlp request

macrouch:


app/assets/javascripts/modules/nlp_parsing.js.coffee
```diff
@@ -37,7 +37,9 @@ do ($=jQuery, currentPage = window.edsc.models.page.current, ajax=@edsc.util.xhr
 
       query = $nlpInput.val().trim()
       currentPage.query.originalKeywords(query)
-      unless query == ''
+      if query == ''
```
abbottry: Previously this would have called `_applyParsedText` with the response from NLP. It seems as though instead of modifying the query object directly we could call `_applyParsedText` with 
```
{
    "edscSpatial": null,
    "edscTemporal": null,
    "keyword": ""
}
```
Just so that we can use that method as the single place to make these changes. **This may be why we were pinging NLP even when the user didn't supply a search.**

macrouch: I went back and forth between using `_applyParsedText` and just updating the query, I'll update

PR #656, opened 2018-08-03
EDSC-1905: Adds ability to download collections in End to End Services

abbottry:


mreese84:
Functionality is there, so that's good. There's room for improvement in the UX, as we discussed in Slack.

We need some sort of loading state so that the user knows something is happening/loading when they visit the page. The wait is too long without any indication as it sounds. @trevorlang can help there.

trevorlang:
I have just pushed my improvements. Here's the basics of what I have changed.
 
- Rather than use the button and spinner to show the loading state, I am now showing a filler animation that looks similar to the content that will be loaded in. 
- Now each collections' details are available as soon as they are loaded, and it will no longer wait until all the collections have finished to show them. This helps with perceived performance, making things seem a little faster. 
- I have added logic that will prevent the Download Data button from going active until all collection access methods have been loaded in. This should prevent users from clicking download before seeing the Customize option load, which was happening previously
- Eventually we can add the filler animation to the project stats as well, but there was a little hang up there that I want to sync up with Ryan on. I didnt think it warranted holding up this ticket.

Here is an example showing what it would look like along with a customizable project:
http://localhost:3000/projects/new?p=!C1200240776-DEV08!C3870-LPDAAC_ECS!C1000000082-EDF_OPS!C1200187767-EDF_OPS&ff=Near%20Real%20Time&fdc=EDF_OPS

PR #666, opened 2018-08-14
EDSC-1948: Implementing GTM events

trevorlang:


config/environments/development.rb
```diff
@@ -33,6 +33,7 @@
   # Uncomment the config.analytics_id to enable the development Google Analytics tracking
   # config.analytics_id = 'UA-50960810-4'
   # config.tag_manager_id = 'GTM-WNP7MLF'
+  config.tag_manager_id = 'GTM-WNP7MLF'
```
macrouch: This is the same as the commented out line above

abbottry: @macrouch feel free to commit that change before approving the PR.

PR #673, opened 2018-08-21
EDSC-1888: Improving display of disabled project download button

trevorlang:
Took this as an opportunity to flatten the standard buttons. Time would be well spent double checking all cases were covered. I think I got them all, but there may be a few hidden buttons I am not aware of.

### DIFF
--- a/app/assets/stylesheets/modules/buttons.css.scss
+++ b/app/assets/stylesheets/modules/buttons.css.scss
@@ -42,26 +42,28 @@
 }
 
 @mixin dark-button {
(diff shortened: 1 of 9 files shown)


mreese84:
screen shot 2018-08-22 at 11 07 02 am
 On my screen, the "add to project" and "metadata" buttons are a little hard to distinguish. I'd say let's lighten up the background color on those blue/greyish rows so those buttons are a little easier to see.

mreese84:
Missed the "back" button on the data access page.

screen shot 2018-08-22 at 11 20 07 am

PR #674, opened 2018-08-21
EDSC-1949: Find cmr_env in config.js.coffee.erb without using Rails

macrouch:


abbottry:
Not a bug fan of reading the URL -- did the `assets:precompile` option not work?

macrouch:
Nothing I tried was working, but I wasn't able to completely figure out how assets get compiled.

PR #681, opened 2018-09-07
EDSC-1952: Make room for granule thumbnails

macrouch:


app/assets/stylesheets/modules/projects.css.scss
```diff
@@ -321,6 +321,9 @@ $card-background: rgba(235,235,235,1);
   }
 
   .modal-body {
+    ul > li {
```
trevorlang: In general, I would prefer if we used a class specific to this element, rather than an overly specific selector like `.modal-body ul>li`. It's a minor thing, and I am not sure worth changing at this point as we will likely revamp this with some upcoming changes to the Project page.

I noticed also that we are using the `.granule-list-thumbnail-container` in that list, outside of the a `.granule-list` component. This type of thing will make the app much more fragile when we want to make changes in the future. We should avoid doing things like this. Again, not sure if its worth cleaning up at this point, but just something to look out for.

PR #685, opened 2018-09-17
EDSC-1874: Fixing panel dragging issues

trevorlang:


app/assets/javascripts/modules/master_overlay.coffee
```diff
@@ -179,6 +179,15 @@ do (document, window, $=jQuery, config=@edsc.config, plugin=@edsc.util.plugin, p
         height = windowHeight - main.position().top - $('body > footer').outerHeight()
         main.height(height)
 
+        # If the timeline is open, let's check to see if the main overlay panel is within 70 px
```
macrouch: where did 70px come from?

trevorlang: 70px is the height of the search bar and it’s container

PR #691, opened 2018-10-02
EDSC-1927: If nlp_parser_url doesn't exist, don't use NLP

macrouch:


mreese84:
@abbottry, the purpose of this work is specifically to make local development easier for external devs. Obviously it reduces the need to be on the VPN for EED devs too. But this won't really be an issue to end users since NLP will be available on Production EDSC. It has been remarkably stable/available since its initial deployment.

So, comment noted, and it's a good point to remember when writing tests. But the benefit outweighs the "gotchas".

PR #694, opened 2018-10-04
EDSC-1990: Run EDSC development in docker containers

macrouch:


macrouch:
I don't know what happened, but Git won this battle

PR #730, opened 2018-12-13
EDSC-2049: Forcing granule visibility by default

trevorlang:


app/assets/javascripts/modules/map/leaflet-plugins/granule_grid_layer.js.coffee
```diff
@@ -37,12 +38,22 @@ ns.GranuleGridLayer = do (
 
   class GranuleGridLayer extends L.GridLayer
     initialize: (@collection, color, @multiOptions) ->
+      console.warn 'collection', @collection
```
abbottry: oops :)

app/assets/javascripts/modules/map/leaflet-plugins/granule_grid_layer.js.coffee
```diff
@@ -37,12 +38,22 @@ ns.GranuleGridLayer = do (
...
       if @collection.granuleDatasource()
         @granules = @collection.granuleDatasource().data()
       else
         @_datasourceSubscription = @collection.granuleDatasource.subscribe(@_subscribe)
 
-      @color = color ? '#25c85b';
+      # If on the search page, the color should always be green.
+      if page.current.page() == 'search'
+        color = '#25c85b'
```
abbottry: Feels like we may want to use a class, or store these colors in variables or configs some how, hard-coding hex values seems like a bad idea.

app/assets/javascripts/modules/map/leaflet-plugins/granule_grid_layer.js.coffee
```diff
@@ -37,12 +38,22 @@ ns.GranuleGridLayer = do (
...
+
+      # If were on the project page, set it to blue while we finish loading the collection data.
+      if page.current.page() == 'project'
```
abbottry: This doesn't seem overly bad but we should ensure we are making this check consistently -- I've seen similar things before checking the URL, is this a better way of making this check?

app/views/shared/components/project_list/_project_list.html.erb
```diff
@@ -14,7 +14,7 @@
                 <div class="dropdown-fancy-content">
                   <ul>
                     <li>
-                      <button class="button-naked project-list-item-toggle" type="button" name="button" data-bind="click: toggleVisibility">
+                      <button class="button-naked project-list-item-toggle" type="button" name="button" data-bind="click: function() { toggleVisibility() }">
```
abbottry: What is the reason for the anonymous function here?

app/assets/javascripts/models/page/search_page.js.coffee
```diff
@@ -78,6 +78,7 @@ ns.SearchPage = do (ko
 
       @bindingsLoaded = ko.observable(false)
       @labs = ko.observable(false)
+      @showFocusedCollections = ko.observable(true)
```
abbottry: Does this mean we can only show or hide all layers? What defines 'focused' in this context?

mreese84: I don't know what this means, but the current Prod EDSC does have the show/hide all layers/collections on the project page. It would be nice if we kept that functionality. It would require a new button on the E2E project page though as the new one doesn't have that button (you have to do them individually).

app/assets/javascripts/modules/map/leaflet-plugins/granule_grid_layer.js.coffee
trevorlang: This was something I've introduced as a base class for pages. That seems like a bit cleaner, more maintainable way to do it, rather than relying on an href. I like that it ties into Knockout-land a little better. Interested to hear any thoughts if you have any.

app/views/shared/components/project_list/_project_list.html.erb
trevorlang: When using the function directly in the click, it was trying to pass the scope of the button as a parameter in the `toggleVisibility` function call, which was causing some issues with the way the function was called. This seemed like a good way to prevent that from happening.

app/assets/javascripts/modules/map/leaflet-plugins/granule_grid_layer.js.coffee
trevorlang: Totally agree here. I will work something up.

app/assets/javascripts/modules/map/leaflet-plugins/granule_grid_layer.js.coffee
trevorlang: Nice catch

app/assets/javascripts/modules/map/leaflet-plugins/granule_grid_layer.js.coffee
abbottry: 👍

app/views/shared/components/project_list/_project_list.html.erb
abbottry: 👍

app/assets/javascripts/models/data/colors.js.coffee
```diff
@@ -0,0 +1,19 @@
...
+#= require models/data/variable
+
+ns = @edsc.models.data
+
+ns.Colors = do () ->
+
+  class ColorsModel
+    constructor: () ->
+      # Defines the color values for collections
+      @collections = ko.observableArray([
```
macrouch: Instead of this being an array, why don't we change it to just being properties on this class.
```
class ColorsModel
  constructor: ->
    @blue = '#3498DB'
    ...
    ...
```
then in `project.js.coffee` you could do 
```
colorPool = new ValuePool([
     colors.blue,
     colors.orange,
     colors.green,
     colors.red,
     colors.purple
   ])	   
```

PR #731, opened 2018-12-14
EDSC-2054: Prepopulate email and remember form input

macrouch:


app/controllers/application_controller.rb
```diff
@@ -4,6 +4,7 @@ class ApplicationController < ActionController::Base
   protect_from_forgery
 
   before_filter :refresh_urs_if_needed, except: [:logout, :refresh_token]
+  before_filter :retrieve_preferences_if_needed, except: [:logout, :refresh_token]
```
macrouch: This never finds new preferences, should happen on login

termux-io:
`b10c38d0450fdc947725681d1715a231bf2db2df`

PR #732, opened 2018-12-18
EDSC-2056: Update CWIC calls to use https

abbottry:


macrouch:
not sure, but maybe this needs to be changed too? edsc/plugins/datasource/cwic/src/js/edsc-plugin.datasource.cwic.jsx:13

PR #735, opened 2018-12-20
EDSC-2060: Sort science keywords and variables when variable subsetting

macrouch:


abbottry:
Still showing `(Suite X of 0)`

PR #738, opened 2018-12-31
Minor formatting proposals mainly on code paragraphs

kant:


mreese84:
@kant I agree with @abbottry. Would you be able to make those quick changes? Then we'll get this pulled.

abbottry:
Closing due to inactivity.

PR #744, opened 2019-01-08
EDSC-2051: Improving display of metadata quality notice

trevorlang:


trevorlang:
This ticket moves the metadata quality notice to a more prominent location and cleans up the design a bit. I will be filing a follow-up ticket to look at how we can improve the display of the "Edit Options" panel, as I feel it is getting a little cluttered and stale. I think we have a few other improvements lined up for the panel and they should probably all be looked at in the same pass.

PR #747, opened 2019-01-12
EDSC-2057: Pulls out user information from ordering

abbottry:
This pulls out additional queries to multiple endpoints to construct contact information for orders. This information will now be pulled and stored on login, and queried from the database when needed for the purposes of placing an order.


app/controllers/concerns/authentication_utils.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    current_user.save
+  end
+
+  def retrieve_urs_profile(oauth_response)
+    # Once we retrieve this from URS we can use the `uid` value but the oauth
+    # endpoint only returns the path to the user
+    username = oauth_response['endpoint'].gsub('/api/users/', '') if oauth_response['endpoint']
+
+    byebug
```
macrouch: `byebug`

abbottry: yikes.

PR #750, opened 2019-01-27
EDSC-2058: Splits out job processings methods into standalone jobs

abbottry:


app/controllers/concerns/granule_utils.rb
```diff
@@ -77,6 +77,7 @@ def fetch_ous_response(options)
     ous_params['bounding_box'] = project_params['bounding_box'] if project_params.key?('bounding_box')
     ous_params['temporal']     = project_params['temporal'] if project_params.key?('temporal')
 
+    puts ous_params
```
macrouch: puts

app/controllers/oauth_tokens_controller.rb
```diff
@@ -24,25 +24,53 @@ def refresh_token
     if json
       render json: { tokenExpiresIn: script_session_expires_in }
     else
-      render json: nil, status: 401
+      render json: nil, status: :unauthorized
     end
   end
 
-  protected
+  def store_user_data(oauth_response)
```
macrouch: Why did you remove `protected`?

abbottry: No reason, must have been removed when I removed other stuff in that file. I'll replace it.

app/controllers/concerns/granule_utils.rb
abbottry: Last minute debug -- good catch.

app/models/order.rb
```diff
@@ -0,0 +1,25 @@
+class Order < ActiveRecord::Base
+  belongs_to :retrieval_collection
+  belongs_to :orderable, polymorphic: true
```
abbottry: Remove

PR #769, opened 2019-03-04
 EDSC-2134: Create clean branch as starting point for EDSC 2.0

macrouch:


CONTRIBUTING.md
```diff
@@ -1,206 +0,0 @@
-# Making Changes
-
```
trevorlang: I feel like we should keep this and update, no?

static/src/js/components/Header/Header.js
```diff
@@ -1,101 +1,34 @@
...
+const Header = () => (
+  <header className="header">
+    <ul className="header__menu">
+      <li className="header__menu-item">
+        <h1 className="header__menu-title">
+          <Link
+            className="header__menu-link"
+            to="/search"
+          >
+              CSW User Interface
```
trevorlang: Leftovers :)

static/src/js/containers/SidebarContainer/Header/Header.scss
```diff
@@ -6,22 +6,22 @@
 .header__site-logo {
   display: flex;
   align-items: flex-end;
-  background-image: url('../../../../../assets/images/app-logo.png');
+  background-image: url('../../../../assets/images/app-logo.png');
```
trevorlang: I don't think we need to fix it now, but I am not a huge fan of these relative links

static/src/js/components/Header/Header.js
macrouch: :doh:

abbottry:
Looks like by removing the `.gitignore`, the items that were in that file will show up as changed files if we remove them after pulling this branch down.

Examples: 

- .byebug_history
- config/
- db/
- doc/
- tmp/
- vendor/

Should we keep these items in the new `.gitignore` until they are removed from all of our local dev environments?

abbottry:
Are the git submodules for echoforms and timeline remaining in the application?

abbottry:
I'm still seeing the `performance` folder.

abbottry:
Keeping `bin/ecc-sync` for potential reuse?

abbottry:
Can we add `.DS_Store` to the `.gitignore`?

PR #772, opened 2019-03-05
EDSC-2152: Keep checking order status if state comes back empty

macrouch:


abbottry:
As you suggested yourself, I think you should add `.set(queue: queue_name)` to the `OrderStatusJob` in `process_retrieval_job.rb`. This should catch a vast majority of the issue discussed, and this current solution will be a fallback instead (you should leave it, though it's far from ideal).

PR #773, opened 2019-03-07
EDSC-2084: Added capability to use shapefile from search in esi order

newlarry:
Caveats:
* A polygon with many points ( like state/country boundaries found on the Internet ) will cause errors in the icesat2 subsetter
* Some KML/Shapefiles when converted to geojson have some characters that ESI complains about.
Will have to work with SDPS team to add more stability to these cases.

### DIFF
--- a/app/jobs/submit_catalog_rest_job.rb
+++ b/app/jobs/submit_catalog_rest_job.rb
@@ -5,6 +5,13 @@ class SubmitCatalogRestJob < ActiveJob::Base
   # @param base_url [String] the base url to provide to the user in their order to check on for status updates
   def perform(order, base_url)
     request_url = "#{base_url}/data/retrieve/#{order.retrieval_collection.retrieval.to_param}"
+    shapefile = nil
+
+    if order.search_params.key?('sf') and not order.search_params['sf'].blank?
+      shapefile = Shapefile.find(order.search_params['sf'].to_i)
+      shapefile = shapefile.nil? ? nil : shapefile.file
+      order.search_params.delete('sf')
+    end
 
     order.submit! do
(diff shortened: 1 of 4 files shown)


app/jobs/submit_catalog_rest_job.rb
```diff
@@ -5,14 +5,22 @@ class SubmitCatalogRestJob < ActiveJob::Base
   # @param base_url [String] the base url to provide to the user in their order to check on for status updates
   def perform(order, base_url)
     request_url = "#{base_url}/data/retrieve/#{order.retrieval_collection.retrieval.to_param}"
+    shapefile = nil
+
+    if order.search_params.key?('sf') and not order.search_params['sf'].blank?
```
macrouch: I think just to match the rest of the project we should change `and not` to be `&& !`

app/services/esi_client.rb
```diff
@@ -11,9 +13,10 @@ def self.get_multi_esi_request(*args)
     new.get_multi_esi_request(*args)
   end
 
-  def submit_esi_request(retrieval_collection, granule_params, request_url, token)
+  def submit_esi_request(retrieval_collection, granule_params, request_url, token, shapefile = nil)
     service_url = get_service_url(retrieval_collection.collection_id, retrieval_collection.client, token)
     options = {}
+    @shapefile = shapefile
```
macrouch: Instead of using an instance variable here, just pass the shapefile in as a parameter to `add_shapefile`

macrouch:
@newlarry 	I can't edit this since it is in your fork. I'm going to merge it into a new branch in the NASA repo, then add some tests

PR #789, opened 2019-03-15
EDSC-2120: Improving customization panel navigation

trevorlang:


app/assets/javascripts/models/data/project.js.coffee
```diff
@@ -346,6 +358,11 @@ ns.Project = do (ko,
       return false for collection in @accessCollections() when !collection.serviceOptions.readyToDownload()
       true
 
+    _computeAllHaveAccessMethod: ->
```
macrouch: Should this also check to make sure that all collections have granules, like `_computeAllReadyToDownload` above? That is something I just added for if a user uses the timeline and ends up with 0 granules in a collection

trevorlang: Ive instead added a new observable for `allHaveGranules` that we can use to check and display a more specific error to the user. I've also made it so that when a collection has no granules, it will trigger the invalid state on the list and panel.

PR #794, opened 2019-03-21
Edsc 2084

newlarry:
Disable the use-shapefile checkbox in echoform if present and no shape file was uploaded during search

### DIFF
--- a/spec/features/data_access/service_options_shapefile_spec.rb
+++ b/spec/features/data_access/service_options_shapefile_spec.rb
@@ -4,7 +4,7 @@
 require 'rails_helper'
 require 'rake'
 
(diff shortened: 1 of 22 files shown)


spec/features/data_access/service_options_shapefile_spec.rb
```diff
@@ -0,0 +1,37 @@
+# EDSC-2084
+# Disable the use-shapefile checkbox if there was no upload during search
+
+require 'rails_helper'
+require 'rake'
+
+describe 'Service options order with use shapefile field', pending_updates: true, single_granule: true do
```
abbottry: Will need to remove `, pending_updates: true, single_granule: true` in order for this test to run

PR #800, opened 2019-04-04
EDSC-2139: Enabling facet search

trevorlang:


.eslintrc
```diff
@@ -40,7 +40,17 @@
     "no-console": "off",
 
     # Allow console log messages
-    "import/no-extraneous-dependencies": ["error", {"devDependencies": true}]
+    "import/no-extraneous-dependencies": ["error", {"devDependencies": true}],
+
+    # Disable the depricated label-has-for rule
```
abbottry: deprecated :)

static/src/js/actions/facets.js
```diff
@@ -0,0 +1,70 @@
...
+ * @param {object} newParams - An object containing the params from the apply/remove link
+ * @param {function} dispatch - A dispatch function provided by redux.
+ */
+
+export const changeCmrFacet = (e, newParams) => (dispatch) => {
+  if (e.target.checked) {
+    dispatch({
+      type: ADD_SELECTED_CMR_FACET,
+      payload: {
+        science_keywords_h: newParams.science_keywords_h ? newParams.science_keywords_h : undefined,
```
abbottry: Using the ternary here wont these parameters always be sent to CMR? It may be more verbose but I think it would be best to prevent sending undefined params to CMR.

static/src/js/actions/index.js
```diff
@@ -8,16 +8,30 @@ import {
...
   getCollections,
   getGranules,
   changeQuery,
   changeMap,
   changeFocusedCollection,
-  clearFilters,
   changeUrl,
-  searchNlp
+  clearFilters,
+  searchNlp,
```
abbottry: Alpha order? Or are we grouping these by something?

static/src/js/actions/ui.js
```diff
@@ -0,0 +1,19 @@
+export const masterOverlayPanelDragStart = data => (dispatch) => {
+  dispatch({
+    type: 'MASTER_OVERLAY_PANEL_DRAG_START',
```
abbottry: Shouldn't we use constants here?

static/src/js/components/MasterOverlayPanel/MasterOverlayPanel.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+const MasterOverlayPanel = (props) => {
+  const { children, panelHeight } = props
+
+  return (
+    <div className="search__panel" style={{ height: `${panelHeight}px`, willChange: 'height' }}>
```
abbottry: Inline styles? I thought that was bad?

static/src/js/containers/CollectionResultsContainer/CollectionResultsContainer.scss
```diff
@@ -12,16 +12,25 @@
...
-  padding: $spacer 1.25rem;
   background: $color__white;
 }
 
+.coll-res__tab-handle {
+  position: absolute;
+  top: 0;
+  height: 10px;
+  width: 100%;
+  background: $color__white url('../../../assets/images/drag.png') no-repeat center 5px;
```
abbottry: Is there an option to use a variable or something to define the asset path instead of having to back out 3 folders?

static/src/js/containers/SidebarContainer/Facets/Facets.js
```diff
@@ -1,137 +1,151 @@
...
+    title: 'Projects',
+    changeHandler: changeCmrFacet
+  }
+
+  const processingLevels = {
+    title: 'Processing levels',
+    changeHandler: changeCmrFacet
+  }
+
+  const facetsTemplate = [
```
abbottry: Alpha order?

static/src/js/containers/SidebarContainer/Facets/Facets.js
```diff
@@ -1,137 +1,151 @@
...
-  onToggle() {
-    this.setState(state => ({ isOpen: !state.isOpen }))
-  }
+  const changeCmrFacet = (e, facetLinkInfo) => {
+    const newParams = qs.parse(queryParamsFromUrlString(facetLinkInfo.destination))
 
-  render() {
-    const { facet } = this.props
-    const { isOpen } = this.state
+    const paramsToSend = {
```
abbottry: Alpha order?

static/src/js/containers/SidebarContainer/Facets/Facets.scss
```diff
@@ -42,11 +47,16 @@
   align-self: flex-end;
 }
 
+.facets__selected {
+  padding: 0.2rem 0.4rem 0.2rem 1.1rem;
+  font-size: 0.875rem;
+  color: $color__black--400;
```
abbottry: Can we or do we want to define our colors as variables?

static/src/js/containers/SidebarContainer/Facets/FacetsItem.js
```diff
@@ -0,0 +1,94 @@
...
+
+  onFacetChange(changeHandlerArgs, e) {
+    const { changeHandler } = this.props
+    const { applied } = this.state
+
+    this.setState({
+      applied: !applied
+    })
+
+    changeHandler(e, changeHandlerArgs)
```
abbottry: Generally I've seen these named `onChangeHandler` is that a convention or am I looking too deep into tutorials?

static/src/js/reducers/ui.js
```diff
@@ -0,0 +1,50 @@
...
+    dragging: false,
+    height: 500,
+    clickStartY: undefined,
+    clickStartHeight: undefined
+  }
+}
+
+const uiReducer = (state = initialState, action) => {
+  switch (action.type) {
+    case 'MASTER_OVERLAY_PANEL_DRAG_START': {
```
abbottry: Should we be using constants here?

static/src/js/containers/SidebarContainer/Facets/FacetsList.js
```diff
@@ -0,0 +1,61 @@
...
+    changeHandler
+  } = props
+
+  let orderedFacets = []
+
+  if (liftSelectedFacets) {
+    // https://stackoverflow.com/questions/17387435/javascript-sort-array-of-objects-by-a-boolean-property#comment25241651_17387454
+    orderedFacets = facets.sort((a, b) => b.applied - a.applied)
+  } else {
+    orderedFacets = facets.map(facet => facet)
```
trevorlang: Double check this

.eslintrc
trevorlang: Even though this is depricated, its still in the AirBnb config, so we need to do some manual workaround temporarily

abbottry: I was talking about the spelling haha -- sorry.

static/src/js/actions/facets.js
trevorlang: Any undefined value will not be set in the Redux store. This is really making sure that any new facet request overwrites missing values in the store. Otherwise, they would stick around once we spread out the values in the reducer.

static/src/js/actions/index.js
trevorlang: I alpha ordered all the things. Just to make _you_ happy, Ryan :)

static/src/js/actions/ui.js
trevorlang: Good catch

static/src/js/components/MasterOverlayPanel/MasterOverlayPanel.js
trevorlang: Moved the will-change to css. Need to keep the height so it can be dynamically updated.

static/src/js/containers/CollectionResultsContainer/CollectionResultsContainer.scss
trevorlang: There maybe something fancy in webpack, but a var wouldnt be super useful as it would need to be different for every level deep. I will look into any improvements we can make on the webpack side. I agree, its pretty ugly.

static/src/js/containers/SidebarContainer/Facets/Facets.scss
trevorlang: Thats what this is :)

static/src/js/containers/SidebarContainer/Facets/FacetsItem.js
trevorlang: You are correct. This is an odd case where it is being passed down as a property on an object, so its not the typical naming convention.

static/src/js/reducers/ui.js
trevorlang: Good catch. Updated.

.eslintrc
trevorlang: 🤦‍♂️ Did not catch that

PR #812, opened 2019-04-23
EDSC-2235: Fix download script when override_temporal is used

macrouch:


app/controllers/granules_controller.rb
```diff
@@ -95,6 +95,11 @@ def fetch_links
     collection = Array.wrap(project['collections']).find { |ds| ds['id'] == collection_id }
 
     query = Rack::Utils.parse_nested_query(collection['params'])
+    puts "query: #{query.inspect}"
```
trevorlang: Did you mean to leave this here?

macrouch: I swear I deleted that!

PR #816, opened 2019-05-01
EDSC-2237: Downtime banner

trevorlang:


mreese84:
We're going to handle this in Google Tag Manager. This isn't needed anymore.

PR #817, opened 2019-05-01
EDSC-2203: Updates Readme

abbottry:


package.json
```diff
@@ -68,7 +69,6 @@
     "resolve-url-loader": "^3.1.0",
     "sass-loader": "^7.1.0",
     "sass-resources-loader": "^2.0.0",
-    "serverless": "^1.41.1",
```
mreese84: I see you pulled this out, but where are we either A) installing this for devs or B) explicitly telling them it is a prerequisite? (a la https://github.com/nasa/earthdata-search#prerequisites)

abbottry: It's in the readme, our readme is no where near as 'detailed' as it was before yet. But its in there. Do you need more detail?

mreese84: It definitely doesn't jump out. I like having a separate prerequisites section that is very clear. So yes, I do think it needs more detail. That's the only thing that was missing from my machine. Remember, our goal with readmes is to make them so that anyone at any level can follow it top to bottom and get up and running.

PR #827, opened 2019-05-19
EDSC-2188: Granule actions

trevorlang:
This PR implements the granule actions. There are some pretty major changes how we're handling information in the redux store. In addition, this creates the granule details page and populates it with the granule details from CMR and CWIC.


static/src/js/components/GranuleDetails/__tests__/GranuleDetailsMetadata.test.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react'
+import Enzyme, { shallow } from 'enzyme'
+import Adapter from 'enzyme-adapter-react-16'
+// import { granuleResultsBodyProps, formattedGranuleInformation } from './data'
```
macrouch: remove

static/src/js/components/GranuleDetails/GranuleDetailsMetadata.js
```diff
@@ -0,0 +1,44 @@
...
+    'echo10',
+    'iso19115'
+  ]
+  return (
+    <div className="granule-details-metadata">
+      <div className="granule-details-metadata__content">
+        <h4 className="granule-details-metadata__heading">Download Metadata:</h4>
+        <ul className="granule-details-metadata__list">
+          {
+            metdataUrlKeys.length && metdataUrlKeys.map((key) => {
```
macrouch: Why not do `Object.keys(metadataUrls)`?

static/src/js/components/GranuleDetails/__tests__/GranuleDetailsMetadata.test.js
trevorlang: Nice catch

static/src/js/components/GranuleDetails/GranuleDetailsMetadata.js
trevorlang: I did it this way so we could order them properly. Using object keys outputs them in a different order than is used currently on prod.

macrouch: Ah, ok. We define that list in a couple of places, or at least pretty close to the same list in `util/granules.js`. Any way to consolidate that code?

static/src/js/components/GranuleDetails/GranuleDetailsInfo.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import './GranuleDetailsInfo.scss'
+
+export const GranuleDetailsInfo = ({ xml }) => {
+  let content = xml
+
+  content = content.replace('<Granule>\n', '') // Remove top level element
```
abbottry: If there aren't tests for this, we should add some and/or ensure that these actually work. These appear to be exact duplicates of regexes ported from ruby and its generally not a safe assumption that regexes across languages are identical.

static/src/js/components/GranuleDetails/GranuleDetailsMetadata.js
```diff
@@ -0,0 +1,44 @@
...
+              const metadataUrl = metadataUrls[key]
+              return (
+                <li
+                  key={`metadata_url_${metadataUrl.title}`}
+                  className="granule-details-metadata__list"
+                >
+                  <a href={metadataUrl.href}>
```
abbottry: This `metadataUrl` may require a token to be accessed. While authentication isn't available yet we need to ensure that it gets documented (likely a new ticket @mreese84 ) to come back and add ensure that it's feasible. This also applies to the single granule downloading.

jest.config.js
```diff
@@ -12,6 +12,7 @@ module.exports = {
     '<rootDir>/test-env.js'
   ],
   testPathIgnorePatterns: [
-    'mocks.js'
+    'mocks.js',
+    'test_data.js'
```
abbottry: We had previously chosen `mocks.js` for this purpose, the data defined in `static/src/js/components/GranuleDetails/__tests__/test_data.js` is defined exactly like the data in `/static/src/js/util/request/__tests__/mocks.js`. I don't care which we choose, but we should stick to an expected filename for mocked data -- we should should remove one of these and rename the files appropriately.

static/src/js/components/GranuleDetails/__tests__/test_data.js
```diff
@@ -0,0 +1,189 @@
...
+}
+
+let formattedGranuleString = '    GranuleUR: 1860_1993_2050_NITROGEN.N-deposition1860.tfw\n'
```
abbottry: `/static/src/js/util/request/__tests__/mocks.js` uses back-ticks and is able to do multiline strings without complaints from the linter -- is this method of concatenating necessary? Also, is the tabbing necessary? This is a brutal variable to maintain.

static/src/js/components/GranuleResults/GranuleResultsItem.js
```diff
@@ -2,32 +2,179 @@
...
+  onExcludeGranule,
+  onFocusedGranuleChange
+}) => {
+  const handleRemoveClick = () => {
+    const {
+      id,
+      is_cwic: isCwic
+    } = granule
+
+    let hashedId
```
abbottry: Seems like this could be cleaned up by just overwriting `id` instead of creating a new variable and then `||` when we use it.

static/src/js/components/GranuleResults/GranuleResultsItem.js
```diff
@@ -2,32 +2,179 @@
...
+    const ftpLinks = links.filter((link) => {
+      const {
+        href,
+        inherited = false,
+        rel
+      } = link
+
+      const filename = href.substr(href.lastIndexOf('/') + 1)
+
+      return href.indexOf('ftp://') !== -1 && rel.indexOf('/data#') !== -1 && inherited !== true && filenames.indexOf(filename) === -1
```
abbottry: This is a lot of code for one line -- can this be a util that we can add comments to for readability?

static/src/js/components/GranuleResults/GranuleResultsItem.js
```diff
@@ -2,32 +2,179 @@
...
       <div className="granule-results-item__body">
-        { thumbnail && (
-          <div className="granule-results-item__thumb">
-            <img src={thumbnail} height="85" width="85" alt={`Browse Image for ${title}`} />
-          </div>
-        )
+        {
+          thumbnail && (
+            <div className="granule-results-item__thumb">
+              <img src={thumbnail} height="85" width="85" alt={`Browse Image for ${title}`} />
```
abbottry: We use these dimensions when we request this data in the collection panel as well, I think it'd be valuable to define these dimensions in a config some place instead of hardcoding them.

static/src/js/components/Timeline/Timeline.js
```diff
@@ -26,14 +26,16 @@ class Timeline extends Component {
...
+      // focusedCollection: oldFocusedCollection,
+      focusedCollectionMetadata: oldFocusedCollectionMetadata,
       temporalSearch: oldTemporalSearch,
       timeline: oldTimeline,
       onChangeTimelineQuery
     } = this.props
 
     const {
-      focusedCollection: nextFocusedCollection,
+      // focusedCollection: nextFocusedCollection,
```
abbottry: Remove?

static/src/js/components/Timeline/Timeline.js
```diff
@@ -26,14 +26,16 @@ class Timeline extends Component {
 
   componentWillReceiveProps(nextProps) {
     const {
-      focusedCollection: oldFocusedCollection,
+      // focusedCollection: oldFocusedCollection,
```
abbottry: Remove?

static/src/js/containers/TimelineContainer/TimelineContainer.js
```diff
@@ -27,9 +30,12 @@ export const TimelineContainer = (props) => {
     onChangeTimelineState
   } = props
 
+  const focusedCollectionMetadata = getFocusedCollectionMetadata(focusedCollection, collections)
+
   return (
     <Timeline
-      focusedCollection={focusedCollection}
+      // focusedCollection={focusedCollection}
```
abbottry: Remove?

static/src/js/reducers/collectionMetadata.js
```diff
@@ -0,0 +1,147 @@
+import {
+  ADD_COLLECTION_GRANULES,
```
abbottry: Linter doesn't know the alphabet ;)

static/src/js/util/url/collectionsEncoders.js
```diff
@@ -0,0 +1,129 @@
...
+ * @param {string} focusedCollection Focused Collection ID
+ * @return {string} An object with encoded Collections
+ */
+export const encodeCollections = (collections, focusedCollection) => {
+  // TODO Encode project Collections
+  // TODO Encode granule filters
+
+  if (!collections) return ''
+  const {
+    // allIds,
```
abbottry: Do we need these two commented variables, or should we just remove them?

static/src/js/components/GranuleDetails/GranuleDetailsInfo.js
trevorlang: They're not exact duplicates. There were some changes that needed to be made. Definitely understand that there may be some inconsistencies between languages. I thought testing the transformed output would be sufficient here. If not, I'll need a little guidance as to how we would like to test them.

static/src/js/components/GranuleDetails/GranuleDetailsMetadata.js
trevorlang: I left a TODO in `static/src/js/util/granules.js` as a reminder and I have added a note in our TODO spreadsheet as well.

jest.config.js
trevorlang: I recall a conversation where we had talked about how this sort of data was different than what had been set in the mocks.js file. Thats the only reason I had set it this way. I may have misunderstood. Totally fine with using _mocks.js_.

static/src/js/components/GranuleDetails/__tests__/test_data.js
trevorlang: So, this was an interesting one. I had done in order to get around the way vscode was trimming trailing whitespace on the items with `: \n` sequences. I was able to modify the the regex to remove that extra space to `:\n` so this is no longer necessary. As for the tabs, I was following the convention currently used on production. I made the assumption that they were there for a reason.

static/src/js/components/GranuleResults/GranuleResultsItem.js
trevorlang: Good call. Went ahead and made that change. @macrouch, just want to make sure you saw you had a chance to call out if it would be an issue for any reason.

static/src/js/components/GranuleResults/GranuleResultsItem.js
trevorlang: @macrouch ill let you figure this one out :)

static/src/js/components/GranuleResults/GranuleResultsItem.js
trevorlang: I'm all for this. Just wondering where we should put it. Any thoughts?

static/src/js/components/GranuleDetails/GranuleDetailsHeader.js
```diff
@@ -0,0 +1,34 @@
+// import React, { Component } from 'react'
```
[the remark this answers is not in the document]
trevorlang: Nice catch

static/src/js/components/GranuleResults/GranuleResultsItem.js
abbottry: I understand this is likely a port, but we should actively improve these types of things as we port them over.

static/src/js/components/GranuleResults/GranuleResultsItem.js
abbottry: We can hold off until we determine a configuration tool. Currently weighing options between using ENV vars and a JSON configuration file.

PR #855, opened 2019-06-23
EDSC-2276: Submit project for download

macrouch:


static/src/js/components/ProjectPanels/AccessMethod.js
```diff
@@ -14,102 +14,155 @@ import './AccessMethod.scss'
...
+  onSelectAccessMethod,
+  onSetActivePanel
+}) => {
+  const {
+    id,
+    tags
+  } = metadata
+
+  const handleAccessMethodSelection = (method) => {
+    console.log('handleAccessMethodSelection', method)
```
abbottry: debug?

static/src/js/components/ProjectPanels/AccessMethod.js
```diff
@@ -14,102 +14,155 @@ import './AccessMethod.scss'
...
+            </Tooltip>
+          )}
+        >
+          <i className="access-method__radio-tooltip fa fa-info-circle" />
+        </OverlayTrigger>
+      </Radio>
+    ))
+  }
+
+  if (Object.keys(tags).indexOf('edsc.extra.serverless.subset_service.echo_orders') !== -1) {
```
abbottry: Within `util/tags.js` we have `getValueForTag` -- we should add an additional util for `hasTag` that does this for us, it will be used in multiple places. 1.0 implements find buy value as well, I'm not sure we'll need that quite yet

```  
  def self.has_tag(collection, tag, value=nil)
    tag = tag_key(tag)
    tags = collection['tags']
    actual = tag_value(collection, tag)
    if tags.is_a?(Array) || !value.nil?
      (value.nil? && actual != nil) || (!value.nil? && actual == value)
    else
      !tags.nil? && !tags[tag].nil?
    end
  end
```

static/src/js/components/ProjectPanels/AccessMethod.js
```diff
@@ -14,102 +14,155 @@ import './AccessMethod.scss'
...
+    radioList.push((
+      <Radio
+        id={`${id}_access-method__stage-for-deliver`}
+        name={`${id}_access-method__stage-for-deliver`}
+        key={`${id}_access-method__stage-for-deliver`}
+        value="Stage For Delivery"
+        onClick={() => handleAccessMethodSelection('echoOrder')}
```
abbottry: What is `echoOrder`? The casing doesn't match the tag postfix and isn't JSON friendly. The other methods seem to use a downcased JSON friendly string here.

static/src/js/components/ProjectPanels/AccessMethod.js
```diff
@@ -14,102 +14,155 @@ import './AccessMethod.scss'
...
+            </Tooltip>
+          )}
+        >
+          <i className="access-method__radio-tooltip fa fa-info-circle" />
+        </OverlayTrigger>
+      </Radio>
+    ))
+  }
+
+  if (Object.keys(tags).indexOf('edsc.extra.serverless.subset_service.opendap') !== -1) {
```
abbottry: `hasTag` method

static/src/js/components/ProjectPanels/AccessMethod.js
```diff
@@ -14,102 +14,155 @@ import './AccessMethod.scss'
...
+            </Tooltip>
+          )}
+        >
+          <i className="access-method__radio-tooltip fa fa-info-circle" />
+        </OverlayTrigger>
+      </Radio>
+    ))
+  }
+
+  if (Object.keys(tags).indexOf('edsc.extra.serverless.subset_service.egi') !== -1) {
```
abbottry: `hasTag` method

static/src/js/components/ProjectPanels/AccessMethod.js
```diff
@@ -14,102 +14,155 @@ import './AccessMethod.scss'
...
+
+  const handleAccessMethodSelection = (method) => {
+    console.log('handleAccessMethodSelection', method)
+    onSelectAccessMethod({
+      collectionId: id,
+      selectedAccessMethod: method
+    })
+  }
+
+  const collectionCapibilities = tags['edsc.extra.serverless.collection_capabilities']
```
abbottry: I think you'll want to use the `getValueForTag` method mentioned in this review -- its within `util/tags.js`

static/src/js/components/ProjectPanels/AccessMethod.js
```diff
@@ -14,102 +14,155 @@ import './AccessMethod.scss'
...
+        </OverlayTrigger>
+      </Radio>
+    ))
+  }
+
+  return (
+    <div className="access-method">
+      <ProjectPanelSection heading="Select Data Access Method">
+        <div className="access-method__radio-list">
+          {/* <RadioList defaultValue="Download"> */}
```
abbottry: Remove?

static/src/js/util/emptyAccessMethod.js
```diff
@@ -0,0 +1,31 @@
+const downloadMethod = {
+  download: {
+    type: 'download'
+  }
+}
+
+// TODO
```
abbottry: Is this meant to stay? TODO should be `TODO: What needs to happen`

static/src/js/util/emptyAccessMethod.js
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Returns an empty access method based on the provided methodName
+ * @param {string} methodName Access method name
+ */
+const emptyAccessMethod = (methodName) => {
+  switch (methodName) {
+    case 'download':
+      return downloadMethod
+    case 'echoOrder':
```
abbottry: It appears as though the code is looking for downcased JSON friendly strings here and you already have checks for `esi` and `opendap` but they aren't represented here. While we don't need them because the functionality isn't there yet they could be here if it's returning an empty object or something -- I think that'd be better than nothing.

static/src/js/util/granules.js
```diff
@@ -98,6 +98,34 @@ export const prepareGranuleParams = (state, projectCollectionId) => {
   }
 }
 
+/**
+ * Translates the values returned from prepareGranuleParams to the camelCased keys that are expected in
+ * the granules.search() function
+ * @param {object} params - Params to be passed to the granules.search() function.
+ * @returns Parameters to be provided to the Granules request with camel cased keys
+ */
+export const buildGranuleSearchParams = (params) => {
```
abbottry: Does this serve a specific piece of functionality that we could rename the method to? `prepareGranuleParams` should be called right before we send a request, this seems to be an extra step that is also stripping out keys that could lead to confusion down the road when we don't know where keys are being stripped from.

static/src/js/util/orders.js
```diff
@@ -0,0 +1,55 @@
...
+  projectIds.forEach((collectionId) => {
+    const projectCollection = byId[collectionId]
+    const { granules } = projectCollection
+
+    const returnValue = {}
+
+    returnValue.id = collectionId
+    returnValue.granule_count = granules.hits
+
+    const params = buildGranuleSearchParams(prepareGranuleParams(state, collectionId))
```
abbottry: This seems to build params with correct casing, change them to javascript casing, then change them back to correct casing for cmr.

macrouch: `prepareGranuleParams` doesn't have them in CMR format, `buildGranuleSearchParams` at least gives them the correct names, then the next two lines clean up the object and snake cases the keys

static/src/js/util/granules.js
macrouch: I just matched what we were doing for collections 🤷‍♂

static/src/js/components/ProjectPanels/AccessMethod.js
macrouch: echoOrder is just the key that I'm storing (or going to store) the access method configs in the redux store, so it is a JS key

PR #858, opened 2019-06-28
Freeze a string literal for `self.table_name`

dainmiller:
Do you mind if we freeze strings like this to reduce object allocation? All good if not! Feel free to close!


abbottry:
Thanks for your input but it's not a necessary change at the moment.

PR #874, opened 2019-07-18
EDSC-2196: Load ESI echo forms into project page and submit

macrouch:


serverless/src/getAccessMethods/getServiceOptionDefinitions.js
```diff
@@ -0,0 +1,55 @@
...
+
+  await serviceOptionDefinitions.forEachAsync(async (serviceOptionDefinition, index) => {
+    const { id: guid } = serviceOptionDefinition
+
+    const url = `${getEarthdataConfig('prod').cmrHost}/legacy-services/rest/service_option_definitions/${guid}.json`
+
+    // Get the access token and clientId to build the Echo-Token header
+    const { clientId, secret } = getSecretEarthdataConfig('prod')
+
+    const verifiedJwtToken = jwt.verify(jwtToken, secret)
```
trevorlang: It looks like we do this exact bit of code to extract the token in quite a few places now. Maybe worth turning it into a util to prevent duplication.

abbottry: Appears that this was taken care of in his other PR 🤷‍♂

PR #875, opened 2019-07-19
EDSC-2256: Implementing order status display for ESI

trevorlang:


static/src/js/components/OrderStatus/__tests__/mocks.js
```diff
@@ -74,3 +82,83 @@ export const orderStatusPropsTwo = {
...
+            dataset_id: 'Test Dataset ID 2'
+          },
+          access_method: {
+            order: {
+              order_status: 'in progress',
+              order_id: [
+                '5000000333461',
+                '5000000333462'
+              ],
+              service_options: {
```
abbottry: This key is no longer necessary, it's an old key from 1.0. I'd rather not go forward with old data -- the best thing here would just be to remove it.

static/src/js/components/OrderStatus/__tests__/mocks.js
```diff
@@ -74,3 +82,83 @@ export const orderStatusPropsTwo = {
   }
 }
 
+export const orderStatusPropsEsi = {
+  authToken: 'testToken2',
+  order: {
```
abbottry: I'm confused with the payload here -- the top level key here is `order` that would suggest the data within in is one order, but the next key is `collections`, which has `esi_orders` inside of it. Based on the data that looks like its a `retrieval_collection` object from the database. `access_method` inside of that object contains the order data, but that data is not stored in the database, and the plan is to leave that unchanged. If we need to discuss this I can make some time this afternoon.

static/src/js/components/OrderStatus/mocks.js
```diff
@@ -234,4 +236,283 @@ export const echoOrder = {
...
+          }
+        ]
+      },
+      edsc_extra_subset_service_esi: {
+        data: [
+          'S1568897222-LPDAAC_ECS'
+        ]
+      },
+      org_ceos_wgiss_cwic_granules_prod: {},
+      'edsc.extra.serverless.subset_service.esi': {
```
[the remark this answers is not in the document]
trevorlang: I think the quotes are required here because of the periods in the keys 🤔

static/src/js/components/OrderStatus/__tests__/mocks.js
trevorlang: I've basically mimicked what I saw in 1.0 for this ticket. I was under the impression any improvements to the payloads would be happening when we hook this up to real order responses. I can try and do some of that now, but I think we should probably chat about what we can make better first.

static/src/js/components/OrderStatus/__tests__/mocks.js
trevorlang: This will be handled under the order status work

static/src/js/components/OrderStatus/__tests__/mocks.js
trevorlang: This will be handled under the order status work

PR #878, opened 2019-07-26
EDSC-2308: Submit shapefiles with supported echo forms

macrouch:


serverless/src/submitCatalogRestOrder/handler.js
```diff
@@ -87,6 +92,20 @@ const submitCatalogRestOrder = async (event, context) => {
 
     const { model, url } = accessMethod
 
+    // Retrieve the shapefile if one was provided
+    const { source } = JSON.parse(jsondata)
```
abbottry: Is there a point in time where we have this information that we can dump it into `jsondata` separately from the `source`?  Pretty inefficient to be parsing query strings to retrieve this information.

PR #884, opened 2019-08-02
EDSC-2282: Implements lambda warmers

abbottry:


serverless/src/storeUserData/handler.js
```diff
@@ -35,15 +39,13 @@ export const storeUserData = async (event) => {
 
   const existingUser = await dbConnection('users').select('id').where({ urs_id: username })
 
-  let queryResponse
-
   if (existingUser.length) {
-    queryResponse = await dbConnection('users').returning(['id', 'urs_id']).update({ ...userPayload }).where({ urs_id: username })
+    await dbConnection('users').returning(['id', 'urs_id']).update({ ...userPayload }).where({ urs_id: username })
```
macrouch: If you don’t need to return queryResponse, do you need .returning()?

abbottry: nope. I'll clean that up.

PR #890, opened 2019-08-07
EDSC-2264: Implements Giovanni Smart Handoffs

abbottry:


static/src/js/components/CollectionDetails/CollectionDetailsHeader.js
```diff
@@ -32,31 +39,45 @@ export const CollectionDetailsHeader = ({ focusedCollectionMetadata }) => {
...
-              <Dropdown.Item
-                className="collection-details-header__more-actions-item collection-details-header__more-actions-vis"
-              >
-                Giovanni Links
-              </Dropdown.Item>
-            </Dropdown.Menu>
-          </Dropdown>
-        </div>
+
+        {
```
macrouch: Is this the same block of code that is in `GranuleResultsHeader.js`? we should pull it out into a reusable component so we don't duplicate code

trevorlang: I think I get where you’re coming from, but I think I am fine with them being duplicated in this case. Seems to me that it’s totally possible we could have different sets of actions based on where the drop down is placed. 

I could see possibly trying to abstract the Dropdown, Dropdown.Toggle and Dropdown.Menu into a single component, whose children could be the Dropdown.Header and Dropdown.Item’s. But that seems like overkill to me when it’s not affording is a ton of benefit.

I’d vote for leaving it the way it is I think

trevorlang: I think I get where you’re coming from, but I think I am fine with them being duplicated in this case. Seems to me that it’s totally possible we could have different sets of actions based on where the drop down is placed. 

I could see possibly trying to abstract the Dropdown, Dropdown.Toggle and Dropdown.Menu into a single component, whose children could be the Dropdown.Header and Dropdown.Item’s. But that seems like overkill to me when it’s not affording is a ton of benefit.

I’d vote for leaving it the way it is I think

abbottry: I went the way of components but wasn't a fan of the `Dropdown.Item` being in the render method. While it'd be great to make the handoffs it's own component that we could just render all of the links in, we don't restrict the dropdown to hand off links so we can't really do that. I see the value, but I ultimately landed here.

PR #899, opened 2019-08-22
EDSC-2338: Remove NLP

macrouch:


abbottry:
Do we want to remove the url config in `static.config.json`?

macrouch:
> Do we want to remove the url config in `static.config.json`?

:doh: I did remove it, I just didn't commit that file because my other changes to edscHost

PR #904, opened 2019-08-31
EDSC-2143: Enable collection sorting and checkboxes

macrouch:


static/src/js/components/CollectionResults/CollectionResultsHeader.js
```diff
@@ -4,8 +4,51 @@ import { Col, Form, Row } from 'react-bootstrap'
 
 import './CollectionResultsHeader.scss'
 
-const CollectionResultsHeader = ({ portal }) => {
+const eosdisTagKey = 'gov.nasa.eosdis'
```
abbottry: Probably want to move this into the application config in static config -- or somewhere not hard coded in a file anyway.

PR #908, opened 2019-09-03
EDSC-2300: Adds download history page

abbottry:


static/src/js/util/request/retrievalRequest.js
```diff
@@ -16,6 +16,10 @@ export default class RetrievalRequest extends Request {
     ]
   }
 
+  all() {
```
macrouch: This file doesn't have tests

static/src/js/reducers/retrievalHistory.js
```diff
@@ -0,0 +1,17 @@
+import { SET_RETRIEVAL_HISTORY } from '../constants/actionTypes'
+
+const initialState = []
+
+const retrievalHistoryReducer = (state = initialState, action) => {
+  switch (action.type) {
+    case SET_RETRIEVAL_HISTORY: {
```
macrouch: write a test

static/src/js/containers/DownloadHistoryContainer/DownloadHistoryContainer.js
```diff
@@ -0,0 +1,126 @@
...
+    // No collection title exists so we'll just display how many collections were downloaded
+    return `${collections.length} ${pluralize('collection', collections.length)}`
+  }
+
+  render() {
+    const { retrievalHistory } = this.props
+
+    return (
+      <>
+        <h2 className="route-wrapper__page-heading">Download Status & History</h2>
```
macrouch: This normally goes in a component, not a container. It is a simple page, but should we follow the way the rest of the application is laid out?

static/src/js/containers/DownloadHistoryContainer/DownloadHistoryContainer.js
```diff
@@ -0,0 +1,50 @@
+import React, { Component } from 'react'
+import { connect } from 'react-redux'
+import PropTypes from 'prop-types'
+import { withRouter } from 'react-router-dom'
+
+import actions from '../../actions'
+import { DownloadHistory } from '../../components/DownloadHistory/DownloadHistory'
+
+const mapStateToProps = state => ({
+  authToken: state.authToken,
```
macrouch: are authToken and portal used? I think they can be removed

static/src/js/containers/DownloadHistoryContainer/DownloadHistoryContainer.js
trevorlang: This is sort of what I was referring to when we need a way to wrap these routes and their subcomponents consistently, probably through a separate component. I think we can probably leave it the way it is until we sort that out 🤷‍♂

abbottry: Eh, I moved it.

PR #910, opened 2019-09-05
Add Bamboo deployment script

jcromartie:


bin/deploy-bamboo.sh
```diff
@@ -0,0 +1,67 @@
...
+}
+
+# Execute serverless commands in Docker
+#######################################
+
+# Deploy AWS Resources
+dockerEnvRun serverless deploy --stage $bamboo_STAGE_NAME
+
+# Migrate the database
+dockerEnvRun serverless invoke --function migrateDatabase --stage sit
```
jonvel: Should this (and two lines lower) hardcode `--stage sit` or use `--stage $bamboo_STAGE_NAME` ?

PR #919, opened 2019-09-10
EDSC-2352: Modifies token management

abbottry:


serverless/src/util/urs/getUserAccessToken.js
```diff
@@ -0,0 +1,25 @@
...
+  const existingUserTokens = await dbConnection('user_tokens')
+    .first([
+      'access_token',
+      'refresh_token',
+      'expires_at'
+    ])
+    .where({ user_id: id })
+    .orderBy('created_at', 'DESC')
+
+  console.log('existingUserTokens', existingUserTokens)
```
abbottry: Should remove this

serverless/src/edlAuthorizer/handler.js
```diff
@@ -48,36 +52,92 @@ const edlAuthorizer = async (event) => {
...
+        console.log(`JWT Token Invalid. ${verifyError}`)
+
         throw new Error('Unauthorized')
       }
 
+      const {
+        id: userId,
+        username
+      } = decodedJwtToken
+      console.log('decodedJwtToken', decodedJwtToken)
```
abbottry: Remove this

serverless/src/util/urs/getAccessTokenFromJwtToken.js
```diff
@@ -0,0 +1,26 @@
+import { getDbConnection } from '../database/getDbConnection'
+import { getVerifiedJwtToken } from '../getVerifiedJwtToken'
+
+// Knex database connection object
+let dbConnection = null
```
abbottry: I'm not convinced this provides us any value, does this value get kept in memory with this scope or are we just overwriting a potentially 'cached' value each time we call this method?

serverless/src/util/urs/getUserAccessToken.js
```diff
@@ -0,0 +1,25 @@
+import { getDbConnection } from '../database/getDbConnection'
+
+// Knex database connection object
+let dbConnection = null
```
abbottry: I'm not convinced this provides us any value, does this value get kept in memory with this scope or are we just overwriting a potentially 'cached' value each time we call this method?

serverless/src/util/urs/getAccessTokenFromJwtToken.js
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Returns the decrypted urs system credentials from Secrets Manager
+ */
+export const getAccessTokenFromJwtToken = async (jwtToken) => {
+  // Retrieve a connection to the database
+  dbConnection = await getDbConnection(dbConnection)
+
+  const { id } = getVerifiedJwtToken(jwtToken)
+
+  const existingUserTokens = await dbConnection('user_tokens')
```
abbottry: Consider returning `user_id` for use in methods that require both a user id and token from the jwtToken.

static/src/js/util/request/logoutRequest.js
```diff
@@ -0,0 +1,18 @@
...
+export default class LogoutRequest extends Request {
+  constructor(authToken) {
+    super(getEnvironmentConfig().apiHost)
+
+    this.authenticated = true
+    this.authToken = authToken
+  }
+
+  logout() {
+    return super.delete('/logout')
```
abbottry: We've used `this.` for all of our other request classes, is there a difference in this call?

macrouch: the one I copied from uses super 🤷‍♂ probably change this one and conceptRequest.js

PR #928, opened 2019-09-13
EDSC-2329: Implement output format for OPeNDAP retrievals

macrouch:


abbottry:
Would you **prefer** to use tag data for this or making a request to get the data?

macrouch:
> Would you **prefer** to use tag data for this or making a request to get the data?

We are using tags to make up for how slow legacy services is, retrieving the UMM-S record is CMR and should be fast so I think we shouldn't use tags

PR #933, opened 2019-09-17
EDSC-2360: Error pages

macrouch:


serverless/src/errorLogger/handler.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Logs an error reported by a client
+ * @param {Object} event
+ */
+const errorLogger = async (event) => {
+  const { body } = event
+  const { params = {} } = JSON.parse(body)
+
+  console.error('Error reported', params)
```
abbottry: I think we should break this up into a message with a known format so that we can setup alerts in Splunk. I'd suggest something like `[uid] (code) message` or something.

static/src/js/components/Errors/NotFound.js
```diff
@@ -0,0 +1,53 @@
...
+  // eslint-disable-next-line global-require
+  require('./ErrorBoundary.scss')
+
+  const guid = uuidv4()
+
+  const requestObject = new LoggerRequest()
+  requestObject.log({
+    error: {
+      guid,
+      message: '404 error',
```
abbottry: I see location here, and stack some place else. Is there no expected payload here? We should try and keep things consistent (even if its by type of error)

abbottry: Do we know what wasn't found? API gateway returns an appropriate message that we should display if possible. Also, I prefer `404 Not Found` 🤷‍♂

serverless/src/errorLogger/handler.js
abbottry: I'm assuming that the message here is the one that included the action, so that will be an additional thing we can setup in our alerts/reports which would be awesome.

static/src/js/components/Errors/ErrorBoundary.js
```diff
@@ -0,0 +1,81 @@
...
+    }
+  }
+
+  render() {
+    const { errorGuid, hasError } = this.state
+    const { children } = this.props
+
+    if (hasError) {
+      // eslint-disable-next-line global-require
+      require('./ErrorBoundary.scss')
```
abbottry: Should we just have sub components?

PR #934, opened 2019-09-17
EDSC-2334: Fixes project collection metadata on project page

trevorlang:


static/src/js/components/ProjectCollections/ProjectCollectionsItem.js
```diff
@@ -107,16 +114,22 @@ const ProjectCollectionItem = ({
...
-        >
-          {`${abbreviate(granuleCount, 1)} Granules`}
-        </li>
-        <li
-          className="project-collections-item__stats-item project-collections-item__stats-item--total-size"
-        >
-          {`Est. Size ${size} ${unit}`}
-        </li>
+        {
+          (typeof granuleCount !== 'undefined' && size && unit) && (
```
abbottry: I'm not necessarily asking for a change here, but I think we need a util for this kind of thing, or maybe lodash has something? checking `typeof` seems undesirable.

abbottry: does `''` evaluate? I see size and unit default to empty strings above. Do we have an `isLoading` for this to check on something instead of checking these three variables? It seems fragile 🤷‍♂

trevorlang: `''` will evaluate to false. I can look into a replacement for the `undefined` check. We dont currently have loading states in place, but we have a ticket coming up to implement loading states. I plan to do exactly that in that ticket.

PR #954, opened 2019-09-25
EDSC-2397: Disable download and project buttons for data migration

macrouch:


mreese84:
This is the final pull request for our knockout codebase. This temporarily disables the download buttons when the proper environment variable is configured so that we can move forward with a database migration. This release will only be deployed for a few hours when our serverless app goes live.

PR #957, opened 2019-09-26
EDSC-2393: Add relevancy ranking metrics

macrouch:


static/src/js/components/Map/GranuleGridLayer.js
```diff
@@ -865,13 +865,13 @@ export class GranuleGridLayer extends MapLayer {
 
     // Create a GranuleGridLayerExtended layer from each data object in getLayerData
     const layerData = this.getLayerData(props)
-    layerData.forEach((data, index) => {
+    Object.keys(layerData).forEach((id, index) => {
```
trevorlang: Looks like we could set `Object.keys(layerData)` to a variable here so we don't have to call it quite so many times

PR #959, opened 2019-09-30
EDSC-2385: Add logo header to downloads and projects pages

macrouch:


static/src/js/routes/Downloads/Downloads.js
```diff
@@ -15,9 +15,12 @@ import OrderStatusContainer from '../../containers/OrderStatusContainer/OrderSta
 import CollectionContainer from '../../containers/CollectionContainer/CollectionContainer'
 import PortalLinkContainer from '../../containers/PortalLinkContainer/PortalLinkContainer'
 import DownloadHistoryContainer from '../../containers/DownloadHistoryContainer/DownloadHistoryContainer'
+import { getApplicationConfig } from '../../../../../sharedUtils/config'
+import Header from '../../components/Sidebar/SidebarHeader'
```
trevorlang: I think the best thing to do here would be to have a connected `AppLogo` container that was imported here, and by `SidebarHeader` rather than reusing a sidebar specific component here.

PR #964, opened 2019-10-03
EDSC-2199: Contact info page

macrouch:


static/src/js/actions/contactInfo.js
```diff
@@ -0,0 +1,75 @@
+import ContactInfoRequest from '../util/request/contactInfoRequest'
+import { UPDATE_CONTACT_INFO } from '../constants/actionTypes'
+
+export const updateContactInfo = data => ({
+  type: UPDATE_CONTACT_INFO,
+  payload: data
+})
+
+export const fetchContactInfo = () => (dispatch, getState) => {
```
trevorlang: These functions could probably use some quick comments.

static/src/js/components/ContactInfo/ContactInfo.scss
```diff
@@ -0,0 +1,25 @@
+fieldset.account-form {
```
trevorlang: I would like to see this CSS and the component `classNames` BEMified like our other components. I can help with this.

static/src/js/components/Button/Button.js
```diff
@@ -82,6 +83,7 @@ export const Button = ({
       aria-label={label}
       type={type}
       disabled={disabled || spinner}
+      target={target}
```
trevorlang: Looks like `react-bootstrap` doesn't do anything to account for the issues with `target="_blank"` I think we can just add the `rel="noopener nofollow"` when `target === "_blank"` or something similar.

https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

static/src/js/components/ContactInfo/ContactInfo.scss
```diff
@@ -1,4 +1,4 @@
-fieldset.account-form {
+.contact-info-form {
```
trevorlang: This looks much better. Only change I would make would be getting rid of these nested element selectors. They are creating unnecessary specificity.

```
.contact-info-form {
  ...
  > ul {
    ...
    > li {
       ...
    }
  }
}
```
should become something like
```
.contact-info-form {
  ...
  &__list {
    ...
  }
  &__item {
      ...
  }
}
```

PR #995, opened 2019-11-03
EDSC-2469: Manually enter point and bounding box spatial

macrouch:


static/src/js/components/SpatialDisplay/SpatialDisplay.js
```diff
@@ -76,21 +78,37 @@ class SpatialDisplay extends Component {
       shapefile
     } = this.props
 
+    this.setState({ error: '' })
```
trevorlang: It seems like there should be a better way to make sure error is getting reset. IIRC, these `setState` calls stack async which I feel could lead to some unpredictable behavior. Probably worth looking into a little closer.

static/src/js/components/SpatialDisplay/SpatialDisplay.js
```diff
@@ -281,14 +446,7 @@ class SpatialDisplay extends Component {
...
       ))
-    } else if ((boundingBoxSearch && !drawingNewLayer) || drawingNewLayer === 'rectangle') {
-      // Arrange the points in the right order
-      const points = boundingBoxSearch
-        ? boundingBoxSearch
-          .match(/[^,]+,[^,]+/g)
-          .map(pointStr => trimSpatial(pointStr.split(','), 5).reverse().join(', '))
-        : [undefined, undefined]
-
+    } else if ((boundingBoxSearch && (boundingBoxSearch[0] || boundingBoxSearch[1]) && !drawingNewLayer) || drawingNewLayer === 'rectangle' || manuallyEntering) {
```
trevorlang: These `if` statements are getting pretty crazy. Wonder if theres a way they could be shortened up.

abbottry: Probably true. Can we do that with the advanced search as that will add yet another layer to this.

PR #999, opened 2019-11-11
EDSC-2434: Schedules collection capability state machine

abbottry:


serverless-configs/aws-step-functions.yml
```diff
@@ -3,6 +3,8 @@ stateMachines:
     name: UpdateOrderStatusWorkflow-${self:provider.stage}
     role: 
       Fn::ImportValue: ${self:provider.stage}-ApplicationRole
+    events:
+          rate: cron(0 5 ? * MON-FRI *)
```
abbottry: ensure tabbing is correct.

PR #1016, opened 2019-12-04
EDSC-2487: Adds Orbital Crossing and Number fields to Granule Filters

abbottry:


static/src/js/components/GranuleFilters/GranuleFiltersForm.js
```diff
@@ -414,7 +414,7 @@ export const GranuleFiltersForm = (props) => {
                         </GranuleFiltersItem>
 
                         <GranuleFiltersItem
-                          heading="Equatorial Crossing Date"
+                          heading="Equator Crossing Date"
```
trevorlang: Any reason for this one? Seems like "Equatorial" is probably the correct term.

PR #1022, opened 2019-12-13
EDSC-2409: Implementing HUC Advanced Search

trevorlang:


static/src/js/components/AdvancedSearchModal/RegionSearchResults.js
```diff
@@ -0,0 +1,187 @@
...
+                          <Button
+                            key={id}
+                            className="region-search-results__item"
+                            variant="naked"
+                            onClick={() => this.onSetSelected(result)}
+                            as="li"
+                            label={id}
+                          >
+                            <span className="region-search-results__selected-region-id">
+                              {`${type.toUpperCase()} ${id}`}
```
abbottry: Seems like another good place for css

static/src/js/components/Button/Button.js
```diff
@@ -77,9 +77,14 @@ export const Button = ({
     rel = 'noopener nofollow'
   }
 
+  let asEl
```
abbottry: Seems like this would always be undefined?

static/src/js/components/AdvancedSearchModal/RegionSearchForm.js
```diff
@@ -0,0 +1,250 @@
...
+                  )
+                }
+              </Col>
+            </Row>
+          )
+        }
+          {
+          !isEmpty(selectedRegion) && (
+            <p className="region-search__selected-region">
+              <span className="region-search__selected-region-id">{`${selectedRegion.type.toUpperCase()} ${selectedRegion.id}`}</span>
```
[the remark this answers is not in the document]
trevorlang: I think since its an acronym, it should be capitalized in the markup. We aren't storing a capitalized version anywhere, so this seems to do the trick.

static/src/js/components/AdvancedSearchDisplay/AdvancedSearchDisplay.js
```diff
@@ -0,0 +1,68 @@
...
+      advancedSearch,
+      onUpdateAdvancedSearch,
+      onChangeQuery
+    } = this.props
+
+    const advancedSearchFiltersApplied = Object.values(advancedSearch).filter((value) => {
+      if (isObject(value)) {
+        return !isEmpty(value)
+      }
+      return !!value
```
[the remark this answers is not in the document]
trevorlang: My thinking with this function is it will return true if any properties have a truthy values

static/src/js/util/url/advancedSearchEncoders.js
```diff
@@ -0,0 +1,42 @@
...
+  return {
+    sr: {
+      ...selectedRegion
+    }
+  }
+}
+
+/**
+ * Decodes a parameter object into an advancedSearch object
+ * @param {*} params URL parameter object from parsing the URL parameter string
```
macrouch: `*` should be `Object`

static/src/js/components/Button/Button.js
trevorlang: The bootstrap button has some built in logic to support hrefs on buttons. This basically says, if `as` is set to `button` and an `href`, let's use bootstraps logic instead

PR #1024, opened 2019-12-17
EDSC-2506: Filtering granule results with region search

trevorlang:


static/src/js/util/collections.js
```diff
@@ -4,10 +4,38 @@ import { encodeTemporal } from './url/temporalEncoders'
...
 import { tagName } from '../../../../sharedUtils/tags'
 
+/**
+ * Takes the current CMR collection params and applies any changes needed to account
+ * for the current advanced search state.
+ * @param {Object} collectionParams The current collection search params.
+ * @param {Object} advancedSearch The current advanced search state params.
+ * @returns {Object} Parameters used in prepareCollectionParams.
+ */
+export const withAdvancedSearch = (collectionParams, advancedSearch) => {
```
abbottry: Can we just have one util? They seem to do the exact same thing

trevorlang: I went back and forth on that a bit. Ultimately I thought this was best since we can’t guarantee that we would always send the same params for the collections and granules requests. At that point, it just seemed to be over complicating things.

PR #1025, opened 2019-12-18
EDSC-2505: Cancel pending requests

macrouch:


static/src/js/util/request/request.js
```diff
@@ -24,6 +24,23 @@ export default class Request {
...
     this.startTime = null
+    this.source = CancelToken.source()
+  }
+
+  /**
+   * Defines the default keys that our API endpoints allow.
+   * @return {Array} An empty array
+   */
+  cancel() {
+    console.warn('cancelling the request')
```
abbottry: Probably want to remove this?

PR #1035, opened 2020-01-16
EDSC-2518: Write integration test for downloading a collection

macrouch:


cypress/integration/project_page/download_spec.js
```diff
@@ -0,0 +1,104 @@
...
+
+    cy.login()
+
+    cy.route({
+      method: 'POST',
+      url: '**/collections/json',
+      response: collectionsJson.body,
+      headers: authHeader
+    })
+
```
abbottry: Just trying to trigger my ocd? :)

static/src/js/components/SecondaryToolbar/SecondaryToolbar.js
```diff
@@ -106,16 +107,18 @@ class SecondaryToolbar extends Component {
...
         className="collection-results__item-title-link"
         to={{
           pathname: '/search',
-          search: stringify({
-            ...params,
-            p
-          })
+          search: newSearch
         }}
+        onClick={() => { onChangePath(`/search/${newSearch}`) }}
```
abbottry: Minor, but we cleaned up the trailing slash everywhere else, should do the same here.

PR #1036, opened 2020-01-17
EDSC-2351: Adds ordering admin dashboard

abbottry:


static/src/js/containers/AdminRetrievalContainer/AdminRetrievalContainer.js
```diff
@@ -0,0 +1,61 @@
+import React, { Component } from 'react'
```
macrouch: I think we should add a test for this container

PR #1046, opened 2020-02-04
EDSC-2547: Display native granule format on collections details page

macrouch:


static/src/js/components/CollectionDetails/CollectionDetailsBody.js
```diff
@@ -57,6 +58,16 @@ const buildScienceKeywordList = (scienceKeywords) => {
   return scienceKeywordList
 }
 
+const buildNativeFormatList = (nativeFormats) => {
+  if (!nativeFormats.length) return []
```
abbottry: Shouldn't this return a React fragment or null? One condition returns markup, the other returns a javascript array.

static/src/js/util/collectionMetadata/nativeFormat.js
```diff
@@ -0,0 +1,19 @@
+import { uniq } from 'lodash'
+
+/**
+ * Extracts the native format(s) from the UMM-C metadata
+ * @param {Object} ummJson UMM-C JSON metadata
+ */
+export const buildNativeFormat = (ummJson) => {
+  const { ArchiveAndDistributionInformation = {} } = ummJson
+  const { FileDistributionInformation = [] } = ArchiveAndDistributionInformation
```
abbottry: Does this survive the linter? I thought it'd require the first letter to be lower case?

static/src/js/util/collectionMetadata/nativeFormat.js
```diff
@@ -0,0 +1,19 @@
...
+
+  const formats = []
+
+  FileDistributionInformation.forEach((info) => {
+    const { Format, FormatType } = info
+    if (FormatType.toLowerCase() === 'native' && Format.toLowerCase() !== 'not provided') formats.push(Format)
```
abbottry: Can this please be three lines, this is much less readable (IMO)

static/src/js/components/CollectionDetails/CollectionDetailsBody.js
```diff
@@ -11,6 +11,7 @@ import CollectionDetailsMinimap from './CollectionDetailsMinimap'
 import SplitBadge from '../SplitBadge/SplitBadge'
 
 import './CollectionDetailsBody.scss'
+import pluralize from '../../util/pluralize'
```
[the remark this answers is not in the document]
trevorlang: +1. I think we should be using named imports everywhere except for the react components

static/src/js/components/CollectionDetails/CollectionDetailsBody.js
macrouch: I just did the same thing that buildRelatedUrlsList and buildScienceKeywordList are doing

abbottry: ok -- well then those are wrong too! I think the 'issue' is that we always have a conditional that wraps the render when these are used so it never actually attempts to render the array. @trevorlang any idea how to best handle this? I know it hasn't caused issues (that I'm aware of) but definitely seems like a quick and easy improvement.

PR #1069, opened 2020-03-05
EDSC-2517: Copy granule results into project with My Project button

macrouch:


macrouch:
Closing after deciding to fix using another approach

PR #1070, opened 2020-03-06
EDSC-2548: Collection details styling

trevorlang:


static/src/js/components/CollectionDetails/CollectionDetailsBody.js
```diff
@@ -132,124 +133,132 @@ export const CollectionDetailsBody = ({
...
+                            >
+                              View All Related URLs
+                            </Button>
+                          </>
+                        )
+                      }
+                      <a
+                        className="link link--external collection-details-body__related-link"
+                        href={urls.html.href}
+                        // eslint-disable-next-line react/jsx-no-target-blank
```
abbottry: I think this can be avoided by using `target="_blank" rel="noopener noreferrer"`

static/src/js/components/Panels/PanelItem.js
```diff
@@ -113,6 +114,43 @@ export class PanelItem extends Component {
...
     ])
 
+    const renderChildren = () => {
+      const contents = (
+        <>
+          {
+            typeof children === 'string' && children
+          }
+          {
+            typeof children !== 'string' && (
```
abbottry: If not a string, what is this? What is the use case here for a string versus.. anything else?

static/src/js/components/CollectionDetails/CollectionDetailsBody.js
trevorlang: Good catch

static/src/js/components/Panels/PanelItem.js
trevorlang: If the child is not a string, it's a react component. In the case it's a string, we just want to render the string. If its a component, we use cloneElement to set some props on the children that get passed in. This is really just a safety net so we don't attempt to set props on a string, which would not work.

PR #1078, opened 2020-03-13
EDSC-2557: Adds spatial searching to autocomplete lambda

abbottry:


serverless/src/autocomplete/handler.js
```diff
@@ -1,17 +1,149 @@
...
+ * @param {query} query The spatial query provided from the user
+ */
+const googleGeocode = async (query) => {
+  const client = new Client({})
+
+  const geocodeResult = await client
+    .geocode({
+      params: {
+        address: query,
+        key: process.env.GOOGLE_MAPS_API_KEY
```
abbottry: Should be secure.

PR #1081, opened 2020-03-18
EDSC-2555: Adds autocomplete to GUI

macrouch:


static/src/js/util/collections.js
```diff
@@ -121,6 +122,12 @@ export const prepareCollectionParams = (state) => {
     ...portalQuery
   }
 
+  const { selected = [] } = autocomplete
+  selected.forEach((param) => {
+    const { type, value } = param
+    collectionParams[type] = value
```
macrouch: this needs to handle multiple types if the field supports it

static/src/js/actions/autocomplete.js
```diff
@@ -0,0 +1,163 @@
...
+
+  const mappedType = autocompleteFacetsMap[type]
+
+  if (!mappedType) return null
+
+  const facets = {
+    [mappedType]: value
+  }
+
+  // TODO I think this code works, but waiting for CMR-xxxx
```
abbottry:
```suggestion
  // TODO I think this code works, but waiting for CMR-6287
```

static/src/js/components/AutocompleteDisplay/AutocompleteDisplay.js
```diff
@@ -0,0 +1,49 @@
...
+
+import { Button } from '../Button/Button'
+
+import './AutocompleteDisplay.scss'
+
+const AutocompleteDisplay = ({ selected, onRemoveAutocompleteValue }) => {
+  if (!selected.length) return null
+
+  return (
+    <li className="filter-stack-item">
```
abbottry: Wanted to note for discussion, I remember @trevorlang asking about this using `filter-stack-item` -- is this actually a filter stack item or does it just resemble it from an aesthetic standpoint? May need it's own class / styles.

static/src/js/components/Facets/Facets.js
```diff
@@ -58,39 +58,46 @@ const Facets = (props) => {
   const keywordsFacet = {
     ...cmrFacetDefaults,
     title: 'Keywords',
+    autocompleteType: 'science-keywords',
```
abbottry: I feel like this is generally an underscore, I know CMR 'offers' both but we currently use `science_keywords_h` for facets, is this our own internal key or something were asking from CMR? Maybe should be included in [CMR-6287](https://bugs.earthdata.nasa.gov/browse/CMR-6287)?

EDIT: Looks like the other autocompleteTypes below are using camel case -- we should match those.

static/src/js/util/autocompleteFacetsMap.js
```diff
@@ -0,0 +1,12 @@
...
+ * Maps the CMR Autocomplete types to CMR Facet types
+ */
+export const autocompleteFacetsMap = {
+  dataCenter: 'data_center_h',
+  instrument: 'instrument_h',
+  granuleDataFormat: 'granule_data_format_h',
+  platform: 'platform_h',
+  processingLevelId: 'processing_level_id_h',
+  project: 'project_h',
+  'science-keywords': 'science_keywords_h'
```
abbottry: Mentioned above, can we make this `scienceKeywords`? If it requires a change from CMR please write an additional ticket for us and link it to [CMR-6287](https://bugs.earthdata.nasa.gov/browse/CMR-6287).

static/src/js/util/collections.js
```diff
@@ -121,6 +123,20 @@ export const prepareCollectionParams = (state) => {
     ...portalQuery
   }
 
+  // Add the autocomplete selected parameters if the type is not a CMR Facet
+  const { selected = [] } = autocomplete
+  selected.forEach((param) => {
+    const { type, value } = param
+
+    if (!autocompleteFacetsMap[type]) {
```
abbottry: Do you think it's necessary to restrict the non facet param list?

PR #1086, opened 2020-03-26
EDSC-2615: Adds ability to reset echoform

macrouch:


static/src/js/components/ProjectPanels/ProjectPanels.js
```diff
@@ -267,6 +282,22 @@ class ProjectPanels extends PureComponent {
           <span className="project-panels__collection-count">
             {`Collection ${index + 1} of ${projectIds.length}`}
           </span>
+          {
+            selectedAccessMethod
+            && (
+              selectedAccessMethod.includes('esi')
```
abbottry: Given that we generate these values, I'd rather compared to the full value of the `type` key within the selected access method -- I understand its more compute and more work on your end but I think its the correct way of doing it.

PR #1092, opened 2020-04-06
EDSC-2588: EDSC-2640: Adds user preferences page

macrouch:


static/src/js/components/Preferences/RadioField.js
```diff
@@ -0,0 +1,87 @@
...
+    } = this.props
+
+    const {
+      enum: values,
+      enumNames,
+      description
+    } = schema
+
+    return (
+      <div className="preferences-form__radio">
```
trevorlang, line 42: For consistency, I think we should use `react-bootstrap`s grid and form components here

PR #1121, opened 2020-06-04
CMRSCI-3626: initial version of AMD Portal adding config and image files.

mmorahan:
CMRSCI-3626: initial version of AMD Portal adding config and image files.


portals/amd/config.json
```diff
@@ -0,0 +1,17 @@
...
+  "hideCollectionFilters": true,
+  "logo": {
+    "id": "amd_logo",
+    "link": "https://www.scar.org/data-products/antarctic-master-directory/",   
+    "title": "Antarctic Master Directory"
+  },
+  "org": "AMD",
+  "pageTitle": "Antarctic Master Directory",   
+  "query": {
+    "tagKeg": "org.scar.amd"
```
macrouch, line 13: this should be `tagKey`, and this line has to end in a comma to be valid JSON

eereiter: I have fixed the issue.

PR #1134, opened 2020-06-17
EDSC-2737: Track granule sort key in the url

abbottry:


macrouch:
After sorting granules and returning to the page, the sort key select is not displaying the correct sort key

PR #1138, opened 2020-06-18
EDSC-2718: Improves map display of additive granule state

trevorlang:


static/src/js/actions/granules.js
```diff
@@ -355,7 +361,33 @@ export const getGranules = ids => (dispatch, getState) => {
...
+        addedGranuleIds = []
+      } = projectCollection
+
+      const { metadata: newMetadata = {} } = getState()
+      const { collections } = newMetadata
+      const { byId: collectionsById } = collections
+      const { granules: collectionGranules = {} } = collectionsById
+      const { allIds: collectionGranulesAllIds } = collectionGranules
+
+      if (addedGranuleIds.length && addedGranuleIds.length < 2000) {
```
abbottry, line 381: Would it make sense to use our static config here? `defaultGranulesPerOrder`

PR #1159, opened 2020-07-16
EDSC-2806: Removes EDSC specific values from config

macrouch:


sharedUtils/getClientId.js
```diff
@@ -0,0 +1,33 @@
...
+    .replace('ENV', envOverride || getApplicationConfig().env)
+    .replace('PORTAL', getApplicationConfig().defaultPortal)
+)
+
+/**
+ * Returns the clientId to send with CMR requests
+ */
+export const getClientId = () => {
+  // Override the Env with 'test' if we are in CI
+  const { ciMode, clientId } = getApplicationConfig()
```
abbottry, line 18: I'd like to eventually figure out a way to not need this `ciMode` config -- we should be able to derive this or use another config.

PR #1166, opened 2020-07-28
EDSC-2843: Reset Form button repopulates model and rawModel after cle…

macrouch:
…aring them


static/src/js/components/AccessMethod/EchoForm.js
```diff
@@ -13,23 +13,44 @@ export const EchoForm = ({
   spatial,
   onUpdateAccessMethod
 }) => {
-  const [rawModel, setRawModel] = useState('')
-  const [model, setModel] = useState('')
+  const [rawModel, setRawModel] = useState(undefined)
```
trevorlang, line 16: I _think_ we can get away with `useState()` for these two lines

PR #1176, opened 2020-08-14
EDSC-2808: Updates libraries and configs for bamboos node docker image

abbottry:


bin/deploy-bamboo.sh
```diff
@@ -47,7 +47,7 @@ node_modules
 EOF
 
 cat <<EOF > Dockerfile
-FROM node:10.15
+FROM node:12.16
```
jonvel, line 50: I _LOVE_ it!

PR #1177, opened 2020-08-17
EDSC-2858: Fixes hovering granule paths that cross the international dateline

macrouch:


static/src/js/components/GranuleResults/GranuleResultsItem.scss
```diff
@@ -47,6 +47,21 @@
     z-index: -1;
   }
 
+  &--hovered {
+    border: 1px solid #48bd7c;
```
abbottry, line 51: @trevorlang seems like we should use variables for these colors in the future

PR #1191, opened 2020-09-15
EDSC-2884: Fixes opendap params when no spatial is applied

macrouch:


static/src/js/actions/granules.js
```diff
@@ -239,19 +239,26 @@ export const fetchOpendapLinks = retrievalCollectionData => (dispatch, getState)
...
-    neLat,
-    neLng
-  } = mbr({
-    boundingBox: boundingBox[0],
-    circle: circle[0],
-    point: point[0],
-    polygon: polygon[0]
-  })
+  if (
+    boundingBox.length
```
abbottry, line 243: In the few places that we check these, we use `.length > 0` -- can you make that minor adjustment here? We may eventually want a `hasSpatialFilterApplied` util given how many places were using this, and we have that ticket to add it to the UI as well.

abbottry (approved):

trevorlang (approved):

PR #1198, opened 2020-09-22
feature/IDN-52: initial creation of the SCAR Portal

mmorahan:
Initial creation of the SCAR Portal.


portals/scar/config.json
```diff
@@ -0,0 +1,20 @@
+{
+  "hasStyles": true,
+  "hideCollectionFilters": true,
```
macrouch, line 3: This key doesn't exist anymore, it should be removed

macrouch (changes requested):
The logo file included is really huge compared to the other portal logos. The image is 1466 × 1420 when we are shooting for something more like 50x50. The whitespace around the logo is also all over the place, meeting the top edge but different amounts of empty space on all the other sides. That needs to be much smaller and cleaned up

mmorahan:
The SCAR provider sent new requirements that will need to be agreed on by the NASA customer. MDQ will write a new pull request if NASA agrees with the provider’s changes.

PR #1204, opened 2020-09-24
EDSC-2829: Defaults output projection and format to None and fixes check for variable subsetting

abbottry:


static/src/js/components/AccessMethod/AccessMethod.js
```diff
@@ -322,9 +322,12 @@ export class AccessMethod extends Component {
       )
 
       // Build options for supportedOutputFormats
-      supportedOutputFormatOptions = supportedOutputFormatOptions.map(format => (
-        <option key={format} value={ousFormatMapping[format]}>{format}</option>
-      ))
+      supportedOutputFormatOptions = [
+        <option key={null} value={null}>None</option>,
```
trevorlang, line 326: Is null the correct value for the key here? Seems like we might want to define a string maybe?

trevorlang (approved):

static/src/js/components/AccessMethod/AccessMethod.js
abbottry: I found _zero_ mentions of this effort online that even included a key, It's not IN the loop, I could remove it if that would be preferable.

abbottry: Cancel that, can't remove it. I don't see an issue with null, personally 🤷‍♂️  -- it indicates the same information to React with respect to change.

macrouch (approved):

PR #1212, opened 2020-09-29
EDSC-2899: Fixes Harmony spatial requests

macrouch:


abbottry (approved):

serverless/src/submitHarmonyOrder/bboxToPolygon.js
```diff
@@ -0,0 +1,18 @@
+/**
```
trevorlang, line 1: Not sure we need to do it now, but this seems like it'd be useful as a shared util.

serverless/src/submitHarmonyOrder/ccwShapefile.js
```diff
@@ -0,0 +1,36 @@
+import { isClockwiseLatLng } from '../../../static/src/js/util/map/granules'
+
+/**
```
trevorlang, line 3: Not sure we need to do it now, but this seems like it'd be useful as a shared util.

trevorlang (approved):

Issue #1223, opened 2020-10-01
EDSC-2912: Add keyboard shortcut to focus on search bar


mreese84:
### Description
Add keyboard shortcut to focus on the search bar input field when on a page that supports search (/search, /search/granules).

The `/` button should be used for this shortcut.

### Acceptance Criteria
- When on a search supported page (/search, /search/granules), hitting the `/` button on the keyboard focuses on the search input field.
- When not on a search supported page (/search, /search/granules), hitting the `/` button on the keyboard has no impact on the application.

i-hardy:
Hi, I'd like to take this!

mreese84:
Excellent. Go for it! Thank you for contributing.

mreese84:
The commit for our first keyboard shortcut can be [found here](https://github.com/nasa/earthdata-search/commit/e58d0195f17441fca1e44ff9094368c3947464e3) (amongst a few other changes 😄 )

PR #1231, opened 2020-10-02
EDSC-2912: Adds keyboard shortcut to focus on search bar

i-hardy:
Implements #1223

I've followed the same pattern as for the panel open/close shortcut in [this PR](https://github.com/nasa/earthdata-search/commit/e58d0195f17441fca1e44ff9094368c3947464e3); I did wonder if it might be desirable to have shortcuts organised somewhere more centrally rather than scattered about between components, but I think that would be outside the scope of this issue.

Please let me know about any needed changes!

### DIFF
--- a/static/src/js/components/SearchForm/SearchForm.js
+++ b/static/src/js/components/SearchForm/SearchForm.js
@@ -140,10 +140,12 @@ class SearchForm extends Component {
   onWindowKeyDown(e) {
     const { key } = e
     const { inputRef, keyboardShortcuts } = this
-    const inputIsRendered = inputRef.current && inputRef.current.input
+    const inputIsRenderedAndNotFocused = inputRef.current
+    && inputRef.current.input
+    && inputRef.current.input !== document.activeElement
 
     // Focus the search input when the forward slash key is pressed
-    if (key === keyboardShortcuts.focusSearchInput && inputIsRendered) {
+    if (key === keyboardShortcuts.focusSearchInput && inputIsRenderedAndNotFocused) {
       inputRef.current.input.focus()
 
       e.preventDefault()
--- a/static/src/js/components/SearchForm/SearchForm.js
+++ b/static/src/js/components/SearchForm/SearchForm.js
@@ -64,6 +64,10 @@ class SearchForm extends Component {
     }
   }
 
+  componentWillUnmount() {
+    window.removeEventListener('keydown', this.onWindowKeyDown)
+  }
+
   onFormSubmit(e) {
     e.preventDefault()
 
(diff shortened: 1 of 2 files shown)


static/src/js/components/SearchForm/SearchForm.js
```diff
@@ -123,6 +137,20 @@ class SearchForm extends Component {
...
 
+  onWindowKeyDown(e) {
+    const { key } = e
+    const { inputRef, keyboardShortcuts } = this
+    const inputIsRendered = inputRef.current && inputRef.current.input
+
+    // Focus the search input when the forward slash key is pressed
+    if (key === keyboardShortcuts.focusSearchInput && inputIsRendered) {
+      inputRef.current.input.focus()
+
```
trevorlang, line 148: I think we can have a block here that checks to see if the input is currently already focused, and if so, allows the `/` to be entered in the input.

trevorlang (changes requested):
This is looking great so far. I noticed a small edge case where the current code is preventing the `/` character from being entered into the input once it's focused. We're going to want to make sure that users are not prevented from using that character in their search terms.

Additionally, a unit test would be great here. I noticed that we were missing a test previously, so Ive added one on this [PR](https://github.com/nasa/earthdata-search/pull/1232) as an example. In addition to checking that the input is focused, It would be great if the unit test you implement alos tests that the `/` character is allowed in the field once focused.

We certainly agree that a single place to manage these shortcuts will work better in the long run. We're going to be putting some thought into that soon.

static/src/js/components/SearchForm/SearchForm.js
i-hardy: Makes sense!

mreese84 (approved):
Tested this out locally. Works as expected. Nice work! Let's make sure @trevorlang has a chance to review since he requested changes.

trevorlang (approved):
Looks good to me!

mreese84:
@i-hardy thank you for your contribution! Once master builds and gives me a release number I will pass that along to you so you can see your contributions out in the wild. All of our releases have to go through thorough tests in lower environments. The release with your contribution is scheduled to be released to our production environment on Wednesday, October 28th. If that changes I'll let you know.

i-hardy:
Happy to help, and thanks for the support in getting it merged!

mreese84:
@i-hardy, this will be in our 1.130.x release which is scheduled for production deployment no later than 10/28. Again, thanks for contributing. Feel free to look through the other issues and grab one if you'd like. We'll be adding more stories as the month progresses.

https://search.earthdata.nasa.gov/search

Issue #1226, opened 2020-10-01
EDSC-2898: Cannot remove filter from granule search box


mreese84:
### Description
Cannot remove search from granule search box. Deleting text and hitting 'Enter' causes the page to reload with the text re-populated.

**Reproduction Steps**

1. Visit https://search.uat.earthdata.nasa.gov/search/granules?p=C447308-ORNL_DAAC&pg[0][gsk]=-start_date&tl=1585761296!4!!
2. Copy/paste "1860_1993_2050_NITROGEN.N-deposition2050.txt" into the granule search field
3. Observe that only a single granule is returned
4. Delete the text in the granule search field and press enter
5. Observe that the page refreshed, the text just deleted remains, and only a single granule is returned

### Acceptance Criteria

1. Load any collection with granules
2. Copy/paste any of the granule titles into the search field and observe that the filter works as intended
3. Remove the granule search text and observe that no granule filter/search is applied and all expected granules return

ManiacalBilby:
Going to look into this.

heykc:
I can cover this issue! Setting up a draft PR in a few ✌️

mreese84:
Awesome! Thank you @heykc

abbottry:
Please make sure that you're running the tests (`npm run test` is the command, you can find other information in the readme) -- your PR has a failing test on some of the changes you made.

heykc:
@abbottry apologies for not running the tests before pushing! I do have a question for you or anyone that would be able to answer. The test directly checks that props.onApplyGranuleFilters is called with `{ readableGranuleName: null }` however my change directly affects what we're sending to props.onApplyGranuleFilters (sending an empty array rather than null). Obviously I don't want to just change the test so that it passes with my fix, but correct me if I'm wrong, I do believe we should be passing in an empty array rather than a null value. 

If we pass a null value, we remove the null filter completely once the prunedFilters reducer runs inside of collections.js (line 116 in `\static\src\js\actions\collections.js` . Unless we somehow retain and pass the null filter to the UPDATE_GRANULE_SEARCH_QUERY action, we won't be able to update the state in Redux with the updated readableGranuleName value. And since an empty array is still truthy, we wouldn't be removing it in the prunedFilters reducer.

Let me know if you agree. If you do, I can make the change to the test file to expect an empty array, otherwise, I'm happy to discuss possible other actions we can take. ✌️

PR #1233, opened 2020-10-02
EDSC-2898: Fix clearing existing granule filter in search bar

heykc:
## Fixes
Fixes: #1226 

## Description
In the collections.js file, in the updateFocusedCollectionGranuleFilters function, we only add a filter or overwrite an existing filter if at least one truthy value exists within the new filter. However if we clear the granule search bar it will pass a null value through to said function, which will never set due to the falsy nature of null. My fix passes an empty array for two reasons:

1) Normally the search input is converted into an array of values when not blank, so this keeps the datatype consistent regardless of content inside the search input.
2) An empty array is a truthy value, therefore it will successfully override the current filter inside the updateFocusedCollectionGranuleFilters function.

<details>
<summary><b>updateFocusedCollectionGranuleFilters</b> function inside <i>/static/src/js/actions/collections.js</i> that contains the condition that checks for at least one truthy value in order to set the new filter:
</summary>
<p>

```js
export const updateFocusedCollectionGranuleFilters = granuleFilters => (dispatch, getState) => {
  const state = getState()

  const focusedCollectionId = getFocusedCollectionId(state)

  const { query = {} } = state
  const { collection = {} } = query
  const { byId = {} } = collection
  const { [focusedCollectionId]: focusedCollectionQuery = {} } = byId
  const { granules: existingGranuleFilters = {} } = focusedCollectionQuery

  const allGranuleFilters = {
    ...existingGranuleFilters,
    ...granuleFilters
  }

  // Prune empty filters before sending to the store
  const prunedFilters = Object.keys(allGranuleFilters).reduce((obj, key) => {
    const newObj = obj

    // If the value is not an object, only add the key if the value is truthy. This removes
    // any unset values
    if (!isPlainObject(allGranuleFilters[key])) {
      if (allGranuleFilters[key]) {
        newObj[key] = allGranuleFilters[key]
      }
    //--------- v CONDITION PREVENTING FALSY VALUES FROM SETTING NEW FILTER v----------//
    } else if (Object.values(allGranuleFilters[key]).some(key => !!key)) {
      // Otherwise, only add an object if it contains at least one truthy value
      newObj[key] = allGranuleFilters[key]
    }

    return newObj
  }, {})

  dispatch({
    type: UPDATE_GRANULE_SEARCH_QUERY,
    payload: {
      collectionId: focusedCollectionId,
      ...prunedFilters
    }
  })
}
```
</p>
</details>

## Tests
After applying the fix, I've repeated the repro steps as defined in issue #1226 and observed the query within the url related to text in the search bar successfully is removed when the search bar is cleared and is either unfocused or when the enter key has been depressed.

## Important Note
I was unsuccessful migrating the db over to a local db, so I cannot say for certain if the granules are refreshed at the time of the clearing of the query within the url. If someone would like to help me solve my issue with migration I'd be happy to confirm this portion of the Acceptance criteria, otherwise if someone can pull my PR and confirm this for me who has access to a running local db, that would be equally as helpful.

### DIFF
--- a/static/src/js/components/GranuleResults/GranuleResultsHeader.js
+++ b/static/src/js/components/GranuleResults/GranuleResultsHeader.js
@@ -101,7 +101,9 @@ class GranuleResultsHeader extends Component {
     const { searchValue, prevSearchValue } = this.state
 
     if (searchValue !== prevSearchValue) {
-      let readableGranuleName = null
+      // empty array (truthy value) necessary
+      // to override existing filter with empty input
+      let readableGranuleName = []
 
       if (searchValue) {
         readableGranuleName = searchValue.split(',')
--- a/static/src/js/components/GranuleResults/__tests__/GranuleResultsHeader.test.js
+++ b/static/src/js/components/GranuleResults/__tests__/GranuleResultsHeader.test.js
@@ -223,7 +223,7 @@ describe('GranuleResultsHeader component', () => {
 
       expect(props.onApplyGranuleFilters).toHaveBeenCalledTimes(1)
       expect(props.onApplyGranuleFilters).toHaveBeenCalledWith({
-        readableGranuleName: null
+        readableGranuleName: []
       })
     })
   })


mreese84:
As mentioned on the issue, please make sure that you're running the tests (`npm run test` is the command, you can find other information in the readme) -- your PR has a failing test on some of the changes you made.

abbottry:
@heykc Your change would suggest that the test would need to be updated (proof that unit tests aren't always as thorough as we'd like) -- without pulling it down and testing it your assumption is correct, update the test and when I have a second I'll pull the branch down and test it all.

abbottry (approved):

macrouch (approved):

PR #1257, opened 2020-10-22
EDSC-2948: Continue loading page after named project params are loaded

macrouch:
# Overview

### What is the feature?

Naming a project and then refreshing the page was causing nothing to load. This fixes that bug.

### What is the Solution?

A return statement in the action that loads the named project's parameters from lambda was causing none of our 'get' actions to be called to load the data. I removed that return statement and updated the code that uses the decoded parameters.

List impacted areas.

# Testing

### Reproduction steps

Name a project, refresh the page

# Checklist

### DIFF
--- a/static/src/js/actions/urlQuery.js
+++ b/static/src/js/actions/urlQuery.js
@@ -81,9 +81,9 @@ export const changePath = (path = '') => async (dispatch, getState) => {
 
     const { projectId } = parse(queryString)
 
-    const projectResponse = await requestObject.fetch(projectId)
-
     try {
+      const projectResponse = await requestObject.fetch(projectId)
+
       const { data } = projectResponse
       const {
         name,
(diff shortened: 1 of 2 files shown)


trevorlang (approved):

abbottry (approved):

static/src/js/actions/urlQuery.js
```diff
@@ -73,13 +73,15 @@ export const changePath = (path = '') => async (dispatch, getState) => {
...
+  let decodedParams
+
   // If query string is a projectId, call getProject
   if (queryString && queryString.indexOf('projectId=') === 0) {
     const requestObject = new ProjectRequest(undefined, earthdataEnvironment)
 
     const { projectId } = parse(queryString)
 
-    const projectResponse = requestObject.fetch(projectId)
+    await requestObject.fetch(projectId)
```
abbottry, line 84: I don't like the combination of async/await and then/catch. This `await` should also prompt a change to using a try/catch where anything in the `then` block` is executed after the `await` within a `try` and the `catch` is moved to the `catch` block of the `try`.

abbottry (changes requested):
See comments in `static/src/js/actions/urlQuery.js`

abbottry (approved):

PR #1378, opened 2021-04-30
EDSC-3081: Updates subscriptions query stringification

macrouch:


abbottry (approved):

static/src/js/util/subscriptions.js
```diff
@@ -16,5 +16,5 @@ export const prepareSubscriptionQuery = (params) => {
   delete prunedParams.sortKey
 
   // Return the remaining parameters stringified, don't encode the values because CMR won't accept them
-  return stringify(prunedParams, { encode: false })
+  return stringify(prunedParams, { encode: false, indices: false, arrayFormat: 'brackets' })
```
ygliuvt, line 19: There are some granule search parameters in CMR that are indexed. e.g. https://cmr.earthdata.nasa.gov/search/site/docs/search/api.html#g-passes.

ygliuvt: Will this code handle the indexed case?

trevorlang (approved):

static/src/js/util/subscriptions.js
macrouch: We don't support those parameters, do you know of any others?

ygliuvt: Good. That is only one on granules. But if this function would also be used for collection searches, you are in trouble as science-keywords parameter are indexed.

macrouch: Cool, this code is only for subscriptions, so only granule queries

ygliuvt: Based on where this code is located, I don't think it is used for collection searches.

PR #1425, opened 2021-06-28
EDSC-3178: Removes wildcard characters when a user enters a quoted string into keyword search

macrouch:
# Overview

### What is the feature?

Quoted strings are now possible in CMR, but we don't want to add wildcard characters between words when a user provides a quoted string

### DIFF
--- a/static/src/js/util/collections.js
+++ b/static/src/js/util/collections.js
@@ -169,7 +169,13 @@ export const buildCollectionSearchParams = (params) => {
   }
 
   // If there is a keyword, add the wildcard character between words and following the final character
-  const keywordWithWildcard = !keyword ? undefined : `${keyword.replace(/\s+/g, '* ')}*`
+  // If the user types a quoted string, don't add the wildcard character
+  let keywordWithWildcard
+  if (keyword && keyword.match(/(".*")/)) {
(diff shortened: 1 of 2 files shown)


static/src/js/util/collections.js
```diff
@@ -169,7 +169,13 @@ export const buildCollectionSearchParams = (params) => {
...
 
   // If there is a keyword, add the wildcard character between words and following the final character
-  const keywordWithWildcard = !keyword ? undefined : `${keyword.replace(/\s+/g, '* ')}*`
+  // If the user types a quoted string, don't add the wildcard character
+  let keywordWithWildcard
+  if (keyword && keyword.match(/(".*")/)) {
+    keywordWithWildcard = keyword
+  } else if (keyword) {
+    keywordWithWildcard = `${keyword.replace(/\s+/g, '* ')}*`
+  }
```
sxu123, line 178: The only comment I have, which I have no solution for, is when they search for "word" and word, they might not get the same result, because of the way you're adding the wildcards(there must be a requirement). If you use the CMR endpoint directly, they should return the same result.

trevorlang (approved):

abbottry (approved):

PR #1426, opened 2021-06-30
EDSC-3208: Updates the methods used to determine collection capabilities

abbottry:
# Overview

### What is the feature?

Uses more than 1 granule to determine collection capabilities.

### What is the Solution?

Uses the granules retrieved for the project collection instead of the tag data (though tag data is still here and has been updated to use more than 1 granule as well).

Access methods within a project.

# Checklist


serverless/src/getAccessMethods/handler.js
```diff
@@ -79,9 +81,12 @@ const getAccessMethods = async (event, context) => {
     let onlineAccessFlag = false
     if (associatedGranules) {
       const { items: granuleItems } = associatedGranules
-      const [firstGranule] = granuleItems;
-
-      ({ onlineAccessFlag = false } = firstGranule)
+      onlineAccessFlag = getCollectionCapabilities({
```
macrouch, line 84: `getCollectionCapabilities` feels like overkill for this. We just need to find if any of the granuleItems have `online_access_flag = true`. I think this would work better
```
      const { items: granuleItems = [] } = associatedGranules
      const downloadable = granuleItems.some((granule) => {
        const { onlineAccessFlag = false } = granule
        return onlineAccessFlag
      })
```

Probably need to throw it into a sharedUtil because `actions/accessMethods.js` needs the same logic. On line 97 of that file it is still looking at tag data to determine if downloadable is true instead of looking at the granules.

macrouch (approved):

sdroger1 (approved):

PR #1428, opened 2021-07-08
EDSC-3204: Use GraphQL request to populate download links

macrouch:
# Overview

### What is the feature?

Uses GraphQL to populate the granule download links. This allows us to minimize the granule metadata payload size by only requesting the links that match the types we define in the request, as well as ignoring all the other granule metadata


static/src/js/actions/granules.js
```diff
@@ -170,23 +172,80 @@ export const fetchLinks = retrievalCollectionData => (dispatch, getState) => {
...
+      linkTypes: ['data', 's3'],
+      collectionConceptId: collectionId
     })
       .then((response) => {
         const { data } = response
-        const { feed } = data
-        const { entry } = feed
+        const { data: granulesData } = data
+        const { granules } = granulesData
+        const { items } = granules
```
abbottry, line 244: Remember that this can be null, we should initialize `items` to be an empty array or throw a conditional around the code below this.

trevorlang (approved):

mschmele (approved):

PR #1432, opened 2021-07-16
EDSC-2460: Use CMR scrolling for granule download links

macrouch:
# Overview

### What is the feature?

This PR changes the granule downloads page from sending off parallel requests with `pageNum` to using CMR's scrolling requests to fetch the granule download links. It also adds a progress bar to the `DownloadFilesPanel` component to show the user their progress while waiting to fetch all the links.

### DIFF
--- a/static/src/js/actions/granules.js
+++ b/static/src/js/actions/granules.js
@@ -237,7 +237,8 @@ export const fetchLinks = retrievalCollectionData => async (dispatch, getState) 
   let response
 
   try {
-    await Array.from(Array(totalPages)).forEachAsync(async (_, currentPage) => {
+    // Request an additional page (totalPages + 1) when scrolling to force GraphQL to clear the scroll session
+    await Array.from(Array(totalPages + 1)).forEachAsync(async (_, currentPage) => {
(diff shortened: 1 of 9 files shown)


static/src/js/actions/granules.js
```diff
@@ -223,51 +226,66 @@ export const fetchLinks = retrievalCollectionData => (dispatch, getState) => {
...
-              s3: granuleS3Links.map(lnk => lnk.href)
-            }
-          }))
-        }
+  let cursor
+  let response
+
+  try {
+    // Request an additional page (totalPages + 1) when scrolling to force GraphQL to clear the scroll session
+    await Array.from(Array(totalPages + 1)).forEachAsync(async (_, currentPage) => {
```
abbottry, line 241: I'm hesitant to implement scroll sessions using known page numbers over looping until we get nothing because thats how CMR expects it to work -- technically speaking the results could change while paging, I guess? I think we should change this to looping until its empty, even though the progress bar uses the page numbers for its math, I don't see a way around that.

abbottry (approved):

PR #1485, opened 2022-02-21
EDSC-3352: Use added granule metadata for access methods

macrouch:
When using the additive method to select granules only the granules selected are analyzed for online_access_flag to determine the available access methods.

### DIFF
--- a/static/src/js/actions/errors.js
+++ b/static/src/js/actions/errors.js
@@ -43,8 +43,6 @@ export const handleError = ({
   requestObject,
   errorAction
 }) => (dispatch, getState) => {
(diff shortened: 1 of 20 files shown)


static/src/js/actions/errors.js
```diff
@@ -43,6 +43,8 @@ export const handleError = ({
   requestObject,
   errorAction
 }) => (dispatch, getState) => {
+  console.log('🚀 ~ file: errors.js ~ line 46 ~ requestObject', requestObject)
```
trevorlang, line 46: Extra logs?

trevorlang (approved):

abbottry (approved):

PR #1517, opened 2022-05-31
EDSC-3482: Adds cloudHosted param to GraphQL query for collections export

macrouch:


trevorlang (approved):

static/src/js/actions/exportSearch.js
```diff
@@ -121,6 +123,11 @@ export const exportSearch = (format) => (dispatch, getState) => {
 
   const response = graphQlRequestObject.search(graphQuery, {
     ...buildCollectionSearchParams(collectionParams),
+    includeFacets: undefined,
```
DanielJDufour, line 126: Would you be able to add in a comment explaining why we are setting all these properties to `undefined`?  What happens when we use undefined?  Does includeTags or includeGranuleCounts defaults to true or false?

macrouch: I'll add a comment. Some of them include extra information in the CMR response that isn't used in the exported data, so we don't need to retrieve it. And pageNum/Size is handled be the export lambda, so we don't need to include those parameters in this request to that lambda

DanielJDufour (approved):

PR #1591, opened 2023-02-22
EDSC-3639: Submit orders to cmr-ordering

macrouch:


jaybarra (approved):

serverless/src/submitCmrOrderingOrder/handler.js
```diff
@@ -1,21 +1,22 @@
 import 'array-foreach-async'
-
 import axios from 'axios'
+import { v4 as uuidv4 } from 'uuid'
```
DanielJDufour, line 3: I think this could be replaced by 
```js
import { randomUUID } from 'node:crypto'

const requestId = randomUUID()
```

DanielJDufour (approved):

DanielJDufour:
approved, but left a comment about replacing uuid generation with node.js built-in. not required for this PR though

serverless/src/submitCmrOrderingOrder/handler.js
macrouch: It does look like it works, I'd like to write a ticket to replace `uuid` across the project so we can remain consistent in the code

PR #1596, opened 2023-03-01
EDSC-3673: Adds eslint job to github actions

macrouch:


.github/workflows/ci.yml
```diff
@@ -6,7 +6,37 @@ on:
   pull_request: {}
 
 jobs:
+  eslint:
+    runs-on: ubuntu-latest
+    strategy:
```
DanielJDufour, line 11: I don't think we need a matrix if we are only testing on one node-version, but lemme know if you have other plans here to expand this.

DanielJDufour (approved):

.github/workflows/ci.yml
DanielJDufour: But this is minor, so definitely not trying to block approval.

trevorlang (approved):

.github/workflows/ci.yml
macrouch: You're right matrix isn't necessary for a single version, but I like that all our workflow files are defined the same way, and some of them do use multiple versions

PR #1598, opened 2023-03-01
EDSC-3678: Ensure getProjectCollections waits for savedAccessConfigs

macrouch:
# Overview

### What is the feature?

Occasionally the graphql call to load project collections returns before the call the retrieve the saved access configs. I'm not able to reproduce locally, but I also saw that a similar error is displayed when the call to retrieve the saved access config fails

### What is the Solution?

`await` the call to retrieve the saved access configs, and default the saved access configs value to ensure access methods are still loaded without a successful saved access configs response

### DIFF
--- a/static/src/js/actions/project.js
+++ b/static/src/js/actions/project.js
@@ -182,13 +182,13 @@ export const getProjectCollections = () => async (dispatch, getState) => {
   }
 
   // Fetch the saved access configurations for the project collections
-  let savedAccessConfigs
+  let savedAccessConfigs = {}
   try {
     const savedAccessConfigsRequestObject = new SavedAccessConfigsRequest(
       authToken,
       earthdataEnvironment
     )
-    savedAccessConfigsRequestObject.search(
+    await savedAccessConfigsRequestObject.search(
       { collectionIds: filteredIds }
     ).then((response) => {
       const { data } = response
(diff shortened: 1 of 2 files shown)


eudoroolivares2016 (approved):

static/src/js/actions/project.js
```diff
@@ -182,13 +182,13 @@ export const getProjectCollections = () => async (dispatch, getState) => {
...
   // Fetch the saved access configurations for the project collections
-  let savedAccessConfigs
+  let savedAccessConfigs = {}
   try {
     const savedAccessConfigsRequestObject = new SavedAccessConfigsRequest(
       authToken,
       earthdataEnvironment
     )
-    savedAccessConfigsRequestObject.search(
+    await savedAccessConfigsRequestObject.search(
```
DanielJDufour, line 191: I think if we are using async/await syntax here, we probably can convert this whole block over to it (instead of leaving Promise/then), such as
```js
const response = await savedAccessConfigsRequestObject.search({ collectionIds: filteredIds })  

savedAccessConfigs = response.data
```

but this is just a minor suggestion and doesn't need to block approval

DanielJDufour (changes requested):

DanielJDufour (approved):

dpesall (approved):

PR #1602, opened 2023-03-08
EDSC-3677: Remove cmr_preferences from EDSC.

m4cy:
# Overview

### What is the feature?

Remove cmr_preferences column from EDSC users table and update references to it. 

### What is the Solution?

Use CMR-ordering request to fetch data directly from CMR when the contact info page loads.

Contact Info page and related lambdas getContactInfo and storeUserData, EDSC users table

# Testing

### Reproduction steps - similar to EDSC-3636

- SIT
1. Sign in to SIT Earthdata Search and go to the contact info page: https://search.sit.earthdata.nasa.gov/contact-info. Before the notification preferences load, you should see a spinner and the "update notification preference" button will be disabled. 
2. After your preferences load, you will see your current CMR-ordering notification level preference.
3. In a new tab, go to [CMR-ordering in SIT](https://studio.apollographql.com/sandbox/explorer) and update your user preferences with the [updateUser mutation](https://git.earthdata.nasa.gov/projects/CMR/repos/cmr-ordering/browse/docs/examples/graphql/mutations/updateUser.md).
}
4. Refresh the Earthdata contact info page and verify that the notification level changed.

### Attachments

Screen Shot 2023-03-08 at 2 43 13 PM
chments

image

Screen Shot 2023-03-08 at 2 42 47 PM

# Checklist


migrations/1677867813114_drop-cmr-preferences.js
```diff
@@ -0,0 +1,7 @@
+/* eslint-disable camelcase */
+
+exports.shorthands = undefined
```
DanielJDufour, line 3: I'm not an expert on node-pg-migrate, so it's more of a question than a comment.  Is this line `export.shorthands = undefined` necessary?

serverless/src/getContactInfo/__tests__/getCmrPreferencesData.test.js
```diff
@@ -15,28 +14,26 @@ beforeEach(() => {
...
         }
       })
 
-    const cmrData = await getCmrPreferencesData('urs_user', 'fake.access.token')
+    const cmrData = await getCmrPreferencesData('urs_user', 'fake.access.token', 'test')
 
-    expect(cmrData).toEqual({
-      ursId: 'urs_user',
-      notificationLevel: 'INFO'
+    const { data: preferences } = cmrData
```
DanielJDufour, line 32: I really like this.  Data is ambiguous, but calling it preferences helps clarify what we are looking at :-)

DanielJDufour (approved):

migrations/1677867813114_drop-cmr-preferences.js
m4cy: looking at the [documentation](https://github.com/salsita/node-pg-migrate/blob/master/docs/migrations.md), I don't believe it is, it's just part of their template. I experimented with removing it and making a migration and it can be dropped

DanielJDufour: sweet. thanks for looking into it.

serverless/src/getContactInfo/handler.js
```diff
@@ -65,6 +68,11 @@ const getContactInfo = async (event, context) => {
       body: JSON.stringify(contactInfoData)
     }
   } catch (e) {
+    console.log({
```
macrouch, line 71: We don't need this console.log here, in all the other lambdas we just return the error formatted by `parseError`, that in turn gets reported to the `errorLogger` lambda for logging

serverless/src/getContactInfo/handler.js
```diff
@@ -31,20 +33,46 @@ const getContactInfo = async (event, context) => {
...
+      access_token: authToken
+    } = await getAccessTokenFromJwtToken(jwtToken, earthdataEnvironment)
+
+    // Make a request to CMR-ordering instead of the EDSC database
+    const cmrPreferencesData = await getCmrPreferencesData(
+      userRecord.urs_id,
+      authToken,
+      earthdataEnvironment
+    )
+    const { status, data, errors } = cmrPreferencesData
```
macrouch, line 53: This is the wrong place to pull `errors` out. `status` and `data` here are returned by axios. Within the `data` object is the response from cmr-ordering, which will have `data` and/or `errors`. 

To see what the response with errors will look like, log out `cmrPreferencesData` here, and within `getCmrPreferencesData`, overwrite the `ursId` being sent as a variable in the graphql call. 
```
    data: {
      variables: {
        ursId: `${username}asdf`
      },
      query: cmrQuery
    },
```

Your logged out cmrPreferencesData will have this 
```
  data: { errors: [ [Object] ], data: { user: null } }
```

macrouch (changes requested):

serverless/src/getContactInfo/__tests__/handler.test.js
```diff
@@ -71,4 +85,31 @@ describe('getContactInfo', () => {
...
+          }
+        ])
+      } else {
+        query.response(undefined)
+      }
+    })
+
+    jest.spyOn(getCmrPreferencesData, 'getCmrPreferencesData').mockResolvedValue({
+      status: 200,
+      errors:
```
macrouch, line 105: I think this errors needs to be within a body object. Add an `expect(response.body).toEqual(...)` to ensure the error is being returned correctly

serverless/src/getContactInfo/handler.js
```diff
@@ -3,12 +3,15 @@ import { getApplicationConfig } from '../../../sharedUtils/config'
...
+import { getAccessTokenFromJwtToken } from '../util/urs/getAccessTokenFromJwtToken'
 import { parseError } from '../../../sharedUtils/parseError'
+import { getCmrPreferencesData } from './getCmrPreferencesData'
 
 /**
  * Handler for retreiving a users contact information
  * @param {Object} event Details about the HTTP request that it received
- * @param {Object} context Methods and properties that provide information about the invocation, function, and execution environment
+ * @param {Object} context Method
+ s and properties that provide information about the invocation, function, and execution environment
```
macrouch, line 14: I think a new line was added here by mistake

serverless/src/getContactInfo/handler.js
```diff
@@ -31,18 +34,41 @@ const getContactInfo = async (event, context) => {
...
+      authToken,
+      earthdataEnvironment
+    )
+    const { status, data } = cmrPreferencesData
+
+    const { errors } = data
+
+    if (errors) throw new Error(JSON.stringify(errors))
+
+    const { user } = data.data
```
macrouch, line 60: We try to avoid chaining like this (`data.data`), what we would typically do is rename `data` on line 54 to `responseData`, then on line 56 pull `errors` and `data` out of `responseData`, then this line becomes `const { user } = data`

serverless/src/getContactInfo/__tests__/getCmrPreferencesData.test.js
```diff
@@ -0,0 +1,39 @@
...
+      .reply(200, {
+        user: {
+          mock: 'cmr'
+        }
+      })
+
+    const cmrData = await getCmrPreferencesData('urs_user', 'fake.access.token', 'test')
+
+    const { data: preferences } = cmrData
+    expect(preferences).toEqual({
```
macrouch, line 33: I'm not sure if this expect is correct. Looking at the usage of the value returned from `getCmrPreferencesData`, I feel like maybe it is missing one of the `data`s, but I'm not 100% sure.

But I think this test can be better by testing the full response of the method. In the handler you are using a status, and the axios data object. The test here should be proving that the whole object is returned as expected, not just the value from cmr-ordering

serverless/src/saveContactInfo/handler.js
```diff
@@ -79,15 +79,17 @@ const saveContactInfo = async (event) => {
...
-    const { updateUser } = data.data
+    const { status, data } = cmrPreferencesResponse
+
+    const { errors } = data
 
-    // Save to database directly instead of queueing job
-    await dbConnection('users').update({ cmr_preferences: updateUser }).where({ id })
+    if (errors) throw new Error(JSON.stringify(errors))
+
+    const { updateUser } = data.data
```
macrouch, line 88: Same comment as `getContactInfo/handler.js` about `data.data`

trevorlang (approved):
I didn't see any issues that had not been noted already. Once that feedback is addressed, this looks good to me.

PR #1614, opened 2023-03-24
EDSC-3731: Download history page returning "Internal server error"

eudoroolivares2016:
# Overview

### What is the feature?

Removing some of the fields which are being passed to the get history of downloads page. Many are not needed and this is likely causing a user issue in Prod due to the payload growing too large

### What is the Solution?

Removing extraneous fields from what is passed to the history page

download history page

# Testing

### Reproduction steps

Download some collections and then check your download history to make sure that the page populates

### Attachments

# Checklist

### DIFF
--- a/serverless/src/getRetrievals/handler.js
+++ b/serverless/src/getRetrievals/handler.js
@@ -61,12 +61,15 @@ export default async function getRetrievals(event, context) {
         environment
       } = firstRow
 
+      const collectionMetadata = retrievalRecord.map((record) => record.collection_metadata)
+      const titles = collectionMetadata.map((collection) => ({ titles: collection }))
+
       retrievalsResponse.push({
         id,
         created_at: createdAt,
         jsondata,
         environment,
-        collections: retrievalRecord.map((record) => record.collection_metadata)
+        collections: titles
       })
     })
 
--- a/serverless/src/getRetrievals/__tests__/handler.test.js
+++ b/serverless/src/getRetrievals/__tests__/handler.test.js
@@ -37,7 +37,7 @@ afterEach(() => {
 describe('getRetrievals', () => {
   test('correctly retrieves retrievals', async () => {
     const determineEarthdataEnvironmentMock = jest.spyOn(determineEarthdataEnvironment, 'determineEarthdataEnvironment')
-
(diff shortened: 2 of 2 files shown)


serverless/src/getRetrievals/handler.js
```diff
@@ -61,12 +61,15 @@ export default async function getRetrievals(event, context) {
         environment
       } = firstRow
 
+      const collectionMetadata = retrievalRecord.map((record) => record.collection_metadata)
+      const titles = collectionMetadata.map((collection) => ({ titles: collection }))
```
macrouch, line 65: `titles` isn't the right key for this, it needs to match the collection metadata key of `title`. The output of the tests should not have changed.

You are still returning the full collection metadata content here, just nested under `titles`. Add more metadata to the in the `query.response` part of the test to show you are filtering out everything but the title

You are looping through the collection metadata twice with two calls to `.map`, you don't need to have two of them

macrouch (changes requested):

serverless/src/getRetrievals/handler.js
eudoroolivares2016: Added more metadata to test, removed the titles and single loop in the inline function

serverless/src/getRetrievals/handler.js
```diff
@@ -61,12 +61,19 @@ export default async function getRetrievals(event, context) {
         environment
       } = firstRow
 
+      const collectionMetadata = retrievalRecord.map((
+        { collection_metadata: collectionMetadata = {} }
```
trevorlang, line 65: Super minor thing, but I might format this like this instead:
```
const collections = retrievalRecord.map(({
    collection_metadata: collectionMetadata = {} 
}) => {
...
```

serverless/src/getRetrievals/handler.js
```diff
@@ -61,12 +61,19 @@ export default async function getRetrievals(event, context) {
         environment
       } = firstRow
 
+      const collectionMetadata = retrievalRecord.map((
```
trevorlang, line 64: Nit-picky, but you could clean this up very slightly by calling this `collections` instead of `collectionsMetadata`.

trevorlang (approved):
Two minor nit-picks, but looks great otherwise. Nice work!

macrouch (approved):
Please squash these commits

PR #1618, opened 2023-03-31
EDSC-3611: Removes legacy services

macrouch:


static/src/js/util/accessMethods/buildAccessMethods.js
```diff
@@ -0,0 +1,196 @@
...
+  } = collectionMetadata
+
+  const accessMethods = {}
+
+  let harmonyIndex = 0
+
+  const { items: serviceItems = null } = services
+
+  if (serviceItems !== null) {
+    serviceItems.forEach((serviceItem) => {
```
jaybarra, line 31: Could this `forEach` be split apart into smaller chunks? The first section could be simplified with a `filter`, and the each of the `if` feel like they might be able to be broken out into smaller functions

jaybarra (changes requested):

serverless/src/getSavedAccessConfigs/handler.js
```diff
@@ -0,0 +1,79 @@
...
+
+    const configs = {}
+
+    accessConfigRecords.forEach((config) => {
+      const {
+        collection_id: collectionId,
+        access_method: accessMethod
+      } = config
+
+      // After EDSC-3638, we are saving the form digest as `formDigest`, but previous access configurations will need to be updated
```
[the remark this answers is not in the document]
macrouch: We don't have versions, so no to a minimum version, it's just before this change we used `form_digest`, and after this change we use `formDigest`. The code right after this comment is accounting for that

static/src/js/util/accessMethods/buildAccessMethods.js
macrouch: Yes I think pulling the `if` statement's code out into smaller utils would be fine, I'm just not sure if I have time to do it before this has to be merged.

static/src/js/util/accessMethods/insertSavedAccessConfig.js
```diff
@@ -0,0 +1,95 @@
...
+              accessMethods[methodName] = {
+                ...methods[methodName],
+                form,
+                model,
+                rawModel,
+                formDigest
+              }
+            } else {
+              console.log('There was a problem parsing the savedAccessConfig values, using the default form instead.')
+            }
```
jaybarra, line 84: This is long enough to also warrant being extracted into a separate function.

jaybarra (approved):

serverless/src/requeueOrder/handler.js
```diff
@@ -60,7 +60,7 @@ const requeueOrder = async (event, context) => {
     } else if (type === 'ECHO ORDERS') {
       // Submits to Legacy Services (CMR) and is often referred to as an
```
eudoroolivares2016, line 61: Is this comment invalid now since legacy-services is no longer getting called for this

eudoroolivares2016 (approved):

PR #1660, opened 2023-08-14
CMR-9266: 2000 granule get-permission call taking too long

daniel-zamora:
# Overview

### What is the feature?

Access-control has problems with processing very large amounts of granule permissions within the AWS lambda timeout.

### What is the Solution?

We can blunt the impact on access-control processing times by avoiding overloading the app with everything all at once.  I've added an environment var ORDER_DELAY_SECONDS which will be added to the queue delay for each page of the total order. 

Order payload insertion into SQS.

# Testing

### Reproduction steps

- UAT, large orders of SMAP granules with large metadata sizes
- I tested with C1236624098-NSIDC_TS1

1. Order many granules
2. Monitor logs for cmr-ordering, to verify granule permission calls to access-control are batching and completing
3. Monitor access-control logs to verify permission calls are completing fast enough
4. Order completes successfully.

### DIFF
--- a/serverless/src/submitRetrieval/__tests__/handler.test.js
+++ b/serverless/src/submitRetrieval/__tests__/handler.test.js
@@ -43,6 +43,7 @@ beforeEach(() => {
   // Manage resetting ENV variables
   // TODO: This is causing problems with mocking knex but is noted as important for managing process.env
   // jest.resetModules()
+  process.env.orderDelaySeconds = 30
   process.env = { ...OLD_ENV }
   delete process.env.NODE_ENV
 })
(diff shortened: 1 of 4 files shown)


dpesall (approved):

serverless/src/submitRetrieval/__tests__/handler.test.js
```diff
@@ -140,6 +141,14 @@ describe('submitRetrieval', () => {
         query.response([{
           id: 5
         }])
+      } else if (step === 7) {
+        query.response([{
+          id: 5
```
macrouch, line 146: We already talked about this but I just want to remind myself Monday, these IDs need to be incremented in each response, and the `MessageBody`s below adjusted accordingly

daniel-zamora: I pushed to the wrong repo.

macrouch (approved):

PR #1677, opened 2023-09-26
EDSC-3601: Convert RadioList TemporalSelectionDropdown to functional components

bnp26:
# Overview

### What is the feature?

Convert RadioList TemporalSelectionDropdown from class components to functional components.
Converted the TemporalSelectionDropdown test to utilize React Testing Library

### What is the Solution?

Change the files to utilize functional component standards such as 
1) Changing classes to look like this
`const <class name> = (<props>) => {}`
`export default <class name>`
2) Utilizing useState instead of props = {} and setProps
3) Utilizing useEffects for tracking changes in state values

TemporalSelectionDropdown.js
RadioList.js

# Testing

### Reproduction steps

1. Step 1
2. Step 2...

### Attachments

# Checklist

### DIFF
--- a/static/src/js/components/FormFields/RadioList/RadioList.js
+++ b/static/src/js/components/FormFields/RadioList/RadioList.js
@@ -1,57 +1,37 @@
-import React, { Component, Children } from 'react'
+import React, { Children, useState, useEffect } from 'react'
 import PropTypes from 'prop-types'
 
 import Radio from '../Radio/Radio'
 import AccessMethodRadio from '../AccessMethodRadio/AccessMethodRadio'
 
-export class RadioList extends Component {
-  constructor(props) {
-    super(props)
+export const RadioList = ({ defaultValue, onChange, children }) => {
+  const [selected, setSelected] = useState(defaultValue)
 
-    this.state = {
-      selected: props.defaultValue
-    }
+  useEffect(() => {
+    setSelected(defaultValue)
+  }, [defaultValue])
 
-    this.onChange = this.onChange.bind(this)
+  const onPropsChange = (e) => {
+    const { target } = e
+    const { value } = target
+
+    onChange(value)
+
+    setSelected(value)
   }
 
-  UNSAFE_componentWillReceiveProps(nextProps) {
-    const { defaultValue: nextDefaultValue } = nextProps
-    const { defaultValue } = this.props
+  return (
+    <div className="radio-list">
+      {Children.map(children, (child) => {
+        const { props, type } = child
+        if (type !== Radio && type !== AccessMethodRadio) return null
 
-    if (nextDefaultValue !== defaultValue) {
-      this.setState({
-        selected: nextDefaultValue
-      })
-    }
-  }
-
-  onChange(e) {
-    const { onChange: propsOnChange } = this.props
-
-    propsOnChange(e.target.value)
-
-    this.setState({
-      selected: e.target.value
-    })
-  }
-
-  render() {
-    const { children } = this.props
-    const { selected } = this.state
-    const radios = Children.map(children, (child) => {
-      if (child.type !== Radio && child.type !== AccessMethodRadio) return null
-      const { props } = child
-      return React.cloneElement(child,
-        { onChange: this.onChange, checked: selected === props.value })
-    })
-
-    return (
-      <div className="radio-list">
-        {radios}
-      </div>
-    )
-  }
+        return React.cloneElement(child,
+          // eslint-disable-next-line react/prop-types
+          { onChange: onPropsChange, checked: selected === props.value })
+      })}
+    </div>
+  )
 }
 
 RadioList.defaultProps = {
--- a/static/src/js/components/TemporalDisplay/TemporalSelectionDropdown.js
+++ b/static/src/js/components/TemporalDisplay/TemporalSelectionDropdown.js
@@ -1,6 +1,5 @@
-import React, { PureComponent } from 'react'
+import React, { useState, useEffect } from 'react'
 import PropTypes from 'prop-types'
-import { isEqual } from 'lodash'
 import moment from 'moment'
 
 import Dropdown from 'react-bootstrap/Dropdown'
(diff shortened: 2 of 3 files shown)


static/src/js/components/TemporalDisplay/TemporalSelectionDropdown.js
```diff
@@ -21,101 +20,55 @@ import './TemporalSelectionDropdown.scss'
...
-    this.onRecurringToggle = this.onRecurringToggle.bind(this)
-    this.onChangeRecurring = this.onChangeRecurring.bind(this)
-    this.onToggleClick = this.onToggleClick.bind(this)
-    this.onDropdownToggle = this.onDropdownToggle.bind(this)
-    this.setEndDate = this.setEndDate.bind(this)
-    this.setStartDate = this.setStartDate.bind(this)
-    this.onValid = this.onValid.bind(this)
-    this.onInvalid = this.onInvalid.bind(this)
-  }
+// export default class TemporalSelectionDropdown extends PureComponent {
```
eudoroolivares2016, line 23: remove

static/src/js/components/TemporalDisplay/__tests__/TemporalSelectionDropdown.test.js
```diff
@@ -1,147 +1,216 @@
...
+    await waitFor(async () => {
+      await user.click(btn)
+    })
+
+    const startLabel = screen.getByText(/Start/i)
 
-    enzymeWrapper.instance().setStartDate(testObj)
-    expect(enzymeWrapper.state().temporal.startDate).toEqual('2012-01-01T12:00:00.000Z')
+    expect(startLabel).toBeInTheDocument()
+    unmount()
```
eudoroolivares2016, line 62: did you need this `unmount` for the test because something we being kept between them?

static/src/js/components/TemporalDisplay/__tests__/TemporalSelectionDropdown.test.js
```diff
@@ -1,147 +1,216 @@
...
+
+    const validEndDate = '2019-03-30T00:00:00.000Z'
+    const validStartDate = '2019-03-29T00:00:00.000Z'
+
+    const { getByRole, getAllByRole } = setup({
       onChangeQuery: onChangeQueryMock,
       temporalSearch: {
-        endDate: '2019-03-30T00:00:00.000Z',
-        startDate: '2019-03-29T00:00:00.000Z'
+        endDate: moment.utc(validEndDate).toISOString(),
```
eudoroolivares2016, line 148: Do you need to run `moment.utc` here? I would have thought that you could just pass the `validEndDate` and `validStartDate` like in the old test

static/src/js/components/TemporalDisplay/__tests__/TemporalSelectionDropdown.test.js
```diff
@@ -1,147 +1,210 @@
...
+    await waitFor(async () => {
+      await user.click(getByRole('button'))
+    })
 
-    expect(onChangeQueryMock).toHaveBeenCalledWith({
-      collection: {
-        temporal: {}
-      }
+    const inputs = getAllByRole('textbox')
+    expect(inputs[0].value).toBe(moment.utc(validStartDate).format('YYYY-MM-DD HH:mm:ss'))
```
eudoroolivares2016, line 152: I should've specified a little more clearly what I mean on that last this that last review. Do we need to call moment here?
```suggestion
    expect(inputs[0].value).toBe(validStartDate))
```
It seems like its already formatted and for react-testing-library I think we'd want to test against what is being rendered right on the DOM

static/src/js/components/TemporalDisplay/__tests__/TemporalSelectionDropdown.test.js
```diff
@@ -1,147 +1,210 @@
...
+
+    await waitFor(async () => {
+      await user.click(screen.getByRole('button'))
+    })
+
+    const inputs = screen.getAllByRole('textbox')
+
+    const invalidDate = '2012-01-efss 12:00:00'
+    const validStartDate = moment.utc('2012-01-01 12:00:00').format('YYYY-MM-DD HH:mm:ss')
+    const validEndDate = moment.utc('2012-01-02 12:00:00')
```
eudoroolivares2016, line 115: nitpick but, can you make the `validStartDate` and `validEnddate` format calls consistent here. Such as
```suggestion
    const validEndDate = moment.utc('2012-01-02 12:00:00').format('YYYY-MM-DD HH:mm:ss')
```
Then remove the `.toISOString()` call on line 126. The old test used toISOString for both not certain which one you need for this but, it seems like they should be the same

Similarly on line 129 I would think the assertion for  `validStartDate` and `validEndDate` would be the same

static/src/js/components/TemporalDisplay/__tests__/TemporalSelectionDropdown.test.js
```diff
@@ -1,147 +1,210 @@
...
+    const validStartDate = '2019-03-29T00:00:00.000Z'
+
+    fireEvent.click(screen.getByRole('button'))
+
+    const inputs = screen.getAllByRole('textbox')
+
+    fireEvent.change(inputs[0], { target: { value: validStartDate } })
+    fireEvent.change(inputs[1], { target: { value: validEndDate } })
+
+    expect(inputs[0].value).toBe(moment.utc(validStartDate).format('YYYY-MM-DD HH:mm:ss'))
```
eudoroolivares2016, line 189: this is the other spot where I would think you would not need to call moment and convert time
```suggestion
    expect(inputs[0].value).toBe(validStartDate)
```

static/src/js/components/TemporalDisplay/__tests__/TemporalSelectionDropdown.test.js
eudoroolivares2016: Same with the one on line 153

static/src/js/components/TemporalDisplay/__tests__/TemporalSelectionDropdown.test.js
eudoroolivares2016: and the one below on line 190

static/src/js/components/TemporalDisplay/__tests__/TemporalSelectionDropdown.test.js
bnp26: @eudoroolivares2016 For line 153-4, that does need to get formatted to be as I wrote. The originals are formatted in ISO format, but on the DOM, they get converted to be 'YYYY-MM-DD HH:mm:ss'

eudoroolivares2016 (approved):

static/src/js/components/FormFields/RadioList/RadioList.js
```diff
@@ -1,57 +1,37 @@
...
-
-    return (
-      <div className="radio-list">
-        {radios}
-      </div>
-    )
-  }
+  return (
+    <div className="radio-list">
+      {Children.map(children, (child) => {
```
trevorlang, line 25: This would be more readable with some new lines

```suggestion
      {
        Children.map(children, (child) => {
          const { props, type } = child
          if (type !== Radio && type !== AccessMethodRadio) return null

          return React.cloneElement(
            child,
            // eslint-disable-next-line react/prop-types
            {
              onChange: onPropsChange,
              checked: selected === props.value
            }
          )
        })
      }
```

trevorlang (approved):

PR #1684, opened 2023-10-16
EDSC-3869: Fixing projection switching between Antarctic and Arctic

eudoroolivares2016:
# Overview

### What is the feature?

There is an issue on EDSC where if you switch from south polar to north polar projects or vice versa the webpage will no longer function. This is to correct that error by preventing the Arithmetic operation that causes this from occurring

### What is the Solution?

Removing the function that is setting the center of the map. This value is being passed into leaflet's `setView` function. It seems to be holding onto the `lastCenter` field so when it tries to parse that value on the opposite polar projection it cannot find the coordinate. This is propagating as our web application crashing. We can remove this part of the code as it is not changing the view from what it is being set to when we already switch projections 

EDSC Map and projection switching

# Testing

### Reproduction steps

Switch from the south polar projection to the north polar projection and vice versa. Ensure the EDSC is still functioning after that as expected

### Attachments

N/A

# Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Earthdata Search uses PostgreSQL in production on AWS RDS. If you don't already 
     brew install postgresql
 
 Start the PostgreSQL server:
-    
+
     # If you have never used brew services before:
     brew tap homebrew/services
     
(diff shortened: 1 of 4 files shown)


dpesall (approved):

bnp26 (approved):
Looks good.

PR #1692, opened 2023-11-07
EDSC-3910: Switching between access methods is not updating the variables displayed

eudoroolivares2016:
# Overview

### What is the feature?

Fixing an issue where selecting a service then selecting from the variables on that service the Tree panel is not updating so we are seeing the variables associated to another service. In the default case should be that the variables which appear are the variable directly associated to the collection in the project, however if the service has variables associated to it we want those to show up instead of the ones on the direct-collection association (even if they are just a subset of those)

### What is the Solution?

Update the key prop in the render of the Tree component so that it has a reference to the `selectedAccessMethodServiceConceptId` such that when we change services this component is going to be re-rendered.

List impacted areas.

Project page and the variables tree view

### Reproduction steps

Find a collection with different sets of variables such as `C1200196166-SCIOPS` in the SIT env in this collection some of the services if selected should have the default set of `variables` and some should have those associated to the service. If we select one of those sets of variables and then switch to another then the variables on the later are going to be seen on the variable Tree view

We can quickly see these associations from `graphql` 
https://studio.apollographql.com/sandbox/explorer?endpoint=https%3A%2F%2Fgraphql.uat.earthdata.nasa.gov%2Fapi&explorerURLState=N4IgJg9gxgrgtgUwHYBcQC4QEcYIE4CeABABQAkADgIZ5VwDO6RAwhADZsJQoCWES9AJJIKMFAEoiwADpIiRKO07c%2BAktVoMmlGnXqSZc%2BUR4oEDKbOPHFSKAgopBYK9aIA3GjyoAjTvUsjNwUIGFRXYNNzAMNg6yQ6BAi423tHZ2TghN53BAygtwBfTKJigvl6fHceexiS%2BSiLWLj5VIcnF3K3FAIKJK7rTzxvPwQ6gbdFMJR6t0bxlqzE2eC29M7F4zLN7bjd633S1zLCkAAaECGR-wwQGRANPWkMZtdntY7n9GfmAEYAJgADIDfgBOABsv3B4IAtABlZiCADyAAU4c9ZIVToUgA

### Attachments

# Checklist

### DIFF
--- a/static/src/js/components/Tree/Tree.js
+++ b/static/src/js/components/Tree/Tree.js
@@ -49,7 +49,7 @@ export const Tree = ({
       variables,
       onUpdateFinished: forceUpdate
     })
-  }, [variables]) // Update TreeModel when the list of variables changes
+  }, []) // Only execute this useEffect once on initial render
 
   /**
    * Update the treeModel with new values
--- a/static/src/js/components/Tree/Tree.js
+++ b/static/src/js/components/Tree/Tree.js
@@ -49,7 +49,7 @@ export const Tree = ({
       variables,
       onUpdateFinished: forceUpdate
     })
-  }, []) // Only execute this useEffect once on initial render
+  }, [variables]) // Update TreeModel when the list of variables changes
 
   /**
    * Update the treeModel with new values
--- a/static/src/js/components/AccessMethod/__tests__/AccessMethod.test.js
+++ b/static/src/js/components/AccessMethod/__tests__/AccessMethod.test.js
@@ -26,6 +26,7 @@ const setup = (overrideProps) => {
   const onSelectAccessMethod = jest.fn()
   const onSetActivePanel = jest.fn()
   const onUpdateAccessMethod = jest.fn()
+  const onTogglePanels = jest.fn()
 
   const props = {
     accessMethods: {},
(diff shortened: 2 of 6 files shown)


static/src/js/components/Tree/Tree.js
```diff
@@ -49,7 +49,7 @@ export const Tree = ({
       variables,
       onUpdateFinished: forceUpdate
     })
-  }, []) // Only execute this useEffect once on initial render
+  }, [variables]) // Update TreeModel when the list of variables changes
```
macrouch, line 52: This does appear to fix the problem in the tree view, but I don't think it is the best way to fix this problem. Variables in a Tree will never change, so the fact this works means the Tree component is hanging around between the change of the access method, which would lead to unintended side effects like this bug. 

You can also this this same behavior in your test SIT collection by selecting an output format, then switching access methods and the output format will not have changed (or reset). I'm not sure if this is actually a problem because the options between those access methods are the same.

You can force React to re-render a component by adding the `key` prop. When the value you give it changes it will re-render the component. There is probably a more correct way to accomplish the same thing, but just wanted to note those thoughts about this particular solution

eudoroolivares2016: I agree thank you for the `key` I'm actively looking for a "better" refactor but, I have updated branch to pass the key with a unique identifier to include both the `id` of the service and the `view` so that it re-renders on a change to either

bnp26 (commented):
I went over this, and it looks good, but I'm not super confident about approving it. Just still feel like I don't have enough experience with the TreeNode and VariableTree code.

dpesall (approved):

rushgeo (approved):

PR #1696, opened 2023-11-22
EDSC-3923: Add clientId to EDD startDownload link

rushgeo:
# Overview

### What is the feature?

Add the `clientId` parameter to the `startDownload` link to track which EDSC environment initializes  the download

### What is the Solution?

Use the `env` property from the Application config for the `clientId` parameter and place it in the EDD URL.

EDD link generation.

# Testing

- Select some granules to download
- Choose the Direct Download access method
- Click the Download Data button
- Once the file URLs populate in the status window, click the Download Files button
- Close the Earthdata Download application if it opens, or refuse your browser's request to open it
- Right-click on the Open Earthdata Download button, and copy the URL
- Paste the URL somewhere, and verify that the it contains `&clientId=dev` after the mocked `authToken`

image

# Checklist


macrouch:
The clientId needs to be more specific, just using the env doesn't tell us EDSC sent the user to EDD. I'd look into the clientIds that we send to CMR and see if we can reuse anything from that logic/config

static/src/js/components/OrderStatus/OrderStatusItem.js
```diff
@@ -71,7 +71,8 @@ export class OrderStatusItem extends PureComponent {
     const { type: accessMethodType } = accessMethod
 
     // TODO: Add a second value and refresh at different intervals for the different types of orders
-    const { orderStatusRefreshTime } = getApplicationConfig()
+    const { orderStatusRefreshTime, env } = getApplicationConfig()
```
eudoroolivares2016, line 74: So in `static.config.json` we have `clientId` which is an object in the `application` JSON. I am not sure if this ticket is supposed to pass that along? We should make sure.

rushgeo: Those don't seem to reflect the running environment, but the ticket asked for `edsc-sit, edsc-uat, edsc-prod`
```json
    "clientId": {
      "background": "eed-PORTAL-ENV-serverless-background",
      "client": "eed-PORTAL-ENV-serverless-client",
      "lambda": "eed-PORTAL-ENV-serverless-lambda"
    }
```

macrouch: We have logic that replaces `PORTAL` and `ENV` with the appropriate values. That `PORTAL` gets replaced with the deployments defaultPortal, in our case that will be `edsc`, but if someone else deploys EDSC the idea is that they would setup their own default portal which would separate their requests from ours.

`sharedUtils/getClientId.js` is where the replacements happens.

To be consistent with how those clientIds are used over the rest of the application we should use `getClientId().client` since this would be the React client-side application sending the request. `background` is used in background lambdas and `lambda` is used for the rest of our application's lambdas

rushgeo: Implemented with latest commit

rushgeo:
The latest commit fixes the tests by using the correct `defaultPortal` value set in `static.config.json`. I was trying to use mocked functions to override the values used by `getClientId()`, which would be helpful if there was also a `overrideStatic.config.json` file present. That ended up not being worth the time investment, though, so I went with the simpler approach.

eudoroolivares2016 (approved):

PR #1698, opened 2023-12-04
EDSC-3930: Fixing issue on EDSC

eudoroolivares2016:
# Overview

### What is the feature?

Fix an issue on EDSC (Further info in the Teams file on this issue) this has been left purposefully vague

### What is the Solution?

Altered parsing of redirect urls for strictness

Regression testing was done locally to ensure that we could still be redirected from EDD to URS and to the EULA page using the prod collection `C1214470488-ASF` with a profile that had not accepted any Eulas

We can check `C2202497474-LPCLOUD` collection to ensure that EDD redirects through URS are getting sent as expected

List impacted areas.

EDD redirects
EDL redirects

### Reproduction steps

(Further info in the Teams file on this issue) this has been left purposefully vague

### Attachments

N/A

# Checklist

- [n/a] I have made corresponding changes to the documentation

### DIFF
--- a/static/src/js/containers/AuthCallbackContainer/AuthCallbackContainer.js
+++ b/static/src/js/containers/AuthCallbackContainer/AuthCallbackContainer.js
@@ -29,7 +29,6 @@ export const AuthCallbackContainer = ({
   location,
   onAddEarthdataDownloadRedirect
 }) => {
-  console.log('🚀 ~ file: AuthCallbackContainer.js:31 ~ location:', location)
   const { edscHost } = getEnvironmentConfig()
 
   useEffect(() => {
@@ -43,48 +42,37 @@ export const AuthCallbackContainer = ({
       redirect = '/'
     } = params
 
+    let eddRedirectUrl = eddRedirect
+    if (redirect.includes('earthdata-download')) {
+      eddRedirectUrl = redirect
+    }
+
     // Handle EDD redirects
-    // EDD redirects must begin with `earthdata-download`
-    const validEddRedirect = eddRedirect && eddRedirect.startsWith('earthdata-download')
-    console.log('🚀 ~ file: AuthCallbackContainer.js:49 ~ useEffect ~ eddRedirect:', eddRedirect)
-    console.log('🚀 ~ file: AuthCallbackContainer.js:49 ~ useEffect ~ validEddRedirect:', validEddRedirect)
+    if (eddRedirectUrl) {
+      const validEddRedirect = eddRedirectUrl.startsWith('earthdata-download')
+      if (validEddRedirect) {
+        if (accessToken) eddRedirectUrl += `&token=${accessToken}`
 
-    console.log('🚀 ~ file: AuthCallbackContainer.js:54 ~ useEffect💀 ', redirect.includes('earthdata-download'))
+        // Add the redirect information to the store
+        onAddEarthdataDownloadRedirect({
+          redirect: eddRedirectUrl
+        })
 
-    if (eddRedirect && !validEddRedirect) {
-      console.log('🚀 ~ file: AuthCallbackContainer.js:54 ~ useEffect ~ redirect:', redirect)
+        // Redirect to the edd callback
+        history.push('/earthdata-download-callback')
 
-      // Redirect to an error page or a safe location if the URL is not a relative path
-      // https://developer.mozilla.org/en-US/docs/Web/API/Location/replace assign prevents back-button use in history
+        return
+      }
+
       window.location.replace('/')
 
       return
     }
 
-    // If the redirect includes earthdata-download, redirect to the edd callback
-    if (validEddRedirect || redirect.includes('earthdata-download')) {
-      console.log('🚀 ~ file: AuthCallbackContainer.js:48 ~ useEffect ~ eddRedirect:', eddRedirect)
-      let eddRedirectUrl = eddRedirect || redirect
-      if (accessToken) eddRedirectUrl += `&token=${accessToken}`
-
-      // Add the redirect information to the store
-      onAddEarthdataDownloadRedirect({
-        redirect: eddRedirectUrl
-      })
-
-      // Redirect to the edd callback
-      history.push('/earthdata-download-callback')
-
-      return
-    }
-
     // Handle redirects
-    // If we are not redirecting to earthdata-search relative path for redirects that have a value
     const invalidRedirectUrl = redirect !== '/' && !redirect.startsWith(edscHost)
 
     if (invalidRedirectUrl) {
-      console.log('🚀 ~ file: AuthCallbackContainer.js:54 ~ useEffect ~ redirect:', redirect)
-
       // Redirect to an error page or a safe location if the URL is not a relative path
       // https://developer.mozilla.org/en-US/docs/Web/API/Location/replace assign prevents back-button use in history
       window.location.replace('/')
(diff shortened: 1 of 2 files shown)


static/src/js/containers/AuthCallbackContainer/AuthCallbackContainer.js
```diff
@@ -39,42 +42,40 @@ export const AuthCallbackContainer = ({
...
 
-      // Add the redirect information to the store
-      onAddEarthdataDownloadRedirect({
-        redirect: eddRedirectUrl
-      })
-
-      // Redirect to the edd callback
-      history.push('/earthdata-download-callback')
+    if (invalidRedirectUrl) {
+      // Redirect to an error page or a safe location if the URL is not a relative path
```
eudoroolivares2016, line 76: determine if should use `assign` or use `replace`?

static/src/js/containers/AuthCallbackContainer/AuthCallbackContainer.js
```diff
@@ -39,42 +42,40 @@ export const AuthCallbackContainer = ({
...
-        redirectUrl
-        && redirectUrl.protocol !== 'http:'
-        && redirectUrl.protocol !== 'https:'
-        && redirectUrl.protocol !== 'earthdata-download:'
-      ) {
-        // The redirectUrl is not a valid protocol
-        console.log('The redirectUrl is not a valid protocol')
-        window.location.replace('/')
+    let eddRedirectUrl = eddRedirect
+    if (redirect.includes('earthdata-download')) {
```
trevorlang, line 46: We typically have blank spaces in between variable assignments and blocks

static/src/js/containers/AuthCallbackContainer/AuthCallbackContainer.js
```diff
@@ -39,42 +42,40 @@ export const AuthCallbackContainer = ({
...
+      eddRedirectUrl = redirect
+    }
+
+    // Handle EDD redirects
+    if (eddRedirectUrl) {
+      const validEddRedirect = eddRedirectUrl.startsWith('earthdata-download')
```
trevorlang, line 52: Same comment as above

trevorlang, line 52: It might be better to do an exact match on the url just in case we use "earthdata-download" in a url sometime down the line. Seems safer to be a little more specific.

static/src/js/containers/AuthCallbackContainer/AuthCallbackContainer.js
trevorlang, line 46: It might be better to do an exact match on the url just in case we use "earthdata-download" in a url sometime down the line. Seems safer to be a little more specific.

static/src/js/containers/AuthCallbackContainer/AuthCallbackContainer.js
```diff
@@ -39,42 +42,40 @@ export const AuthCallbackContainer = ({
...
-      onAddEarthdataDownloadRedirect({
-        redirect: eddRedirectUrl
-      })
-
-      // Redirect to the edd callback
-      history.push('/earthdata-download-callback')
+    if (invalidRedirectUrl) {
+      // Redirect to an error page or a safe location if the URL is not a relative path
+      // https://developer.mozilla.org/en-US/docs/Web/API/Location/replace assign prevents back-button use in history
+      window.location.replace('/')
```
trevorlang, line 78: It seems like we could send the user to the standard error page here. I _think_ a "replace" makes more sense than a "assign" here.

static/src/js/containers/AuthCallbackContainer/AuthCallbackContainer.js
```diff
@@ -39,42 +42,40 @@ export const AuthCallbackContainer = ({
...
 
-    // Verify that the redirect params are real URLs
-    try {
-      let redirectUrl
-      if (eddRedirect) redirectUrl = new URL(eddRedirect)
-      if (redirect && redirect !== '/') redirectUrl = new URL(redirect)
-
-      if (
-        redirectUrl
-        && redirectUrl.protocol !== 'http:'
```
trevorlang, line 50: Seems like we lost where were checking for the allowed protocols. Is that intentional?

eudoroolivares2016: Yea that is intentional. It doesn't sufficiently cover the strictness we need to properly handle url redirects. Here we are now only allowing redirects to addresses relative to EDSC or earthdata-download

static/src/js/containers/AuthCallbackContainer/AuthCallbackContainer.js
eudoroolivares2016: A little confused about the exact matching. We wouldn't know what the `eddRedirectUrl` value would be only its prefix?

static/src/js/containers/AuthCallbackContainer/AuthCallbackContainer.js
eudoroolivares2016: For instance one value I see in testing with downloading a collection where EDL auth was required is `earthdata-download://authCallback?fileId=3&ee=prod&token=<my-token>`
I'm not sure we know what the fileId is going to be or really that the token matches whatever token we have saved

static/src/js/containers/AuthCallbackContainer/AuthCallbackContainer.js
eudoroolivares2016: Yea so the `fileId` coming from `EDD` is `The authUrl will have a `fileId` parameter appended to it, which is the file ID of the file that triggered the authentication workflow`

static/src/js/containers/AuthCallbackContainer/AuthCallbackContainer.js
rushgeo: One possibility is moving the test on line 53 with `startsWith` up here, which would mean only valid EDD redirects get processed by the following `if` block

rushgeo (commented):
Other than Trevor and I's comments on `includes('earthdata-download')`, it looks good to me.

static/src/js/containers/AuthCallbackContainer/AuthCallbackContainer.js
eudoroolivares2016: @rushgeo We could check if it starts with up there for `redirect`  but, we still need to have a `startsWith` check in case the `eddRedirectUrl` has been initialized to the value of the `eddRedirect`. I used `includes` because that was previous logic.

static/src/js/containers/AuthCallbackContainer/__tests__/AuthCallbackContainer.test.js
```diff
@@ -154,7 +154,41 @@ describe('AuthCallbackContainer component', () => {
...
+      }
+    })
+
+    expect(setSpy).toBeCalledTimes(0)
+
+    expect(window.location.replace.mock.calls.length).toBe(1)
+    expect(window.location.replace.mock.calls[0]).toEqual(['/not-found'])
+  })
+
+  test('does not follow the eddRedirect it is not a valid earthdata-download redirect', () => {
```
eudoroolivares2016, line 177: I don't know that this test is really very helpful. its very similar to the test below it and running though the same code. I added it for explicitness when it comes to what solution this PR is trying to do but, I'll pull it out if people deem to totally redundant

dpesall (approved):

PR #1699, opened 2023-12-05
EDSC-3881: Create portal for CASEI

rushgeo:
# Overview

### What is the feature?

Add a portal to EDSC for [CASEI](https://impact.earthdata.nasa.gov/casei/), the Catalog of Archived Suborbital Earth Science Investigations.

### What is the Solution?

A new portal was created using a query for the 107 projects/campaigns currently available from CASEI.

Note that if new projects are added to CASEI, we will need to update this portal config.

We have an old unused portal called [Suborbital](https://github.com/nasa/earthdata-search/tree/2b6b38b3f34c8af6c0cadd9e3f01470cad5fb374/portals/suborbital) that looks like it was intended to be use used for CASEI. I tested it, but its query returned fewer collections (669).

# Testing

- Click Browse Portals
- Ensure there is an option for CASEI, then click it to enter the portal
- There should be around 1,768 matching collections

### Attachments

CASEI added to the list of portals:

image

After entering portal:

image

# Checklist


portals/casei/config.json
```diff
@@ -0,0 +1,13 @@
...
...ct=OWLETS&project=ATom&project=NAAMES&project=ICE-POP&project=SPURS&project=ACEPOL&project=EPOCH&project=ASCENDS&project=CPEX&project=LMOS&project=CORAL&project=CARAFE&project=GOES-R&project=POSIDON&project=KORUS-AQ&project=KORUS-OC&project=OLYMPEX&project=AfriSAR&project=RADEX&project=CARVE&project=VIRGAS&project=AirMOSS&project=COMEX&project=Polar&project=ATTREX&project=CalWater&project=HS3&project=ARISE&project=DISCOVER-AQ&project=SABOR&project=IPHEx&project=SEAC4&project=MIZOPEX&project=IFloodS&project=HyMeX&project=PODEX&project=DC3&project=BROMEX&project=GCPEx&project=DEVOTE&project=COAST&project=ECO-3&project=ICESCAPE&project=MC3&project=MACPEX&project=LPVEx&project=GRIP&project=CASIE&project=AMISA&project=ARCTAS&project=TC4&project=CLASIC07&project=C3&project=NAMMA&project=CC-VEx&project=INTEX-NA&project=CR-AVE&project=TCSP&project=BigFoot&project=ACES&project=CRYSTAL-FACE&project=CAMEX&project=PAC2001&project=CLAMS&project=TRACE&project=SAFARI&project=ARESE&project=PEM&project=TEFLUN&project=FIRE&project=POLARIS&project=PROVE&project=BOREAS&project=TARFOX&project=ASHOE&project=SPADE&project=AASE&project=OTTER&project=ABLE&project=CITE&project=FIFE&project=AAOE&project=SNF"
```
macrouch, line 7: This is an absolute maintenance nightmare. Tags feel much more useful than a list of projects. But if we absolutely must do a list of projects then the config file needs to be an actual list of projects, not a very long string of url parameters.

eudoroolivares2016, line 7: Check if you can do this in the config
```suggestion
    "project": "ABoVE",
    "project":"SnowEx"
```
...

unless we customize some logic from the docs it didn't seem like CMR could parse `project=['proj1', 'proj2`] etc

rushgeo: Nope, EDSC just keeps the last line if there are multiple lines of `"project"` keys.
An array would be a nice compromise and might be worth altering parsing logic for.

macrouch: Why isn't this using tags?

rushgeo: We'll discuss with the product owner next week whether a tags-based workflow would be better for the data owners and users.

In the meantime, the latest commit changes the project list to a string array in `portals/casei/config.json`, and adds `project` to the array of keys in `sharedConstants/nonIndexedCmrKeys.js`. That allows it to be stringified correctly by `sharedUtils/prepKeysForCmr.js`.

eudoroolivares2016:
After discussing the context was that we don't expect these `projects` to change. This is a stop-gap measure until the UMM-C changes (or perhaps other determined CMR changes) can be made to accommodate casei field data at which point likely other measures (such as a new facet) may be used to filter the collections

Bccorb (approved):

eudoroolivares2016 (changes requested):
Lets fix the project facet or determine if we think that warrants holding this off

eudoroolivares2016 (approved):
We discussed that facet issue it is something existing so that should not stop this from completing

PR #1701, opened 2023-12-14
EDSC-3899: Prevent Users from Inputting Matching Coordinates in EDSC Spatial Selection with Rectangle

bnp26:
# Overview

### What is the feature?

Prevent Users from Inputting Matching Coordinates in EDS Spatial Selection with Rectangle

### What is the Solution?

Added to the validation of bounding box coordinates, and added to the error if the 2 coordinates match each other.

static/src/js/components/SpatialDisplay/SpatialDisplay.js

# Testing

### Reproduction steps

- Any
- Any

1. Navigate to http://localhost:8080/search
2. Click on the Crop icon.
3. Click on "Rectangle"
4. Add 2 coordinates that match each other
5. Observe that the error message that comes up is "SW and NE points contain matching coordinates. Please use point selection instead."
6. Change the coordinates to be 2 matching invalid coordinates
7. Observe that the error message is "Latitude (-91.119) must be between -90 and 90. SW and NE points contain matching coordinates. Please use point selection instead."
### Attachments

image

image

# Checklist

- [n/a] I have made corresponding changes to the documentation

### DIFF
--- a/static/src/js/components/SpatialDisplay/SpatialDisplay.js
+++ b/static/src/js/components/SpatialDisplay/SpatialDisplay.js
@@ -365,7 +365,7 @@ class SpatialDisplay extends Component {
     errorMessage = this.validateCoordinate(swPoint) + this.validateCoordinate(nePoint)
 
     if (swPoint === nePoint) {
-      const message = 'Coordinates must not match. Point subsetting is available.'
+      const message = 'SW and NE points contain matching coordinates. Please use point selection instead.'
       errorMessage += errorMessage !== '' ? `\n${message}` : message
     }
 
--- a/static/src/js/components/SpatialDisplay/SpatialDisplay.js
+++ b/static/src/js/components/SpatialDisplay/SpatialDisplay.js
@@ -362,7 +362,8 @@ class SpatialDisplay extends Component {
     let errorMessage = ''
     const [swPoint, nePoint] = boundingBox
 
-    errorMessage = this.validateCoordinate(swPoint) + this.validateCoordinate(nePoint)
+    errorMessage += this.validateCoordinate(swPoint)
+    if (errorMessage === '') { errorMessage += this.validateCoordinate(nePoint) }
 
     if (swPoint === nePoint) {
       const message = 'SW and NE points contain matching coordinates. Please use point selection instead.'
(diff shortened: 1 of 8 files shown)


static/src/js/components/SpatialDisplay/SpatialDisplay.js
```diff
@@ -359,9 +359,17 @@ class SpatialDisplay extends Component {
    * @param {Array} boundingBox Array with [swPoint, nePoint] values
    */
   validateBoundingBoxCoordinates(boundingBox) {
+    let errorMessage = ''
     const [swPoint, nePoint] = boundingBox
 
-    return this.validateCoordinate(swPoint) + this.validateCoordinate(nePoint)
+    errorMessage = this.validateCoordinate(swPoint) + this.validateCoordinate(nePoint)
+
+    if (swPoint === nePoint) {
```
eudoroolivares2016, line 367: So if a user has both an invalid coordinate and they are the same coordinates what does the error message look like its just both of those combined right?

bnp26: Yes sir!

static/src/js/components/SpatialDisplay/__tests__/SpatialDisplay.test.js
```diff
@@ -585,4 +585,28 @@ describe('SpatialDisplay component', () => {
...
+    test('returns an error coordinates match each other.', () => {
+      const { enzymeWrapper } = setup()
+      const input = '-77.119,38.791'
+
+      const boundingBoxResult = enzymeWrapper.instance()
+        .validateBoundingBoxCoordinates([input, input])
+
+      expect(boundingBoxResult).toEqual('SW and NE points contain matching coordinates. Please use point selection instead.')
+    })
+  })
```
eudoroolivares2016, line 599:
```suggestion
  })
      test('returns an error coordinates match each other.', () => {
      const { enzymeWrapper } = setup()
      const input = '-91.119,38.791'

      const boundingBoxResult = enzymeWrapper.instance()
        .validateBoundingBoxCoordinates([input, input])

      expect(boundingBoxResult).toEqual('Latitude (-91.119) must be between -90 and 90.\nSW and NE points contain matching coordinates. Please use point selection instead.')
    })
```
I think we can just move that one into the same describe block wdyt?

eudoroolivares2016 (approved):

eudoroolivares2016 (approved):

PR #1702, opened 2023-12-18
EDSC-3945: Resolve Snyk Vulnerability: Earthdata Search Client /earthdata-search

eudoroolivares2016:
# Overview

### What is the feature?

Removing the `node-pre-gyp` as it is not being used by EDSC but, the import contains a vulnerability. There a a more updated package which is the successor `@mapbox/node-pre-gyp` however, this package is not being used. The purpose of this package is to take publicly hosted `Addon` Files and using install script of package.json so they can be dynamically pulled in.

### What is the Solution?

Summarize what you changed.

Removed this unused package

Since this package is not being used it should not affect EDSC

# Testing

N/A 

Ensure regression

### Attachments

# Checklist


package.json
```diff
@@ -89,9 +89,9 @@
     "@aws-sdk/client-sqs": "^3.352.0",
     "@babel/core": "^7.23.2",
     "@babel/eslint-parser": "^7.22.15",
+    "@babel/plugin-syntax-dynamic-import": "^7.8.3",
     "@babel/plugin-transform-class-properties": "^7.22.5",
     "@babel/plugin-transform-object-rest-spread": "^7.22.15",
-    "@babel/plugin-syntax-dynamic-import": "^7.8.3",
```
eudoroolivares2016, line 94: The purpose of this package is to handle pre-building C++ plugins to the node env instance that an app is running on. The documentation of that package lists out requirements to use it. In those we see that we need to have `binaries` listed as an object in our `package.json` with a destination to compile the modules from (this can include an S3 bucket). We don't have that configuration in our package.json implying that this package is not being used. We may just be able to remove it

package.json
```diff
@@ -102,6 +102,7 @@
     "@edsc/geo-utils": "^1.0.4",
     "@edsc/smart-handoffs": "^1.0.5",
     "@edsc/timeline": "^1.1.7",
+    "@mapbox/node-pre-gyp": "^1.0.11",
```
eudoroolivares2016, line 105: This package is the successor the the previous one and is still being actively updated

bnp26 (approved):
Looks good

rushgeo (approved):
Looks like we're good to remove node-pre-gyp. Thanks!

PR #1704, opened 2023-12-21
EDSC-3884: Set latitude check to Ant/arctic circle

rushgeo:
# Overview

### What is the feature?

When a geoJSON file is uploaded, EDSC would change to one of the polar basemaps if all of the coordinates were above 45° north or south. However, the polar basemaps don't actually extend that far towards the equator, so the uploaded shape is in an unexpected location over a black background. Users also wouldn't expect the basemap to change unless the region of interest is much closer to the poles.

### What is the Solution?

This fix updates the latitude check to 66.5°, the approximate latitude of the Arctic and Antarctic Circles.

Display of uploaded geospatial files. 

# Testing

### Reproduction steps

1. Download [georgia_straight_vancouver.geojson.zip](https://github.com/nasa/earthdata-search/files/13746212/georgia_straight_vancouver.geojson.zip)
2. Unzip the file
3. Click the upload file button as shown:
image

4. Upload the unzipped geoJSON file
5. Verify that the basemap stays on the default Geographic/Equirectangular projection and does not switch to the north polar projection. The uploaded geoJSON file should be plotted as shown:
image

# Checklist

### DIFF
--- a/static/src/js/components/Map/ShapefileLayer.js
+++ b/static/src/js/components/Map/ShapefileLayer.js
@@ -215,8 +215,8 @@ class ShapefileLayerExtended extends L.Layer {
           featureLatLngs = featureLayer.getLatLngs().flat()
         }
 
-        allLatsArctic = featureLatLngs.every((latlng) => latlng.lat > 45)
-        allLatsAntarctic = featureLatLngs.every((latlng) => latlng.lat < -45)
+        allLatsArctic = featureLatLngs.every((latlng) => latlng.lat > 66.5)
+        allLatsAntarctic = featureLatLngs.every((latlng) => latlng.lat < -66.5)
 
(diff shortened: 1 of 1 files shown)


Bccorb (approved):

daniel-zamora (approved):

rushgeo:
We discussed alternative latitudes and have decided to keep the 66.5° value in the PR. 

The attached map shows the current latitude of 45°, 66° (standing in for the proposed 66.5°), and 60° as an alternative. 

Automatically switching to a polar CRS at 60° may be more convenient for analyses in Greenland or Iceland, but switching to a polar CRS for the large land regions between 60° and 66.5° in Europe and Russia may be less convenient given their orientation to the sides and top of the map.

image

PR #1706, opened 2024-01-08
EDSC-3951: Update GraphQL queries to use new `params` parameter objects

Bccorb:
# Overview

### What is the feature?

Updated some lingering graphQL requests to use the `params` attribute with supported types.

### What is the Solution?

Updated GraphQL queries and calls.

Impacts Project view, main search view, and focused collection view.

# Testing

### Reproduction steps

- Performed outline regression tests in compared by Prod and Dev.

### Attachments

None.

# Checklist

## SIDEBAR - SEEKING FEEDBACK
- I had to remove hasGranuleCounts from the query as that wasn't a type of either param. I want to have someone take a look at that in detail and let me know, if there is a way to add that back or is it depreciated anyway?


dpesall (approved):

static/src/js/actions/exportSearch.js
```diff
@@ -132,78 +132,8 @@ export const exportSearchAsStac = (format) => (dispatch, getState) => {
...
-        serviceType: $serviceType
-        sortKey: $sortKey
-        spatialKeyword: $spatialKeyword
-        tagKey: $tagKey
-        temporal: $temporal
-        twoDCoordinateSystemName: $twoDCoordinateSystemName
-        limit: $limit,
-        cursor: $cursor
-      ) {
+    query SearchCollections($params: CollectionsInput) {
```
eudoroolivares2016, line 135: this function might be OBE I don't see any way to access it on the controller and its not called anywhere else

serverless/src/retrieveGranuleLinks/fetchCmrLinks.js
```diff
@@ -8,58 +8,8 @@ import { getS3Urls } from '../../../sharedUtils/getS3Urls'
...
-      options: $options
-      orbitNumber: $orbitNumber
-      point: $point
-      polygon: $polygon
-      readableGranuleName: $readableGranuleName
-      sortKey: $sortKey
-      temporal: $temporal
-      twoDCoordinateSystem: $twoDCoordinateSystem
-    ) {
+  query GetGranuleLinks( $params: GranulesInput ) {
```
eudoroolivares2016, line 11: I am seeing an issue on this one where we can't get the granule downloads due to this causing a 500 error. To replicate: Find a collection add some granules to project use the Download option. Normally we get the granules links on the results page but, its erroring out

eudoroolivares2016 (changes requested):

eudoroolivares2016 (commented):
I added a commit that seems to fix the issue locally. We needed to wrap the request in a `params` object but, also there were issues calling the `(prepareGranuleAccessParams(granuleParams)` because the `pageSize` field is not defined for granules in graphql. See https://github.com/nasa/cmr-graphql/blob/main/src/types/granule.graphql

eudoroolivares2016 (changes requested):
Blocking step as I need to update the sprint in bamboo after merges get done

PR #1714, opened 2024-02-06
EDSC-3947: Fix new and qualitative colormaps/legends

rushgeo:
# Overview

### What is the feature?
There are two issues:

1. EDSC does not render colormaps (legends) for qualitative data. 
2. EDSC has not been running its scheduled task to process colormaps from GIBS and store them in its database since 2021. This means any new colormaps created since then will fail to load.

### What is the Solution?

1. Changes to `Legend` files allow populating the legend with qualitative classes instead of skipping it
    - Quantitative legends label the first and last value. For qualitative legends, we've removed those labels and instead prompt the user to hover over the classes to see their labels (see screenshot below). 
2. ~~Changes to `aws-functions` enable the colormap processing by default~~
   - The environment variables in CI still need to either be set to `true` ~~or removed~~.

EDSC legend component and EDSC colormap lambdas

# Testing

### Reproduction steps

- **Environment for testing:** PROD
- **Collection to test with:** `C2617126679-POCLOUD`, `C2812413911-ASIPS`

1. To test locally, your development database will need a recent copy of the colormaps. Download [colormaps_2024-02-05.txt](https://github.com/nasa/earthdata-search/files/14185915/colormaps_2024-02-05.txt), then run `psql -d edsc_dev < colormaps_2024-02-05.txt`
2. Search for the OPERA collection `C2617126679-POCLOUD`, or use [this link](http://localhost:8080/search/granules?p=C2617126679-POCLOUD&pg[0][v]=f&pg[0][gsk]=-start_date&q=C2617126679-POCLOUD&sb[0]=-74.31152%2C40.19547%2C-72.11426%2C41.17057&qt=2023-08-18T18%3A11%3A04.620Z%2C2023-08-21T20%3A32%3A56.293Z&tl=1690117225!3!!&lat=39.9287109375&long=-74.9443359375&zoom=7) to show the same granules in the screenshot below.
    - Note that because this colormap is newer than 2021, it is currently missing in production, so searching for this collection on search.earthdata.nasa.gov will result in a red error banner saying the colormap wasn't found.
3. This is a *qualitative* colormap, so ensure the legend appears and has the text "Hover for class names"
4. Hover over each of the five classes and see their values are: Not Water, Open Water, Partial Surface Water, Snow/Ice, and Cloud
5. The attached colormaps table includes new colormaps since 2021. To verify that new *quantitative* colormaps also work, search for `C2812413911-ASIPS`, and ensure a legend with a colormap from yellow to red appears, with values ranging from < 0.000 to 5.000.

### Attachments

image

# Checklist

### DIFF
--- a/serverless-configs/aws-functions.yml
+++ b/serverless-configs/aws-functions.yml
@@ -164,7 +164,7 @@
     events:
       - schedule:
           rate: cron(45 12 ? * MON-FRI *)
-          enabled: ${env:GIBS_JOB_ENABLED, false}
+          enabled: ${env:GIBS_JOB_ENABLED, file(./serverless.yml):custom.enableGenerateColorMaps.${file(./serverless.yml):provider.stage}, true}
 
   #
   # API Gateway Endpoints
--- a/serverless-configs/aws-functions.yml
+++ b/serverless-configs/aws-functions.yml
@@ -140,22 +140,22 @@
           rate: cron(0 12 ? * MON-FRI *)
           input:
             projection: 'epsg4326'
-          enabled: ${env:COLORMAP_JOB_ENABLED, true}
+          enabled: ${env:COLORMAP_JOB_ENABLED, file(./serverless.yml):custom.enableGenerateColorMaps.${file(./serverless.yml):provider.stage}, true}
       - schedule:
           rate: cron(5 12 ? * MON-FRI *)
           input:
             projection: 'epsg3857'
-          enabled: ${env:COLORMAP_JOB_ENABLED, true}
+          enabled: ${env:COLORMAP_JOB_ENABLED, file(./serverless.yml):custom.enableGenerateColorMaps.${file(./serverless.yml):provider.stage}, true}
       - schedule:
           rate: cron(10 12 ? * MON-FRI *)
           input:
             projection: 'epsg3413'
-          enabled: ${env:COLORMAP_JOB_ENABLED, true}
+          enabled: ${env:COLORMAP_JOB_ENABLED, file(./serverless.yml):custom.enableGenerateColorMaps.${file(./serverless.yml):provider.stage}, true}
       - schedule:
           rate: cron(15 12 ? * MON-FRI *)
           input:
             projection: 'epsg3031'
-          enabled: ${env:COLORMAP_JOB_ENABLED, true}
+          enabled: ${env:COLORMAP_JOB_ENABLED, file(./serverless.yml):custom.enableGenerateColorMaps.${file(./serverless.yml):provider.stage}, true}
 
   generateGibsTags:
     handler: serverless/src/generateGibsTags/handler.default
--- a/serverless-configs/aws-functions.yml
+++ b/serverless-configs/aws-functions.yml
@@ -140,22 +140,22 @@
           rate: cron(0 12 ? * MON-FRI *)
           input:
             projection: 'epsg4326'
-          enabled: ${env:COLORMAP_JOB_ENABLED, file(./serverless.yml):custom.enableGenerateColorMaps.${file(./serverless.yml):provider.stage}, true}
+          enabled: ${env:COLORMAP_JOB_ENABLED, false}
       - schedule:
           rate: cron(5 12 ? * MON-FRI *)
           input:
             projection: 'epsg3857'
-          enabled: ${env:COLORMAP_JOB_ENABLED, file(./serverless.yml):custom.enableGenerateColorMaps.${file(./serverless.yml):provider.stage}, true}
+          enabled: ${env:COLORMAP_JOB_ENABLED, false}
       - schedule:
           rate: cron(10 12 ? * MON-FRI *)
           input:
             projection: 'epsg3413'
-          enabled: ${env:COLORMAP_JOB_ENABLED, file(./serverless.yml):custom.enableGenerateColorMaps.${file(./serverless.yml):provider.stage}, true}
+          enabled: ${env:COLORMAP_JOB_ENABLED, false}
       - schedule:
           rate: cron(15 12 ? * MON-FRI *)
           input:
             projection: 'epsg3031'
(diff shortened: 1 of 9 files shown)


static/src/js/components/Legend/Legend.js
```diff
@@ -75,12 +78,23 @@ export const Legend = ({
...
+  let maxLabel
+
+  const qualitative = Object.keys(classes).length > 0
+  const hoverPrompt = 'Hover for class names'
+  if (qualitative) {
+    ({ colors = [], labels = [] } = classes)
+    minLabel = hoverPrompt
+    maxLabel = null
+  } else {
+    [minLabel] = labels
```
eudoroolivares2016, line 95:
```suggestion
    minLabel = labels[0]
```
nitpick; I know what you currently have does the same thing this is just a preference I have in this case to match the line below more clearly

serverless-configs/aws-functions.yml
```diff
@@ -140,22 +140,22 @@
           rate: cron(0 12 ? * MON-FRI *)
           input:
             projection: 'epsg4326'
-          enabled: ${env:COLORMAP_JOB_ENABLED, false}
+          enabled: ${env:COLORMAP_JOB_ENABLED, true}
```
eudoroolivares2016, line 143: My preference would be to leave these as false because if the variable is removed from bamboo it would still be set to true which I think is more confusing and harder to track

eudoroolivares2016 (approved):

static/src/js/components/Legend/Legend.js
rushgeo: Oh I totally agree. This is one of the sillier ESLint rules we might consider disabling globally.

serverless-configs/aws-functions.yml
rushgeo: I am in favor of it being true:
- The intended design is for these lambdas to run, so the default should be true. Anyone deploying EDSC needs these lambdas to run, or they'll get an error banner for every colormap loading attempt.
- The ability to disable the lambdas wasn't added until there was a need to disable colormap support for a while, but that was a temporary need we don't anticipate having again.
- The code is more accessible than Bamboo, so it's an easier place to check behavior. Generally I would like to see as few environment variables in routine use as possible.

macrouch: With these set to true does that mean that the jobs will run on these schedules in your dev environment? I think the answer to that is yes.

_If_ the answer is yes:
Is there any reason to have these jobs running locally on a schedule? No, and the less things you can have running locally the better
Can these jobs run in your local environment without errors? I think SQS is involved, so that would not work locally.

static/src/js/components/Legend/Legend.js
eudoroolivares2016: That might be an issue on your configuration in vscode.. eslint isn't complaining if you do it either way, otherwise this ci build would fail.

serverless-configs/aws-functions.yml
rushgeo: I don't think they can run locally on a schedule without additional serverless plugins. So there's no risk of them running on a schedule locally.

static/src/js/components/Legend/Legend.js
rushgeo: I'll need to figure out why the VSCode ESLint plugin isn't using the same configuration as the project. I'll push a fix to switch it to the more readable version.

serverless-configs/aws-functions.yml
macrouch: I switched the first schedule to this and set the default to true
```
rate: cron(0 * * * MON-FRI *)
```

and it ran with errors
```
Server ready: http://localhost:3001 🚀
Found 1086 layers within GIBS
sending message {"id":897,"product":"AMSR2_Cloud_Liquid_Water_Day","url":"https://gibs.earthdata.nasa.gov/colormaps/v1.3/AMSR2_Cloud_Liquid_Water_Day.xml"}
InvalidClientTokenId: The security token included in the request is invalid.
(λ: generateColorMaps) RequestId: 1117578b-2c9c-4a61-bf7d-46dac5a1731a  Duration: 1726.52 ms  Billed Duration: 1727 ms
Successfully invoked scheduled function: [generateColorMaps]
Completed Capabilities request with status 500.
```

static/src/js/components/Legend/Legend.js
rushgeo: ESLint failed after pushing the change

eudoroolivares2016: I see I was wrong. Its because the var isn't a const. I'll dismiss this. just use the eslint preference

serverless-configs/aws-functions.yml
rushgeo: Thanks for the heads-up! I thought that they wouldn't run because I saw serverless plugins for that purpose, but apparently at some point after we disabled these lambdas in 2021, `serverless offline` added support for scheduled lambdas.

I just pushed a commit that prevents them from being scheduled in the local dev environment.

rushgeo:
I've also re-enabled the GIBS tag job by default. We will also need to set their Bamboo environment variables to true or delete them in SIT and UAT. It was already set to true in PROD.

The GIBS and ColorMaps jobs now include a somewhat verbose lookup to keep them from being scheduled if we're running in the local dev environment. 

To test how they evaluate, you can use the following commands:
```sh
serverless print --path functions.generateColorMaps
serverless print --path functions.generateGibsTags
```
And see that they are enabled if any value other than `dev` is set as the default in `serverless.yml:provider.stage`.

serverless-configs/aws-functions.yml
eudoroolivares2016: I'm sure this solution can work but, it does seem really overcomplicated, having it just left as `false` seems like the correct solution here and consistent with the way the rest of our jobs worked

eudoroolivares2016 (changes requested):
I think we should leave the `serverless-configs/aws-functions.yml` with the default value as false this current solution overcomplicates things

eudoroolivares2016 (approved):

bnp26 (commented):
Looks good to me.

dpesall (approved):

PR #1718, opened 2024-02-21
EDSC-4000: Install GSA's Digital Analytics Program (DAP) tracking code

bnp26:
# Overview

### What is the feature?

Added dap script

### What is the Solution?

Summarize what you changed.

# Testing

### Reproduction steps

ensure https://dap.digitalgov.gov/Universal-Federated-Analytics-Min.js?agency=NASA&subagency=GSFC is in the network tab.
### Attachments

# Checklist

### DIFF
--- a/static/src/partials/body.html
+++ b/static/src/partials/body.html
@@ -126,4 +126,5 @@
 <% } %>
 <script async type="text/javascript"
     src="https://dap.digitalgov.gov/Universal-Federated-Analytics-Min.js?agency=NASA&subagency=GSFC"
-    id="_fed_an_ua_tag"></script>
+    id="_fed_an_ua_tag">
+</script>
--- a/static/src/partials/body.html
+++ b/static/src/partials/body.html
@@ -124,7 +124,9 @@
 ></script>
 <script type="text/javascript">feedback.init({ showIcon: false });</script>
 <% } %>
+<% if (env != 'dev') { %>
 <script async type="text/javascript"
     src="https://dap.digitalgov.gov/Universal-Federated-Analytics-Min.js?agency=NASA&subagency=GSFC"
     id="_fed_an_ua_tag">
 </script>
+<% } %>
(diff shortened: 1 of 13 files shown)


static/src/partials/wrapper.html
```diff
@@ -11,8 +11,13 @@
...
   <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
   <meta name="viewport" content="width=device-width, initial-scale=1" />
-  <meta name="description" content="Search, discover, visualize, refine, and access NASA Earth Observation data in your browser with Earthdata Search" data-react-helmet="true" />
+  <meta name="description"
+    content="Search, discover, visualize, refine, and access NASA Earth Observation data in your browser with Earthdata Search"
+    data-react-helmet="true" />
   <link rel="shortcut icon" href="/favicon.ico?v=1.1" />
+  <script async type="text/javascript"
+    src="https://dap.digitalgov.gov/Universal-Federated-Analytics-Min.js?agency=NASA&subagency=GSFC"
+    id="_fed_an_ua_tag"></script>
```
dpesall, line 20: Closing `</script>` tag should be on a newline since it has a multi-line body.

trevorlang: This script tag should go in after the body if possible. Otherwise it could cause some blocking. It probably makes more sense to throw it in the `body.html` partial near the feedback app. Seems like this is something we might only want to run when not in the `development` env too.

dpesall (approved):

static/src/partials/body.html
```diff
@@ -124,3 +124,7 @@ <h3 class="fallback-page__heading">Your browser is out of date</h3>
 ></script>
 <script type="text/javascript">feedback.init({ showIcon: false });</script>
 <% } %>
+<script async type="text/javascript"
```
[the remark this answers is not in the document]
eudoroolivares2016: It was mentioned during standup that it was okay for this to be on during development mode.

eudoroolivares2016 (changes requested):
If we look at what we did for google analytics I think that is really what we would be looking to replicate here or similar to the feedback app. It would be preferred that we didn't have this script running while in development mode because otherwise 1 we are going to pollute their metrics and we have additional scripts being worked during development mode

eudoroolivares2016 (approved):
I'm okay with using `env` here I probably think it should be its own variable

PR #1719, opened 2024-02-28
EDSC-4002: added retrieveVariablesRequest and added it to focusedCollection.js and project.js

bnp26:
# Overview

### What is the feature?

Allow for variables to be fully returned for collection requests.

### What is the Solution?

Added the function retrieveVariablesRequest and added it to focusedCollections.js and project.js
which goes through and returns all associated variables for collections.

List impacted areas.

# Testing

### Reproduction steps

- **Environment for testing:**UAT
- **Collection to test with:**C1241484650-LPCLOUDUAT

1. navigate to localhost:8080/search?ee=uat
2. search for C1241484650-LPCLOUDUAT
3. Click on the collection
4. Ensure that everything loads properly and if possible, use Redux chrome extension to make sure all 4000+ variables are added when the UPDATE_COLLECTION_METADATA is dispatched.
5. add the collection to the project and click Download
6. Ensure that everything loads properly and if possible, use Redux chrome extension to make sure all 4000+ variables are added when the UPDATE_COLLECTION_METADATA is dispatched.

### Attachments

*Nothing changes on the frontend

# Checklist


static/src/js/actions/focusedCollection.js
```diff
@@ -180,130 +184,155 @@ export const getFocusedCollection = () => async (dispatch, getState) => {
...
+        versionId
+      } = collection
 
-        // Formats the metadata returned from graphql for use throughout the application
-        const focusedMetadata = createFocusedCollectionMetadata(
-          collection,
-          authToken,
-          earthdataEnvironment
+      // Retrieves all variables if there are more than `maxCmrPageSize`
+      if (variables && variables.count && variables.count > maxCmrPageSize) {
```
eudoroolivares2016, line 240: any way we can simplify this conditional? From what I can see should just be able to use two assertions since if there are variables from the response it will include count
```suggestion
      if (variables && variables.count > maxCmrPageSize) {
```

static/src/js/actions/project.js
```diff
@@ -338,131 +343,160 @@ export const getProjectCollections = () => async (dispatch, getState) => {
...
-      pageSize: filteredIds.length
-    },
-    subcriptionParams: {
-      subscriberId: username
-    }
-  })
-    .then((responseObject) => {
-      const payload = []
+  try {
+    response = await graphQlRequestObject.search(graphQuery, {
```
eudoroolivares2016, line 349:
```suggestion
    const response = await graphQlRequestObject.search(graphQuery, {
```

static/src/js/actions/project.js
```diff
@@ -338,131 +343,160 @@ export const getProjectCollections = () => async (dispatch, getState) => {
...
-    })
+  } catch (error) {
+    dispatch(actions.handleError({
+      error,
+      action: 'getProjectCollections',
+      resource: 'project collections'
+    }))
+  }
 
   return response
```
eudoroolivares2016, line 501: Can this response be `null` (the initialized value?)

static/src/js/actions/focusedCollection.js
```diff
@@ -180,130 +184,155 @@ export const getFocusedCollection = () => async (dispatch, getState) => {
...
+        variables.items = await retrieveVariablesRequest(
+          variables,
+          {
+            params: {
+              conceptId: focusedCollectionId,
+              includeHasGranules: true,
+              includeTags: defaultCmrSearchTags.join(',')
```
eudoroolivares2016, line 247: You don't need to have the tags be part of the subsequent variable requests, those just add more metadata but, you aren't parsing any of it in your `retrieveVariablesRequest` functions

static/src/js/actions/focusedCollection.js
```diff
@@ -180,130 +184,155 @@ export const getFocusedCollection = () => async (dispatch, getState) => {
         }
       }
     }`
+  try {
+    const response = await graphQlRequestObject.search(graphQuery, {
+      params: {
+        conceptId: focusedCollectionId,
+        includeHasGranules: true,
+        includeTags: defaultCmrSearchTags.join(',')
```
eudoroolivares2016, line 192: You don't need to have the tags be part of the subsequent variable requests, those just add more metadata but, you aren't parsing any of it in your `retrieveVariablesRequest` functions

eudoroolivares2016 (changes requested):

static/src/js/actions/focusedCollection.js
bnp26: which ones would you like me to remove?

static/src/js/actions/project.js
bnp26: I set it as a let in line 346 and then return it at the end of everything

static/src/js/actions/project.js
bnp26: That's what we do in line 346

static/src/js/actions/focusedCollection.js
eudoroolivares2016: Oh I just mean that when you pass that parameter: `includeTags ` to graphql/CMR you are asking to, alongside the standard metadata, also get the `tag` concepts associated to that collection. If you aren't parsing those tag concepts then you don't need to request them

static/src/js/actions/project.js
eudoroolivares2016: Yea that's what I'm wondering so if I'm reading correct, we are only going to return `null` if the try fails and we catch an error. If that is the case then I'd prefer we don't have a let but, instead like in `focusedCollection`
we just return `null` on line 501 makes it a easier to trace

static/src/js/util/retrieveVariablesRequest.js
```diff
@@ -0,0 +1,89 @@
...
+    const {
+      data: variablesData
+    } = results
+
+    const { data: pagedData } = variablesData
+    const formattedData = formatCollectionObject(pagedData)
+
+    const { collection: pagedCollection } = formattedData
+    const { variables: pagedVariables } = pagedCollection
+    const { cursor: newCursor, items: vars } = pagedVariables
```
daniel-zamora, line 81: Why not just have this returned by formatCollectionObject, instead of pulling it out of the returned value?

static/src/js/util/retrieveVariablesRequest.js
```diff
@@ -0,0 +1,89 @@
...
+
+    collectedItems = collectedItems.concat(vars)
+    if (newCursor === nextCursor) break
```
daniel-zamora, line 84: Is this for an edge case? I'm not sure I see the point of a break here.

static/src/js/util/retrieveVariablesRequest.js
```diff
@@ -0,0 +1,89 @@
...
+  graphQlRequestObject
+) => {
+  // Pull out cursor and already retrieved items from the variablesObj
+  const { cursor, items } = variablesObj
+
+  let collectedItems = items
+  let nextCursor = cursor
+  while (nextCursor !== null) {
+    const varsGraphQuery = `
+        query GetCollection(
```
daniel-zamora, line 43: Are there alternatives to using the GetCollection query? Can you just directly query for variables instead?

macrouch: Great catch, you absolutely should query variables directly. As is this is doubling the amount of queries that hit CMR from what you actually need

eudoroolivares2016: Can we do that? We need a reference for which variables to get `cmr-graphql` doesn't allow us to query variables by the collection concept-id. CMR itself doesn't support that in the documentation either `https://cmr.earthdata.nasa.gov/search/site/docs/search/api.html#variable`

static/src/js/util/retrieveVariablesRequest.js
eudoroolivares2016: Theres a branching statement in `formatCollectionObject` so if we moved this there wouldn't we have to do it twice for both cases?

static/src/js/util/retrieveVariablesRequest.js
eudoroolivares2016: I just tried with `keyword` that does work we should use that to search for the variables directly

static/src/js/util/retrieveVariablesRequest.js
bnp26: If you could take a look at this again after the changes to the query, it should be much cleaner now.

eudoroolivares2016 (approved):

daniel-zamora (approved):

PR #1730, opened 2024-03-22
EDSC-4030: Nullify the retrieval_collections foreign key constraint b…

eudoroolivares2016:
# Overview

### What is the feature?

Due to the update to the database where we encrypted/created a new database instance something changed in either the `knex` commands (which would normally be skipped during the migrateDatabases lambda if the tables/schema already existed) or in the database version that we had to update our RDS instance to (the old one was not available). that allowed this constraint which should have always existed but, simply did not to be created

### What is the Solution?

By nullifying the foreign key before we do the deletion command we are able to get around this introduced constraint

List impacted areas.

# Testing

1) Create a retrieval ensure that it can be deleted from the history page; if you have historical retrievals ensure we can delete some of those as well

2) Ensure that we can create a project and delete that project

### Attachments

# Checklist

### DIFF
--- a/serverless/src/deleteRetrieval/handler.js
+++ b/serverless/src/deleteRetrieval/handler.js
@@ -37,7 +37,7 @@ export default async function deleteRetrieval(event, context) {
 
     // Nullify the foreign key constraint on the table before deleting
     await dbConnection('retrieval_collections')
-      .where('retrieval_id', '=', decodedRetrievalId)
+      .where('retrieval_id', decodedRetrievalId)
       .update('retrieval_id', null)
 
     // Retrieve the authenticated users' id to ensure the retrieval being deleted belongs to them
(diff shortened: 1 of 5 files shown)


serverless/src/deleteRetrieval/handler.js
```diff
@@ -35,6 +35,11 @@ export default async function deleteRetrieval(event, context) {
     // Retrieve a connection to the database
     const dbConnection = await getDbConnection()
 
+    // Nullify the foreign key constraint on the table before deleting
+    await dbConnection('retrieval_collections')
+      .where('retrieval_id', '=', decodedRetrievalId)
```
abbottry, line 40: For future self, `.where('retrieval_id', decodedRetrievalId)` and is a bit less typing, I would typically only use `=` if the = was something that wasn't the default, like `>` or something.

abbottry (approved):

PR #1731, opened 2024-03-26
EDSC-3717: Fix shape searching cursor remaining if you cancel creating a shape

eudoroolivares2016:
# Overview

### What is the feature?

Minor update so that if a user does not commit to creating a `shapefile` and instead hits remove for rectangular and circular subsetting the cursor for creating the shapefile is not maintained

### What is the Solution?

Added an event emission so that the cursor is removed and the leaflet state is updated added referance to disable the polygon from the current toolbars, otherwise when the `eventEmitter.emit('map.drawCancel')` was called it was not removing it if the user had selected polygon subsetting and canceled

Shapefile subsetting when canceling (really when using the remove button before having made a shapefile) for circles and rectangles

# Testing

Ensure that when you start trying to subset by rectangle or circle if you hit remove the cursor for the shapefile is no longer maintained

### Attachments

# Checklist

### DIFF
--- a/static/src/js/components/SpatialDisplay/SpatialDisplay.js
+++ b/static/src/js/components/SpatialDisplay/SpatialDisplay.js
@@ -274,6 +274,7 @@ const SpatialDisplay = ({
     setManuallyEnteringVal(false)
 
     onRemoveSpatialFilter()
+    eventEmitter.emit('map.drawCancel')
   }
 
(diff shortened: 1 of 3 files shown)


bnp26 (approved):
Looks good

dpesall (approved):

PR #1739, opened 2024-04-25
EDSC-4050: Added ability to generate L2 Raster Products using SWODLR in EDSC.

bnp26:
# Overview

### What is the feature?

Have an endpoint/lambda to process SWODLR orders in earthdata search.

### What is the Solution?

Added the endpoint/lambda to process SWODLR orders in earthdata search.

List impacted areas.

# Testing

### Reproduction steps

- **Collection to test with:** C2799438271-POCLOUD

1. go to localhost:8080/search/granules?ee=prod&p=C2799438271-POCLOUD
2. Add the 2 granules  
  1. G2938391118-POCLOUD (SWOT_L2_HR_Raster_250m_UTM34L_N_x_x_x_013_514_087F_20240414T225546_20240414T225558_PIC0_01)
  2. G2938390924-POCLOUD (SWOT_L2_HR_Raster_250m_UTM34L_N_x_x_x_013_514_086F_20240414T225526_20240414T225547_PIC0_01)
4. Update the db entry in *retrieval_orders* and change the values to match
`granule_params: {
    exclude: {},
    options: {},
    page_num: 1,
    sort_key: '-start_date',
    json_data: {
      params: {
        rasterResolution: 6,
        outputSamplingGridType: 'GEO',
        outputGranuleExtentFlag: false
      },
      custom_params: {
        'G2938391118-POCLOUD': {
          utmZoneAdjust: null,
          mgrsBandAdjust: null
        },
        'G2938390924-POCLOUD': {
          utmZoneAdjust: null,
          mgrsBandAdjust: null
        }
      }
    },
    page_size: 2000,
    concept_id: [
      'G2938390910-POCLOUD',
      'G2938390924-POCLOUD'
    ],
    echo_collection_id: 'C2799438271-POCLOUD',
    two_d_coordinate_system: {}
  }`

5. You will also need to edit the url of the service used in the `retrieval_collections` table under the `access_method` column

6. check https://github.com/nasa/earthdata-search/pull/1739/files#diff-93cc28b3641d217b6bb7b578b51ca669f0861b5aa021f358de158e14b11da700 and save it with the id equal to the *retrieval_orders* entry id.

7. run `npm run invoke-local -- --function submitSwodlrOrder -p /Users/bporeh/gsfc/earthdata-search/payloads/mock_swodlr_lambda_1.json` switching the mock_swodlr_lambda_1.json to the file referenced in step 

8. ensure the lambda runs all the way.

9. Subsequent requests to get the status can be sent to the swoldr API along with the productId that it generates in the response

### Attachments

# Checklist

### DIFF
--- a/serverless-configs/aws-functions.yml
+++ b/serverless-configs/aws-functions.yml
@@ -123,7 +123,7 @@
       - sqs:
           arn:
             Fn::GetAtt:
-              - HarmonyOrderQueue
+              - SwodlrOrderQueue
               - Arn
     environment:
       updateOrderStatusStateMachineArn: ${self:resources.Outputs.UpdateOrderStatusWorkflow.Value}
--- a/serverless-configs/aws-functions.yml
+++ b/serverless-configs/aws-functions.yml
@@ -116,6 +116,18 @@
     environment:
       updateOrderStatusStateMachineArn: ${self:resources.Outputs.UpdateOrderStatusWorkflow.Value}
 
+  submitSwodlrOrder:
+    handler: serverless/src/submitSwodlrOrder/handler.default
+    timeout: 600
+    events:
+      - sqs:
+          arn:
+            Fn::GetAtt:
+              - HarmonyOrderQueue
+              - Arn
+    environment:
+      updateOrderStatusStateMachineArn: ${self:resources.Outputs.UpdateOrderStatusWorkflow.Value}
+
   fetchCatalogRestOrder:
     handler: serverless/src/fetchCatalogRestOrder/handler.default
     timeout: ${env:LAMBDA_TIMEOUT, '30'}
--- a/serverless/src/submitSwodlrOrder/__tests__/constructOrderPayload.test.js
+++ b/serverless/src/submitSwodlrOrder/__tests__/constructOrderPayload.test.js
@@ -1,0 +1,143 @@
+import nock from 'nock'
+
+import * as getEarthdataConfig from '../../../../sharedUtils/config'
+
+import { constructOrderPayload } from '../constructOrderPayload'
+
+describe('constructOrderPayload', () => {
+  beforeEach(() => {
+    jest.spyOn(getEarthdataConfig, 'getEarthdataConfig').mockImplementation(() => ({
+      cmrHost: 'https://cmr.earthdata.nasa.gov'
+    }))
+  })
+
+  test('constructs an order payload', async () => {
+    nock(/cmr/)
+      .matchHeader('Authorization', 'Bearer access-token')
+      .get(/search\/granules.json/)
+      .reply(200, {
+        feed: {
+          entry: [{
+            id: 'G2938391118-POCLOUD',
+            title: 'SWOT_L2_HR_Raster_250m_UTM34L_N_x_x_x_013_514_087F_20240414T225546_20240414T225558_PIC0_01'
+          }, {
+            id: 'G2938390924-POCLOUD',
+            title: 'SWOT_L2_HR_Raster_250m_UTM34L_N_x_x_x_013_514_086F_20240414T225526_20240414T225547_PIC0_01'
+          }
+          ]
+        }
+      })
+
+    const collectionConceptId = 'C2799438271-POCLOUD'
+
+    // Const accessMethod = {
+    //   optionDefinition: {
+    //     conceptId: 'C2799438271-POCLOUD',
+    //     name: 'SWOT Level 2 Water Mask Raster Image Data Product, Version 2.0'
+    //   },
+    //   url: 'https://swodlr.podaac.earthdatacloud.nasa.gov',
+    //   type: 'SWODLR'
+    // }
+
+    const granuleParams = {
+      collection_concept_id: 'C2799438271-POCLOUD'
+    }
+
+    const accessToken = 'access-token'
+    const earthdataEnvironment = 'prod'
+
+    const response = await constructOrderPayload({
+      collectionConceptId,
+      earthdataEnvironment,
+      accessToken,
+      granuleParams
+    })
+
+    expect(response).toEqual({
+      collectionConceptId,
+      orderItems: [{
+        granuleConceptId: 'G2938391118-POCLOUD',
+        granuleUr: 'SWOT_L2_HR_Raster_250m_UTM34L_N_x_x_x_013_514_087F_20240414T225546_20240414T225558_PIC0_01'
+      },
+      {
+        granuleConceptId: 'G2938390924-POCLOUD',
+        granuleUr: 'SWOT_L2_HR_Raster_250m_UTM34L_N_x_x_x_013_514_086F_20240414T225526_20240414T225547_PIC0_01'
+      }
+      ]
+    })
+  })
+
+  test('constructs an order payload with added granule params', async () => {
+    nock(/cmr/)
+      .matchHeader('Authorization', 'Bearer access-token')
+      .get(/search\/granules.json/)
+      .reply(200, {
+        feed: {
+          entry: [{
+            id: 'G2938391118-POCLOUD',
+            title: 'SWOT_L2_HR_Raster_250m_UTM34L_N_x_x_x_013_514_087F_20240414T225546_20240414T225558_PIC0_01'
+          }, {
+            id: 'G2938390924-POCLOUD',
+            title: 'SWOT_L2_HR_Raster_250m_UTM34L_N_x_x_x_013_514_086F_20240414T225526_20240414T225547_PIC0_01'
+          }
+          ]
+        }
+      })
+
+    const collectionConceptId = 'C2799438271-POCLOUD'
+
+    const granuleParams = {
+      exclude: {},
+      options: {},
+      page_num: 1,
+      sort_key: '-start_date',
+      json_data: {
+        params: {
+          rasterResolution: 6,
+          outputSamplingGridType: 'GEO',
+          outputGranuleExtentFlag: false
+        },
+        custom_params: {
+          'G2938391118-POCLOUD': {
+            utmZoneAdjust: null,
+            mgrsBandAdjust: null
+          },
+          'G2938390924-POCLOUD': {
+            utmZoneAdjust: null,
+            mgrsBandAdjust: null
+          }
+        }
+      },
+      page_size: 2000,
+      concept_id: [
+        'G2938390910-POCLOUD',
+        'G2938390924-POCLOUD'
+      ],
+      echo_collection_id: 'C2799438271-POCLOUD',
+      two_d_coordinate_system: {}
+    }
+
+    const accessToken = 'access-token'
+    const earthdataEnvironment = 'prod'
+
+    const response = await constructOrderPayload({
+      accessToken,
+      collectionConceptId,
+      earthdataEnvironment,
+      granuleParams
+    })
+
+    expect(response).toEqual({
+      collectionConceptId,
+      orderItems: [{
+        granuleConceptId: 'G2938391118-POCLOUD',
+        granuleUr: 'SWOT_L2_HR_Raster_250m_UTM34L_N_x_x_x_013_514_087F_20240414T225546_20240414T225558_PIC0_01'
+      },
+      {
+        granuleConceptId: 'G2938390924-POCLOUD',
+        granuleUr: 'SWOT_L2_HR_Raster_250m_UTM34L_N_x_x_x_013_514_086F_20240414T225526_20240414T225547_PIC0_01'
+      }
+      ]
+    })
+  })
+})
--- a/serverless/src/submitSwodlrOrder/__tests__/constructOrderPayload.test.js
+++ b/serverless/src/submitSwodlrOrder/__tests__/constructOrderPayload.test.js
@@ -30,15 +30,6 @@ describe('constructOrderPayload', () => {
 
     const collectionConceptId = 'C2799438271-POCLOUD'
 
-    // Const accessMethod = {
-    //   optionDefinition: {
-    //     conceptId: 'C2799438271-POCLOUD',
-    //     name: 'SWOT Level 2 Water Mask Raster Image Data Product, Version 2.0'
-    //   },
-    //   url: 'https://swodlr.podaac.earthdatacloud.nasa.gov',
-    //   type: 'SWODLR'
-    // }
-
     const granuleParams = {
       collection_concept_id: 'C2799438271-POCLOUD'
     }
--- a/serverless/src/submitSwodlrOrder/__tests__/handler.test.js
+++ b/serverless/src/submitSwodlrOrder/__tests__/handler.test.js
@@ -308,6 +308,161 @@ describe('submitSwodlrOrder', () => {
     expect(queries[1].bindings).toEqual(['create_failed', 'Unknown Error', 12])
   })
 
+  test('Doesnt submit download when there are too many granules', async () => {
+    nock(/cmr/)
+      .matchHeader('Authorization', 'Bearer access-token')
+      .get(/granules/)
+      .reply(200, {
+        feed: {
+          entry: [
+            {
+              id: 'G2938391118-POCLOUD1',
+              title: 'SWOT_L2_HR_Raster_250m_UTM34L_N_x_x_x_013_514_087F_20240414T225546_20240414T225558_PIC0_01'
+            }, {
+              id: 'G2938390924-POCLOUD2',
+              title: 'SWOT_L2_HR_Raster_250m_UTM34L_N_x_x_x_013_514_086F_20240414T225526_20240414T225547_PIC0_01'
+            },
+            {
+              id: 'G2938391118-POCLOUD3',
+              title: 'SWOT_L2_HR_Raster_250m_UTM34L_N_x_x_x_013_514_087F_20240414T225546_20240414T225558_PIC0_01'
+            }, {
+              id: 'G2938390924-POCLOUD4',
+              title: 'SWOT_L2_HR_Raster_250m_UTM34L_N_x_x_x_013_514_086F_20240414T225526_20240414T225547_PIC0_01'
+            },
+            {
+              id: 'G2938391118-POCLOUD5',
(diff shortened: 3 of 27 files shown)


payloads/mock_swodlr_lambda.json
```diff
@@ -0,0 +1,12 @@
...
+    {
+      "body": "{\"accessToken\":\"eyJ0eXAiOiJKV1QiLCJvcmlnaW4iOiJFYXJ0aGRhdGEgTG9naW4iLCJzaWciOiJlZGxqd3RwdWJrZXlfb3BzIiwiYWxnIjoiUlMyNTYifQ.eyJ0eXBlIjoiVXNlciIsInVpZCI6ImJwb3JlaCIsImV4cCI6MTcxODc0MDM0MiwiaWF0IjoxNzEzNTU2MzQyLCJpc3MiOiJFYXJ0aGRhdGEgTG9naW4ifQ.yYq_RJItUaq5ezNhRF1_cii2hjOqatrBm2_H2BS5D9jSSSgIDVRvCA1o8jwOp2S8JVpn3wPkIET4iifg4YiX-EAME2WiS96X9SOPmkVrsY34q8NG6Aeto_qkdTrAphhUWh7RokAeLwsZ2fRxXvVVvM9y8YzECxlqZU9cm2ikMl0UJVfvTaFmGyVoo2DopyLo6yk3Ovr_oZ5uG6B6OhIbDTL2taCE6e3i6rxeT5c4hEB2axO3f84KHvjNjbGPeQNwqvWjoPgZAwn2CfgaiSCDAtfwWORg40INbVukgUxc-P0sWkcg63OcJZY12u2gPlRFaaCkaU8abiFM2WZEQgCKrA\",\"id\":20}",
+      "requestContext": {
+        "authorizer": {
+          "jwtToken": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ0b2tlbiI6eyJhY2Nlc3NfdG9rZW4iOiIyZThlOTk1ZTc1MTFjMmM2NjIwMzM2Nzk3YiIsInRva2VuX3R5cGUiOiJCZWFyZXIiLCJleHBpcmVzX2luIjozNjAwLCJyZWZyZXNoX3Rva2VuIjoiNWExMDg2NWVlYjNmNDRhODBhYjk1IiwiZW5kcG9pbnQiOiIvYXBpL3VzZXJzL2Vkc2MiLCJleHBpcmVzX2F0IjoiMjAxOS0wNi0xNlQwMTowMTo0OS41NDVaIn0sImlhdCI6MTU2MDY0MzMwOX0.Xn0SuAMdEcU8amhZM0YunyRQN-e3cVzjoo7qJBr-EwI"
+        }
+      }
+    }
+  ]
+}
```
eudoroolivares2016, line 12: add newline

serverless/src/submitSwodlrOrder/__tests__/constructOrderPayload.test.js
```diff
@@ -0,0 +1,143 @@
...
+            id: 'G2938390924-POCLOUD',
+            title: 'SWOT_L2_HR_Raster_250m_UTM34L_N_x_x_x_013_514_086F_20240414T225526_20240414T225547_PIC0_01'
+          }
+          ]
+        }
+      })
+
+    const collectionConceptId = 'C2799438271-POCLOUD'
+
+    // Const accessMethod = {
```
eudoroolivares2016, line 33: remove this

serverless/src/submitSwodlrOrder/handler.js
```diff
@@ -0,0 +1,190 @@
...
+        id
+        timestamp
+        state
+        reason
+    }
+  }
+}`
+
+/**
+ * Submits an order to CMR Ordering (cmr-ordering)
```
eudoroolivares2016, line 33:
```suggestion
 * Submits an order to Swodlr
```

serverless/src/submitSwodlrOrder/handler.js
```diff
@@ -0,0 +1,190 @@
...
+ * npm run invoke-local -- --function submitSwodlrOrder -p tmp/test_echo_order.json
```
eudoroolivares2016, line 34:
```suggestion
 * npm run invoke-local -- --function submitSwodlrOrder -p tmp/test_swodlr_order.json
```
its just an example we gitignore the tmp folder

serverless/src/submitSwodlrOrder/handler.js
```diff
@@ -0,0 +1,190 @@
...
+        const { id: jobId, state } = jobStatus[0]
+
+        await dbConnection('retrieval_orders').update({
+          order_number: productId,
+          order_information: { jobId },
+          state
+        }).where({ id })
+      })
+
+      // Not sure if we can parse errors from swodlr
```
eudoroolivares2016, line 170: We should try to give swoldr a bad request and see how we handle it

serverless/src/submitSwodlrOrder/constructOrderPayload.js
```diff
@@ -0,0 +1,46 @@
+import axios from 'axios'
+
+import { stringify } from 'qs'
+
+import { getEarthdataConfig } from '../../../sharedUtils/config'
+import { getClientId } from '../../../sharedUtils/getClientId'
+import { readCmrResults } from '../util/cmr/readCmrResults'
+
```
eudoroolivares2016, line 8:
```suggestion
/**
 * Construct the payload that we'll send to Swodlr to create this order
 */
```

serverless/src/submitSwodlrOrder/constructOrderPayload.js
```diff
@@ -0,0 +1,46 @@
...
+export const constructOrderPayload = async ({
+  collectionConceptId,
+  earthdataEnvironment,
+  accessToken,
+  granuleParams
+}) => {
+  const newGranuleParams = { ...granuleParams }
+
+  delete newGranuleParams.json_data
```
eudoroolivares2016, line 17:
```suggestion
  // Remove data to use for CMR query
  delete newGranuleParams.json_data
```

eudoroolivares2016: Just some kind of comment something different is fine its just kind of a weird line without context

serverless-configs/aws-functions.yml
```diff
@@ -116,6 +116,18 @@
...
       updateOrderStatusStateMachineArn: ${self:resources.Outputs.UpdateOrderStatusWorkflow.Value}
 
+  submitSwodlrOrder:
+    handler: serverless/src/submitSwodlrOrder/handler.default
+    timeout: 600
+    events:
+      - sqs:
+          arn:
+            Fn::GetAtt:
+              - HarmonyOrderQueue
```
eudoroolivares2016, line 126: We need to update this; swodlr should definitely have its own SQS queue

eudoroolivares2016 (changes requested):

serverless/src/submitSwodlrOrder/__tests__/mocks.js
```diff
@@ -0,0 +1,13 @@
+export const mockSwodlrOrder = {
```
eudoroolivares2016, line 1: I don't think we need/want to check in this file. On the unit tests we can just call the handler with an object as I think you are already doing, this can be used for testing in your `/tmp` file or something uncommitted to git

serverless/src/submitSwodlrOrder/handler.js
```diff
@@ -0,0 +1,190 @@
...
+        accessToken,
+        earthdataEnvironment
+      })
+
+      const { orderItems } = granuleInfo
+
+      await orderItems.forEachAsync(async (granule) => {
+        const { granuleUr, granuleConceptId } = granule
+
+        const splitTitle = granuleUr.split('_')
```
eudoroolivares2016, line 99: I really wanna talk to the swodlr team to make sure this is absolutely the only way its super odd that we have to do this to get this information out

serverless/src/submitSwodlrOrder/handler.js
```diff
@@ -0,0 +1,190 @@
...
+        const { data } = responseData
+        const { generateL2RasterProduct } = data
+
+        // ID Information
+        // product Id is the Id of the product that's being generated
+        // jobId is the Id of the job that is being run to generate the product
+        const { id: productId, status: jobStatus } = generateL2RasterProduct
+        const { id: jobId, state } = jobStatus[0]
+
+        await dbConnection('retrieval_orders').update({
```
eudoroolivares2016, line 163: Does this need to be outside of the foreach? Otherwise we are just overriding this value each granule

serverless/src/submitSwodlrOrder/constructOrderPayload.js
```diff
@@ -0,0 +1,49 @@
...
+
+import { stringify } from 'qs'
+
+import { getEarthdataConfig } from '../../../sharedUtils/config'
+import { getClientId } from '../../../sharedUtils/getClientId'
+import { readCmrResults } from '../util/cmr/readCmrResults'
+/**
+ * Construct the payload that we'll send to Swodlr to create this order
+ */
+export const constructOrderPayload = async ({
```
eudoroolivares2016, line 11: nitpick; Lets rename this method to something more descriptive of what its doing such as `retrieveCMRGranules.js` etc the order payload is really getting set in `handler` which is overriding the variables for the graphql call to their API

serverless/src/submitSwodlrOrder/handler.js
bnp26: It'll probably be a week until we can get a meeting set up.

serverless/src/submitSwodlrOrder/handler.js
bnp26: Good question. Let me think about it.

serverless/src/submitSwodlrOrder/__tests__/mocks.js
bnp26: I'm not sure what you mean. This object is critical for the test I think. We can make it a const but we have that sort of file in other submit tests.

eudoroolivares2016: No you are right I thought this wasn't getting used

serverless/src/submitSwodlrOrder/handler.js
bnp26: This is how they handle errors. Looks good.

serverless/src/submitSwodlrOrder/handler.js
```diff
@@ -0,0 +1,194 @@
...
+      const granuleInfo = await retrieveCMRGranules({
+        collectionConceptId,
+        granuleParams,
+        accessToken,
+        earthdataEnvironment
+      })
+
+      const { orderItems } = granuleInfo
+
+      if (orderItems.length > 10) {
```
eudoroolivares2016, line 96: Can we add a test case for this conditional

serverless/src/submitSwodlrOrder/handler.js
```diff
@@ -0,0 +1,194 @@
...
+        throw new Error('Too many granules')
+      }
+
+      await orderItems.forEachAsync(async (granule) => {
+        const { granuleUr, granuleConceptId } = granule
+
+        const splitTitle = granuleUr.split('_')
```
eudoroolivares2016, line 103: just some comment here; this is just an example if there is a ticket on them to do that we should just link it
```suggestion
// TODO: Call Swoldr API when it supports passing concept-id and deriving these values
        const splitTitle = granuleUr.split('_')
```

serverless/src/submitSwodlrOrder/__tests__/retrieveCMRGranules.test.js
```diff
@@ -0,0 +1,134 @@
...
+            title: 'SWOT_L2_HR_Raster_250m_UTM34L_N_x_x_x_013_514_086F_20240414T225526_20240414T225547_PIC0_01'
+          }
+          ]
+        }
+      })
+
+    const collectionConceptId = 'C2799438271-POCLOUD'
+
+    const granuleParams = {
+      collection_concept_id: 'C2799438271-POCLOUD'
```
eudoroolivares2016, line 34: nitpick;
```suggestion
      collection_concept_id: collectionConceptId
```

serverless/src/submitSwodlrOrder/handler.js
```diff
@@ -0,0 +1,202 @@
...
+      const {
+        access_method: accessMethod,
+        collection_id: collectionConceptId,
+        environment: earthdataEnvironment,
+        granule_params: granuleParams
+      } = retrievalRecord
+
+      const { type, url: swodlrUrl } = accessMethod
+
+      // No need for Accessmethod at this stage of dev.
```
eudoroolivares2016, line 86: nitpick;
```suggestion
      // TODO: Add accessMethod after design implementation
```
is this accurate? I think we should put this under TODO and explain when its expected will be added if it is needed.

serverless/src/submitSwodlrOrder/handler.js
```diff
@@ -0,0 +1,202 @@
...
+          utmZoneAdjust,
+          mgrsBandAdjust
+        }
+
+        const requestId = uuidv4()
+
+        console.log(`Submitting retrieval_order ${id} to swodlr with requestId ${requestId}`)
+
+        // Submit the order
+
```
eudoroolivares2016, line 141: nitpick;
```suggestion
```

serverless/src/submitSwodlrOrder/handler.js
```diff
@@ -0,0 +1,202 @@
...
+          },
+          headers: {
+            authorization: `Bearer ${accessToken}`,
+            'X-Request-Id': requestId
+          }
+        })
+
+        const { data: responseData, errors } = response
+
+        console.log(errors)
```
eudoroolivares2016, line 157: Maybe only print this if there are errors in the catch block

serverless/src/submitSwodlrOrder/handler.js
eudoroolivares2016: We should remove the comment then

eudoroolivares2016:
From the branch I was able to submit Jobs to the `Swodlr` API

eudoroolivares2016 (approved):

serverless/src/submitSwodlrOrder/__tests__/handler.test.js
```diff
@@ -0,0 +1,550 @@
...
+    const dbCon = knex({
+      client: 'pg',
+      debug: false
+    })
+
+    // Mock the db connection
+    mockKnex.mock(dbCon)
+
+    return dbCon
+  })
```
Bccorb, line 31: Do we need to mock this every time we start a test? Wouldn't `mockImplementaton` serve better here as this implementation is not changing. Then it can be outside of the before each and save resources when running?

serverless/src/submitSwodlrOrder/__tests__/retrieveCMRGranules.test.js
```diff
@@ -0,0 +1,134 @@
...
+
+import * as getEarthdataConfig from '../../../../sharedUtils/config'
+
+import { retrieveCMRGranules } from '../retrieveCMRGranules'
+
+describe('retrieveCMRGranules', () => {
+  beforeEach(() => {
+    jest.spyOn(getEarthdataConfig, 'getEarthdataConfig').mockImplementation(() => ({
+      cmrHost: 'https://cmr.earthdata.nasa.gov'
+    }))
```
Bccorb, line 11: Does this need to be done beforeEach? Why not beforeAll?

serverless/src/submitSwodlrOrder/handler.js
```diff
@@ -0,0 +1,198 @@
...
+          headers: {
+            authorization: `Bearer ${accessToken}`,
+            'X-Request-Id': requestId
+          }
+        })
+
+        const { data: responseData, errors } = response
+
+        if (errors) {
+          console.log(errors)
```
Bccorb, line 155: Such a small thing, but consider changing this to `console.error`

serverless/src/submitSwodlrOrder/retrieveCMRGranules.js
```diff
@@ -0,0 +1,47 @@
+import axios from 'axios'
+
+import { stringify } from 'qs'
```
Bccorb, line 3: What is this stringify? Is built in javascript version not the same? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

Bccorb (changes requested):
Considerations:
- Using console.error, instead of log when throwing/acknowledging errors
- Use the Javascript native over library stringify for longer term support and consistency
- Updating unit tests to create spies only as many times as needed. Before all vs Before Each vs Global Spies.

serverless/src/submitSwodlrOrder/retrieveCMRGranules.js
bnp26: It formats the object in a specific way (with brackets for example)

serverless/src/submitSwodlrOrder/__tests__/handler.test.js
eudoroolivares2016: That is a good point, I think the idea was to mock it the way that the handler runs the test so each time it is going to be mocked per lambda call (when a job is picked up in the queue that is) mocking it once I think is to ensure that it is only making one database connection per `test` if it made more it would be uncaught and fail the test. To be fair this is the test apparatus on the other handlers such as `serverless/src/submitCatalogRestOrder/__tests__/handler.test.js` or `serverless/src/submitHarmonyOrder/__tests__/handler.test.js` which at least I would rather be consistent.  A good work item for innovation might be to see if we can reduce the amount of time that the unit tests take to run more generally

bnp26: I agree with Ed. We should keep this as it is (for the time being)

serverless/src/submitSwodlrOrder/retrieveCMRGranules.js
Bccorb: Is this different from JSON Stringify?

Bccorb (approved):

PR #1745, opened 2024-05-07
EDSC-4031: Always route to saved Projects when selected.

Bccorb:
# Overview

### What is the feature?

Created a SavedProjects page and route to avoid the accidental appending of route params to the project page.

### What is the Solution?

- New Component
- New Route
- Updated the link for saved projects to the new route

- Saved Projects
- Projects when missing all params

# Testing

### Reproduction steps

This is confirmable by:

1. Logging in to EDSC
2. Selecting the save project icon in the right hand tool bar
3. Entering a project name
4. Adding a collection to the project via the plus button on a collection
5. Clicking the user icon in the right hand tool bar and selecting "saved projects"
6. You should observe that you are routed to  '/savedProjects' and it is in face your saved projects.

### Attachments

# Checklist

### DIFF
--- a/static/src/js/routes/Project/Project.js
+++ b/static/src/js/routes/Project/Project.js
@@ -1,7 +1,7 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
-import { withRouter } from 'react-router-dom'
+import { withRouter, Redirect } from 'react-router-dom'
 import { Helmet } from 'react-helmet'
 
 import { getEnvironmentConfig } from '../../../../../sharedUtils/config'
@@ -17,7 +17,6 @@ import ProjectPanelsContainer from '../../containers/ProjectPanelsContainer/Proj
 import OverrideTemporalModalContainer
   from '../../containers/OverrideTemporalModalContainer/OverrideTemporalModalContainer'
 import EdscMapContainer from '../../containers/MapContainer/MapContainer'
-import SavedProjectsContainer from '../../containers/SavedProjectsContainer/SavedProjectsContainer'
 import AuthRequiredContainer from '../../containers/AuthRequiredContainer/AuthRequiredContainer'
 
 const mapDispatchToProps = (dispatch) => ({
@@ -66,24 +65,10 @@ export class Project extends Component {
     const { search } = location
     const { edscHost } = this
 
-    // If there are no params in the URL, show the saved projects page
+    // If there are no project params in the URL, re-route to the saved Projects page
     if (search === '') {
       return (
-        <AuthRequiredContainer>
-          <Helmet>
-            <title>Saved Projects</title>
-            <meta name="title" content="Saved Projects" />
-            <meta name="robots" content="noindex, nofollow" />
-            <link rel="canonical" href={`${edscHost}/projects`} />
-          </Helmet>
-          <div className="route-wrapper route-wrapper--dark route-wrapper--content-page">
-            <div className="route-wrapper__content">
-              <div className="route-wrapper__content-inner">
-                <SavedProjectsContainer />
-              </div>
-            </div>
-          </div>
-        </AuthRequiredContainer>
+        <Redirect to="/saved-projects" />
       )
     }
 
--- a/static/src/js/routes/SavedProjects/SavedProjects.js
+++ b/static/src/js/routes/SavedProjects/SavedProjects.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react'
+import React from 'react'
 import { withRouter } from 'react-router-dom'
 import { Helmet } from 'react-helmet'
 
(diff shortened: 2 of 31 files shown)


static/src/js/routes/SavedProjects/SavedProjects.js
```diff
@@ -0,0 +1,40 @@
+import React, { Component } from 'react'
+import { withRouter } from 'react-router-dom'
+import { Helmet } from 'react-helmet'
+
+import { getEnvironmentConfig } from '../../../../../sharedUtils/config'
+import SavedProjectsContainer from '../../containers/SavedProjectsContainer/SavedProjectsContainer'
+import AuthRequiredContainer from '../../containers/AuthRequiredContainer/AuthRequiredContainer'
+
+export class SavedProjects extends Component {
```
eudoroolivares2016, line 9: We should make this a functional component like in `static/src/js/routes/Preferences/Preferences.js`

static/src/js/routes/SavedProjects/__tests__/SavedProjects.test.js
```diff
@@ -0,0 +1,60 @@
...
+import SavedProjectsContainer
+  from '../../../containers/SavedProjectsContainer/SavedProjectsContainer'
+
+// Mock react-leaflet because it causes errors
+jest.mock('react-leaflet', () => ({
+  createLayerComponent: jest.fn().mockImplementation(() => {}),
+  createControlComponent: jest.fn().mockImplementation(() => {})
+}))
+
+Enzyme.configure({ adapter: new Adapter() })
```
eudoroolivares2016, line 18: If we can avoid it we should use RTL on new tests so we don't accumulate new tech debt

Bccorb: Great point! I converted all of these! Thank you for keeping me honest!

Bccorb:
> `/savedProjects` is not a valid URL, it should be `/saved-projects` to match every other URL in `App.js`.
> 
> Instead of fixing the bug in the ticket you added a new feature and left the original buggy code abandoned in the codebase. The `Project.js` route already has code with comments that explain how the behavior should work. This is occurring because `?projectId=...` is still in the URL when it switches to the `/projects` endpoint. Either track down the bug and fix it, or remove the buggy code.
> 
> See CONTRIBUTING.md, _all_ commits have to start with the ticket number.

All feedback incorporated. 
- Removed saved project logic from the project url
- Added logic to correct an issue where a project url with no params could still be navigated to
- Added the back to search to the saved-projects url
- renamed the url endpoint

static/src/js/routes/Project/Project.js
```diff
@@ -66,24 +65,10 @@ export class Project extends Component {
     const { search } = location
     const { edscHost } = this
 
-    // If there are no params in the URL, show the saved projects page
+    // If there are no project params in the URL, re-route to the saved Projects page
     if (search === '') {
```
eudoroolivares2016, line 69: I'm a bit confused I'll pull this down and test it myself but, if the conditional here `if (search === '')` is failing to properly work i.e. its `false` when it should be true, which is why we are getting routed to the other component wouldn't this still have the same issue even if  it is on a difference page?

static/src/js/routes/SavedProjects/SavedProjects.js
```diff
@@ -0,0 +1,31 @@
...
+export const SavedProjects = () => {
+  const { edscHost } = getEnvironmentConfig()
+
+  return (
+    <AuthRequiredContainer>
+      <Helmet>
+        <title>Saved Projects</title>
+        <meta name="title" content="Saved Projects" />
+        <meta name="robots" content="noindex, nofollow" />
+        <link rel="canonical" href={`${edscHost}/projects`} />
```
eudoroolivares2016, line 18:
```suggestion
        <link rel="canonical" href={`${edscHost}/saved-projects`} />
```

static/src/js/routes/Project/Project.js
eudoroolivares2016: So the issue I seem to be seeing is that something like `http://localhost:8080/saved-projects?projectId=3519105921`is always going to strip out the projectId but, also that `http://localhost:8080/projects?projectId=3519105921` is redirecting me (fill in any projectId) back into the `saved-projects` page so I can't go back to my previous projects

eudoroolivares2016:
Update I was looking at this some because of a related ticket I have. It seems like the issue is that we can't really differentiate between an empty project and the saved project page. Its possible to fix the issue where it will go to saved project. The way that we are currently checking makes sense i.e. if there are search parameters then we go to the project page the issue is that somehow the saved projects page is carrying some search parameters when clicking on the saved projects page

eudoroolivares2016:
Closing PR likely will use a different solution otherwise we'll reference this one but, its gotten quite behind the `HEAD`

PR #1754, opened 2024-06-25
EDSC-4114: Set the concatenate flag to harmony to false if it is unse…

eudoroolivares2016:
# Overview

### What is the feature?

Update the submitHarmonyOrder so that if concatenation is not selected pass the flag as `false` instead of not at all

### What is the Solution?

Update the logic so that we pass what the state of the checkbox is. It is being initialized to false already

Harmony orders where concatenation is enabled

# Testing

### Reproduction steps

- **Environment for testing: PROD**
- **Collection to test with: C2930725014-LARC_CLOUD**

Ensure that with this collection and using the service we can submit a Harmony service that has concatenation and that the Harmony order (can be submitted locally with `npm run invoke-local -- --function submitHarmonyOrder --path ./event.json` where `event.json` is a mock of the order going to SQS.

### Attachments

I have attached my test documentation to the Jira ticket where I tried a number of different Harmony submissions from local machine with method above

# Checklist

### DIFF
--- a/serverless/src/submitHarmonyOrder/__tests__/constructOrderPayload.test.js
+++ b/serverless/src/submitHarmonyOrder/__tests__/constructOrderPayload.test.js
@@ -694,8 +694,8 @@ describe('constructOrderPayload', () => {
     })
   })
 
-  describe('when supportsConcatenation = true and enableConcatenateDownload = true', () => {
-    test('constructs a payload containing concatenate = true', async () => {
+  describe('when the collection supports concatenation and it has been selected', () => {
+    test('constructed payload sets concatenation to true', async () => {
       nock(/cmr/)
         .matchHeader('Authorization', 'Bearer access-token')
         .get('/search/granules.json?point%5B%5D=-77%2C%2034')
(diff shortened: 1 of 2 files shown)


rushgeo (approved):

serverless/src/submitHarmonyOrder/__tests__/constructOrderPayload.test.js
```diff
@@ -732,8 +732,8 @@ describe('constructOrderPayload', () => {
     })
   })
 
-  describe('when supportsConcatenation = false or enableConcatenateDownload = false', () => {
-    test('constructed payload does not contain concatenate', async () => {
+  describe('when the accessMethod supports concatenation but, it is not selected', () => {
```
dpesall, line 735:
```suggestion
  describe('when the accessMethod supports concatenation, but it is not selected', () => {
```

Minor grammatical correction

dpesall (approved):

bnp26 (approved):

PR #1759, opened 2024-07-02
EDSC-4141: Adds panels playwright spec

daniel-zamora:
# Overview

### What is the feature?

Adds playwright spec for the cypress e2e panels test with an aim of removing cypress completely

Playwright test suite

# Testing

### Reproduction steps

N/A

# Checklist


eudoroolivares2016 (approved):
Tested this locally looked good on playwright:ui added comments as preferences but, fine either way. There were some comments on the `cypress/e2e/panels/panels_behavior.cy.js` which were somewhat useful that we've left out on the new one my preference would be to just port those as well but, not necessary

PR #1768, opened 2024-07-17
EDSC-3538 sets collection sort to env var or default -score

daniel-zamora:
# Overview

### What is the feature?

Please summarize the feature or fix.

### What is the Solution?

Summarize what you changed.

List impacted areas.

# Testing

### Reproduction steps

1. Step 1
2. Step 2...

### Attachments

# Checklist


static/src/js/reducers/query.js
```diff
@@ -18,7 +21,7 @@ const initialState = {
     pageNum: 1,
     spatial: {},
     temporal: {},
-    sortKey: ['-usage_score']
+    sortKey: [collectionSearchResultsSortKey || '-score']
```
eudoroolivares2016, line 24: So '-score' === `` as far as CMR is concerned because its the default?

static/src/js/components/SearchPanels/SearchPanels.js
```diff
@@ -177,10 +177,10 @@ class SearchPanels extends PureComponent {
 
     const {
       pageNum: collectionsPageNum = 1,
-      sortKey: collectionsSortKey = ''
+      sortKey: collectionsSortKey = '-score'
```
macrouch, line 180: All of the places where we have a string value for a sort key would benefit from creating a constant. So instead of typing/copying `-score` everywhere, you could do something like `CollectionSortKeys.ScoreDescending`. This makes writing the code a little easier since you can get autocomplete from the editor to help you with the types of values to expect

static/src/js/reducers/query.js
macrouch, line 24: This doesn't need a default value added, it can just be the `collectionSearchResultsSortKey`

schemas/sitePreferencesSchema.json
```diff
@@ -17,7 +17,6 @@
 			"default": "default",
 			"enum": [
 				"default",
-				"relevance",
```
daniel-zamora, line 20: revert this change or add -score

jest.config.cjs
```diff
@@ -2,7 +2,7 @@
 const esModulesToIgnore = ['nanoid', 'lodash-es', 'sinon'].join('|')
 
 module.exports = {
-  collectCoverage: true,
+  collectCoverage: false,
```
macrouch, line 5: Set this back to true

macrouch (approved):

dpesall (approved):

PR #1769, opened 2024-07-17
EDSC-4195: Add/expand Metric event for Spatial filtering

eudoroolivares2016:
# Overview

### What is the feature?

This is to add a metric for spatial selection so that we can track when users select a shape using the `spatialSelectionsDropDownMenu` in the feature analysis it was requested to be able to take those values and compare them against when users are using the leaflet edit-control for the map which we already track as `map` event. This also reduces some technical debt by removing some console logs which were not necessary and flooded the dev console in `serverless-offline`, updates class components into functional components, and updates some tests written in Enzyme into RT. This also works a bug in the `SpatialSelection` component where an error was getting thrown whenever a user would edit and save an existing shape on the map. This also adds Circle to the possible prints for the spatialEdit metric which is meant to capture that behavior

### What is the Solution?

Create new metric, update existing components, and tests,

For the spatialSelection bug the issue was that the state in the onStop was not consistent across the two callback functions so I added a reference which is always up to date with the current state 

List impacted areas.

# Testing

### Reproduction steps

Two types of tests will be needed here.

1) We need to do a regression test of the spatialSelectionDropdown to ensure that the update maintains the same functionality. We also need to ensure that the `error` message from the edit shape is no longer getting reported.

2) We need to test that the metric is going to the google data layer correctly. One way to do this is to use the preview feature of google looker studio. Create a new workspace clone but, also include the new tag that you are using ensure that given the action (selecting the shape from the dropdown) an event is registered on google tag manager. If we can test e2e that would be better i.e. add the tag into GA as a custom param then pull in that custom param into a new looker-studio document to visualize events. We will need a GA account tied to the lower-env for that to work though.

### Attachments

# Checklist


static/src/js/middleware/metrics/events.js
```diff
@@ -11,6 +11,7 @@ const { dataLayer = [] } = window
 
 /**
 * Pushes a virtualPageView event on the dataLayer. Only fires on PUSH events.
+$ These `PUSH` events only fire during page transitions
```
dpesall, line 14: Is these supposed to be a `$` instead of `*` in the comment?

dpesall (approved):

static/src/js/middleware/metrics/events.js
eudoroolivares2016: No, good catch thanks

static/src/js/components/SpatialSelection/SpatialSelection.js
```diff
@@ -468,21 +471,23 @@ const SpatialSelection = (props) => {
 
   // Callback from EditControl, called when clicking the edit button
   const onEditStart = () => {
-    setPreEditBounds(layers.current.map((layer) => boundsToPoints(layer)))
+    const editBounds = layers.current.map((layer) => boundsToPoints(layer))
```
eudoroolivares2016, line 474: Put back since we aren't debugging anymore

static/src/js/containers/SpatialSelectionDropdownContainer/__tests__/SpatialSelectionDropdownContainer.test.js
```diff
@@ -34,13 +33,21 @@ describe('mapDispatchToProps', () => {
...
 })
 
 describe('SpatialSelectionDropdownContainer component', () => {
   test('passes its props and renders a single SpatialSelectionDropdown component', () => {
-    const { enzymeWrapper } = setup()
-
-    expect(enzymeWrapper.find(SpatialSelectionDropdown).length).toBe(1)
-    expect(typeof enzymeWrapper.find(SpatialSelectionDropdown).props().onToggleShapefileUploadModal).toEqual('function')
+    setup()
+    expect(screen.getByRole('button', { name: 'spatial-selection-dropdown' })).toBeInTheDocument()
```
eudoroolivares2016, line 51: add newline between

static/src/js/middleware/metrics/__tests__/helpers.test.js
```diff
@@ -34,6 +34,7 @@ describe('helpers', () => {
     })
   })
 
+  // TODO these data structures don't seem right they're arrays in the console
```
eudoroolivares2016, line 37: For the other tests these were arrays should just update them to better match what we are seeing during run-time

static/src/js/components/SpatialSelection/SpatialSelection.js
```diff
@@ -99,6 +99,9 @@ const SpatialSelection = (props) => {
   const drawControl = useRef(null)
   const [preEditBounds, setPreEditBounds] = useState(null)
   const featureGroupRef = useRef(null)
+  // https://stackoverflow.com/questions/57847594/accessing-up-to-date-state-from-within-a-callback/62453660#62453660
+  const stateRefPreEditBounds = useRef()
```
eudoroolivares2016, line 103: We should try to fix this with `useCallback` useRef outside the context of manipulation DOM elements is not preferred

eudoroolivares2016: We discussed this further didn't seem to be an alternative way without potentially significantly more refactoring here to do this. useRef is already being used and currently this block isn't actually running due to hte bug

static/src/js/middleware/metrics/__tests__/index.test.js
```diff
@@ -132,6 +133,22 @@ describe('metrics middleware', () => {
     expect(events.spatialEdit).toHaveBeenCalledWith(action)
   })
 
+  test('calls spatialSelection event', () => {
+    // eslint-disable-next-line no-import-assign
```
macrouch, line 137: You can remove all of the `eslint-disable`s in this file with this mock at the top of the file
```
jest.mock('../events', () => ({
  collectionSortChange: jest.fn(),
  dataAccess: jest.fn(),
  defaultClick: jest.fn(),
  map: jest.fn(),
  relatedCollection: jest.fn(),
  spatialEdit: jest.fn(),
  spatialSelection: jest.fn(),
  timeline: jest.fn(),
  timing: jest.fn(),
  virtualPageview: jest.fn()
}))
```

eudoroolivares2016: Great call out thanks

macrouch (approved):

static/src/js/components/SpatialDisplay/SpatialSelectionDropdown.jsx
```diff
@@ -21,16 +21,15 @@ import EDSCIcon from '../EDSCIcon/EDSCIcon'
 
 import './SpatialSelectionDropdown.scss'
 
-export class SpatialSelectionDropdown extends PureComponent {
-  constructor(props) {
-    super(props)
-    this.onItemClick = this.onItemClick.bind(this)
-  }
+const SpatialSelectionDropdown = (props) => {
```
trevorlang, line 24: I wonder if this should be memoized with `React.memo`, since it was a pure component previously

trevorlang (approved):
Approved but left a comment on something that caught my eye

PR #1777, opened 2024-07-31
EDSC-4177: Implement performance metric: time to query and display collection results

rushgeo:
# Overview

### What is the feature?
Adding performance measurement for the collections panel. The collections panel is populated by results from CMR, and is roughly 1/3 of the app as seen by the users. To most users, the app will not appear ready to use until these collection results are displayed (even though they can still do text searches).

### What is the Solution?

Measure the time it takes to request the collections and render them into the app. 
Including the rendering time makes this distinct from the existing `FINISHED_COLLECTIONS_TIMER` event, which only includes the request time,and is displayed that in the footer of the app.

The collections rendering time is printed to the dev console. In the future, this performance metric will be sent to Google Analytics to measure real user performance. 
It is also tested by a GitHub action triggering the Playwright performance tests. In the future, these test results will be sent to Cloudwatch. The Playwright tests mock the CMR and GIBS requests, which could otherwise introduce variable response times into the performance metrics. By mocking those requests, this test should only measure the impact of our code changes on the rendering time.

State, Logging, e2e tests

# Testing

### Reproduction steps

- **Environment for testing:** Any
- **Collection to test with:** None

A developer server of EDSC (but not a deployed instance) will print to your browser's developer console a log message resembling:
```
[performance] Collections load time: 1842
```

See also that the `performance` GitHub action plan executed and the performance spec passed. The log of the performance test should include a lines for three browsers resembling:
```
[performance] Collections load time, chromium: 163 ms
[performance] Collections load time, firefox: 270 ms
[performance] Collections load time, webkit: 244 ms
```

# Checklist


tests/e2e/performance/performance.spec.js
```diff
@@ -1,18 +1,25 @@
+/* eslint-disable capitalized-comments */
 import { test, expect } from 'playwright-test-coverage'
+import twentyCollections from './__mocks__/twenty_collections.json'
 
 test.describe('Performance Benchmarking', () => {
   test.beforeEach(async ({ page }) => {
-    await page.route('**/*.{png,jpg,jpeg}', (route) => route.abort())
```
eudoroolivares2016, line 5: This is just to mock the map so we're making gibs call here aren't those going to mess with the metric values?

rushgeo:
Switched back to draft while I work on replacing console logs from the frontend and mocking requests to GIBS.

tests/e2e/performance/performance.spec.js
rushgeo: I reverted this change so that all GIBS requests get aborted again.

static/src/js/reducers/__tests__/collectionsResults.test.js
```diff
@@ -78,6 +80,32 @@ describe('FINISHED_COLLECTIONS_TIMER', () => {
   })
 })
 
+describe('FINISHED_COLLECTIONS_RENDERING_TIMER', () => {
+  test('returns the correct state', () => {
+    const action = {
+      type: FINISHED_COLLECTIONS_RENDERING_TIMER
+    }
+
+    // Set current time to 10, and future time to 15
```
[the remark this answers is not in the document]
rushgeo: I updated the wording in the `FINISHED_COLLECTIONS_TIMER` test in this script and in the new test that I've moved to the middleware at `static/src/js/middleware/metrics/__tests__/events.test.js:finishedCollectionsRendering`

static/src/js/middleware/metrics/events.js
```diff
@@ -7,6 +7,8 @@ import {
   computeFacets
 } from './helpers'
 
+import staticConfig from '../../../../../static.config.json'
```
daniel-zamora, line 10: use sharedUtils/config.js

rushgeo: This is a workaround because `config.js` uses lodash's merge to combine the static config and the override, and for some reason, just can't get lodash to play nicely within Playwright tests.

static/src/js/reducers/__tests__/collectionsResults.test.js
```diff
@@ -58,15 +58,14 @@ describe('FINISHED_COLLECTIONS_TIMER', () => {
       type: FINISHED_COLLECTIONS_TIMER
     }
 
-    // Set current time to 10, and future time to 15
-    // Load time will equal 5
+    // Set start time to 10, current time to 15 and expect a difference of 5
```
eudoroolivares2016, line 61:
```suggestion
    // Set start time to 10, current time to 15 and expect a difference of 5
    // Load time will equal 5
```
seems like that comment from before was still accurate but, also is kind of unnecessary I'm good either way

static/src/js/middleware/metrics/events.js
rushgeo: Nevermind - I thought I needed this workaround, but that was for a different earlier design where Playwright would have needed to get the URLs for our lambdas. In the current design, only the app code needs to load the `config.js`, which works fine. The latest commit switches this to use `config.js`

eudoroolivares2016 (approved):

static/src/js/reducers/__tests__/collectionsResults.test.js
rushgeo: Yeah this is in response to Drew's earlier feedback. Originally I copied this comment line from this old test for my new test. I reworded it based on his feedback, then realized that I had gotten it from the older test, so then I edited the old one to match. Either way works for me.

static/src/js/middleware/metrics/events.js
```diff
@@ -305,3 +306,18 @@ export const relatedCollection = (action) => {
     relatedCollectionLabel: collectionId
   })
 }
+
+/**
```
rushgeo, line 310: This makes the rendering time available to Playwright via the console log when EDSC is running in development (i.e. locally and in GitHub Actions).
Later work will handle the other side of the conditional: when EDSC is deployed, send the metric to Google Analytics so it can report on the diversity of real user experiences.

dpesall (approved):

rushgeo:
This has a test failing that doesn't make much sense - state is persisting between Jest tests. We're also not very excited about the actual method we've used to instrument the performance metrics. So I'll close this for now and we can revisit it later potentially with a different implementation.

This PR includes some changes to the two existing performance tests that may be worth spinning out into their own smaller PR.

PR #1788, opened 2024-08-19
EDSC-4064: Updated Error banner to match mockups

bnp26:
# Overview

### What is the feature?

Updated the banner to match what was provided in the mockups.

### What is the Solution?

Summarize what you changed.

List impacted areas.

# Testing

### Reproduction steps

1. Step 1
2. Step 2...

### Attachments

# Checklist

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,9 +7,16 @@
   },
   "rules": {
     // Allowing cyclic dependancies
-    "import/no-cycle": "off"
+    "import/no-cycle": "off",
+    "import/no-unresolved": [
+      "error",
+      {
+        "ignore": [
+          "^@edsc/earthdata-react-icons/.+"
+        ]
+      }
+    ]
   },
-
   "overrides": [
     {
       "files": [
@@ -28,12 +35,10 @@
       }
     }
   ],
-
   // Use AirBnb settings as a base
   "extends": [
     "@edsc"
   ],
-
   // Define version settings
   "settings": {
     "react": {
--- a/static/src/js/components/Banner/Banner.jsx
+++ b/static/src/js/components/Banner/Banner.jsx
@@ -25,7 +25,7 @@ export const Banner = ({
     <div className={bannerClassNames}>
       <div className="banner__content">
         <h2 className="banner__title">
-          <span className="banner__icon icon"><AlertHighPriority size={20} /></span>
+          <AlertHighPriority className="banner__icon icon" size="1.375rem" />
           {title}
         </h2>
         {
--- a/static/src/js/components/Banner/Banner.jsx
+++ b/static/src/js/components/Banner/Banner.jsx
@@ -1,7 +1,8 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import classNames from 'classnames'
-import { FaTimesCircle } from 'react-icons/fa'
+import { Close } from '@edsc/earthdata-react-icons/horizon-design-system/hds/ui'
+import { AlertHighPriority } from '@edsc/earthdata-react-icons/horizon-design-system/earthdata/ui'
 
 import Button from '../Button/Button'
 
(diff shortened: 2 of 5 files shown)


static/src/js/components/Banner/Banner.jsx
```diff
@@ -1,7 +1,8 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import classNames from 'classnames'
-import { FaTimesCircle } from 'react-icons/fa'
+import { Close } from '@edsc/earthdata-react-icons/horizon-design-system/hds/ui'
+import { AlertHighPriority } from '@edsc/earthdata-react-icons/horizon-design-system/earthdata/ui'
```
eudoroolivares2016, line 5: image
This one?

static/src/js/components/Banner/__tests__/Banner.test.js
```diff
@@ -27,7 +27,7 @@ describe('Bannner component', () => {
   test('should render self', () => {
     const { enzymeWrapper } = setup()
 
-    expect(enzymeWrapper.find('h2.banner__title').text()).toEqual('title')
+    expect(enzymeWrapper.find('h2.banner__title').text()).toEqual('<AlertHighPriority />title')
```
eudoroolivares2016, line 30: I am a bit confused why the banner has HTML in it?

.eslintrc
```diff
@@ -7,9 +7,16 @@
   },
   "rules": {
     // Allowing cyclic dependancies
-    "import/no-cycle": "off"
+    "import/no-cycle": "off",
```
eudoroolivares2016, line 10: Just as a comment for others here. I also run into an issue here for `eslint`. I am not sure root-cause but, it seems like the structure of the npm package. We should try to fix this if we can but, I don't think we should spend the time to do that right now

dpesall (approved):

static/src/js/components/Banner/__tests__/Banner.test.js
```diff
@@ -15,43 +14,47 @@ function setup(overrideProps = {}) {
...
+  test('clicking the close button calls onClose', async () => {
+    const user = userEvent.setup()
+    const { props } = setup()
 
-    const button = enzymeWrapper.find(Button)
-    button.simulate('click')
+    const closeBtn = screen.getByRole('button', { name: 'close' })
+    await user.click(closeBtn)
 
     expect(props.onClose).toBeCalledTimes(1)
```
eudoroolivares2016, line 44:
```suggestion
    expect(props.onClose).toBeCalledTimes(1)
    expect(props.onClose).toBeCalledWith()
```
I think we should still check that its called with no arguments

static/src/js/components/Banner/__tests__/Banner.test.js
```diff
@@ -15,43 +14,47 @@ function setup(overrideProps = {}) {
...
     props
   }
 }
 
 describe('Bannner component', () => {
   test('should render self', () => {
-    const { enzymeWrapper } = setup()
+    setup()
+
+    expect(screen.getByRole('heading', { name: 'Banner Title' })).toBeInTheDocument()
```
eudoroolivares2016, line 28: nit; you don't need to check if its in the document if we check

static/src/js/components/Banner/__tests__/Banner.test.js
bnp26: shouldn't have that in there anymore

static/src/js/components/Banner/__tests__/Banner.test.js
bnp26: It's not called with no arguments, that was enzyme. Because this is RTL, it has an event passed into the onClick.

eudoroolivares2016: We should check that it was called with an event that we expect in that case

PR #1792, opened 2024-08-23
EDSC-4072: Update typography to match look and feel toolkit

rushgeo:
# Overview

### What is the feature?
Update typography.

### What is the Solution?

Overrode the default font families from Bootstrap with [Public Sans](https://fonts.google.com/specimen/Public+Sans) for body text, [Inter](https://fonts.google.com/specimen/Inter) for headings, and [DM Mono](https://fonts.google.com/specimen/DM+Mono) for monospaced text.

Updated the logo text to better match the example from the Figma render we received.

Followup on [EDSC-4073](https://github.com/nasa/earthdata-search/pull/1786): included a fix for disabling underlines for DOI badge links.

Displayed text.

# Testing

### Reproduction steps

- **Environment for testing:** Any
- **Collection to test with:** Any

The body and heading text is very similar to the existing fonts from Bootstrap, at least on our Mac developer machines. The easiest way to verify the fonts have been set as intended is to use the web inspector to check the applied styles for various elements. You can also use an identical URL for the dev and PROD instances of EDSC and toggle back and forth between them.

Currently, the only monospace font usage is in the Download Script panel. Headings and body text can be checked on any page. The app logo text should be in Inter, the primary heading font.

#### App Logo Text (Inter)
_Before_
image

_After_
image

In addition to the `font-family`, this PR changes the `font-size` (smaller), `letter-spacing` (none), and transparency on the letters "EARTH" (none).

#### Headings (Inter)
_Before_
image

_After_
image

#### Body Text (Public Sans)
_Before_
image

_After_
image

#### Download Script Panel (DM Mono)
_Before_
image

_After_
image

# Checklist

### DIFF
--- a/static/src/js/components/AppLogo/AppLogo.jsx
+++ b/static/src/js/components/AppLogo/AppLogo.jsx
@@ -33,8 +33,8 @@ const AppLogo = () => (
         }
         updatePath
       >
-        <span className="app-logo__site-name-ent app-logo__site-name-ent--e">Earth</span>
-        <span className="app-logo__site-name-ent app-logo__site-name-ent--s">data Search</span>
+        <span className="app-logo__site-name">Earth</span>
+        <span className="app-logo__site-name-emphasis">data Search</span>
       </PortalLinkContainer>
     </span>
     <Switch>
--- a/static/src/js/components/AppLogo/AppLogo.jsx
+++ b/static/src/js/components/AppLogo/AppLogo.jsx
@@ -33,8 +33,8 @@ const AppLogo = () => (
         }
         updatePath
       >
-        <span className="app-logo__site-name-ent app-logo__site-name-ent--e">Earthdata</span>
-        <span className="app-logo__site-name-ent app-logo__site-name-ent--s">Search</span>
+        <span className="app-logo__site-name-ent app-logo__site-name-ent--e">Earth</span>
+        <span className="app-logo__site-name-ent app-logo__site-name-ent--s">data Search</span>
       </PortalLinkContainer>
     </span>
     <Switch>
--- a/static/src/js/components/AppLogo/AppLogo.scss
+++ b/static/src/js/components/AppLogo/AppLogo.scss
@@ -81,7 +81,7 @@
   }
 
   &__site-name-emphasis {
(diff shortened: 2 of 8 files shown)


static/src/js/components/AppLogo/AppLogo.scss
```diff
@@ -60,10 +60,10 @@
   &__site-name {
     display: inline-block;
     color: $color__white;
-    font-size: 1.125rem;
+    font-size: 1rem;
     text-transform: uppercase;
-    letter-spacing: 0.05rem;
     font-weight: 300;
+    font-family: "Inter";
```
trevorlang, line 66: This should probably have some fallback to the system font stack

rushgeo: This now uses the same overridden bootstrap variable so it falls back to those fonts if Inter isn't available.

eudoroolivares2016 (approved):
LGTM

static/src/js/components/AppLogo/AppLogo.jsx
```diff
@@ -33,8 +33,8 @@ const AppLogo = () => (
         }
         updatePath
       >
-        <span className="app-logo__site-name-ent app-logo__site-name-ent--e">Earthdata</span>
-        <span className="app-logo__site-name-ent app-logo__site-name-ent--s">Search</span>
+        <span className="app-logo__site-name-ent app-logo__site-name-ent--e">Earth</span>
+        <span className="app-logo__site-name-ent app-logo__site-name-ent--s">data Search</span>
```
trevorlang, line 37: Okay, this is a minor thing but I don't think the modifier classes on these items or the html makes a ton of sense now.

I think jsx inside this link should change to

```suggestion
        <span className="app-logo__site-name-emphasis">Earth</span>
        data Search
```

And the scss would look something like this:

```scss
&__site-name {
    // ...current styles
    // the lighter font weight
}

&__site-name-emphasis {
    // the heavier font weight
}
```

rushgeo: Good catch! Implemented in the latest commits

trevorlang (approved):

PR #1801, opened 2024-09-11
EDSC-4170: Add Admin Preferences Metrics to Admin page

bnp26:
# Overview

### What is the feature?

Add new Preferences Metrics to the admin page.

### What is the Solution?

Created new handler to pull the site_preferences data and calculate the percentage of different values users use as their preferences . New page to display that data.

List impacted areas.
Admin pages

# Testing

### Reproduction steps

- **Environment for testing:*N/A*
- **Collection to test with:*Any*

1. go to localhost:8080/admin
2. Observe the displayed data matches the users.site_preferences table in the postgresql

### Attachments

# Checklist


static/src/js/routes/Admin/Admin.jsx
```diff
@@ -63,12 +64,20 @@ export class Admin extends Component {
...
+                  >
+                    View Preferences Metrics
+                  </PortalLinkContainer>
                 </Route>
                 <Route exact path={`${path}/retrievals`} component={AdminRetrievalsContainer} />
                 <Route exact path={`${path}/retrievals/:id`} component={AdminRetrievalContainer} />
                 <Route exact path={`${path}/projects`} component={AdminProjectsContainer} />
                 <Route exact path={`${path}/projects/:id`} component={AdminProjectContainer} />
                 <Route exact path={`${path}/retrievalsMetrics`} component={AdminRetrievalsMetricsContainer} />
+                <Route exact path={`${path}/preferencesMetrics`} component={AdminPreferencesMetricsContainer} />
```
macrouch, line 80: It looks like you copied what `retrievalsMetrics` is doing, but both are wrong. URLs are supposed to by hyphenated, see the URLs defined in App.jsx, https://github.com/nasa/earthdata-search/blob/main/static/src/js/App.jsx#L137

static/src/js/components/AdminPreferencesMetrics/__tests__/AdminPreferencesMetricsList.test.js
```diff
@@ -0,0 +1,139 @@
+import React from 'react'
+import { render, screen } from '@testing-library/react'
+
+import '@testing-library/jest-dom'
```
macrouch, line 4: If you import this into the jest setup file you just have to import it the one time and not into every test file that needs it

static/src/js/components/AdminPreferencesMetrics/__tests__/AdminPreferencesMetrics.test.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+  const onFetchAdminMetricsPreferences = jest.fn()
+
+  const props = {
+    onFetchAdminMetricsPreferences,
+    metricsPreferences
+  }
+
+  // https://testing-library.com/docs/example-react-router/
+  render(<AdminPreferencesMetrics {...props} />, { wrapper: BrowserRouter })
```
macrouch, line 37: React Router recommends using MemoryRouter in tests, https://reactrouter.com/en/v6.3.0/api#memoryrouter

static/src/js/components/AdminPreferencesMetrics/AdminPreferencesMetricsList.jsx
```diff
@@ -0,0 +1,72 @@
+/* eslint-disable react/prop-types */
```
macrouch, line 1: Don't add eslint-disable lines for a full file. You should never add this specific ignore, just define the prop types. If you absolutely have to ignore eslint it should only be for the specific line, and never the full file.

static/src/js/components/AdminPreferencesMetrics/AdminPreferencesMetrics.jsx
```diff
@@ -0,0 +1,52 @@
...
+
+import { AdminPage } from '../AdminPage/AdminPage'
+import { AdminPreferencesMetricsList } from './AdminPreferencesMetricsList'
+
+export const AdminPreferencesMetrics = ({
+  onFetchAdminMetricsPreferences,
+  metricsPreferences
+}) => {
+  useEffect(async () => {
+    await onFetchAdminMetricsPreferences()
```
macrouch, line 13: Fetching should be done within the container component, not the presentation component

serverless/src/adminGetPreferencesMetrics/handler.js
```diff
@@ -0,0 +1,163 @@
...
+
+  return sortedEntries.slice(0, 6)
+}
+
+/**
+ * Retrieve all the preferences metrics for the authenticated user
+ * @param {Object} event Details about the HTTP request that it received
+ * @param {Object} context Methods and properties that provide information about the invocation, function, and execution environment
+ */
+const adminGetPreferencesMetrics = async (event, context) => {
```
macrouch, line 25: This feels very inefficient. You are pulling data for every user out of the database, then looping and sorting the entire list multiple times. For a user list as big as prod that isn't going to work well.

I think it is worth looking at using better SQL to offload all of this logic onto the database.

serverless-configs/aws-functions.yml
```diff
@@ -686,6 +686,21 @@
...
 
+
+  adminGetPreferencesMetrics:
+    handler: serverless/src/adminGetPreferencesMetrics/handler.default
+    timeout: ${env:LAMBDA_TIMEOUT, '30'}
+    events:
+      - http:
+          method: get
+          cors: ${file(./serverless-configs/${self:provider.name}-cors-configuration.yml)}
+          path: admin/preferencesMetrics
```
macrouch, line 697: I'm not sure why, but all the URLs for our API Gateway use underscores, not hyphens. But this `path` should match all the others

static/src/js/components/AdminPreferencesMetrics/__tests__/AdminPreferencesMetricsList.test.js
bnp26: This should probably be it's own ticket (1 point)

static/src/js/routes/Admin/Admin.jsx
bnp26: Should this be converted on a different ticket then? Feels like changing retrievalMetrics is out of scope for this and keeping things consistent is good. It would be a very quick fix and would take little to no resources.

serverless/src/adminGetPreferencesMetrics/handler.js
bnp26: Tested on 300,000 users. Loaded in 2 sec.

static/src/js/containers/AdminPreferencesMetricsContainer/AdminPreferencesMetricsContainer.jsx
```diff
@@ -0,0 +1,46 @@
...
+import { connect } from 'react-redux'
+import PropTypes from 'prop-types'
+import { withRouter } from 'react-router-dom'
+
+import actions from '../../actions'
+import AdminPreferencesMetrics from '../../components/AdminPreferencesMetrics/AdminPreferencesMetrics'
+
+export const mapStateToProps = (state) => ({
+  metricsPreferences: state.admin.metricsPreferences,
+  preferencesLoading: state.admin.metricsPreferences.isLoading,
```
macrouch, line 11: The loading/loaded props aren't being used. You could either remove them from `mapStateToProps` and in the redux store, or add loading indicators to the UI.

static/src/js/util/request/admin/__tests__/metricsPreferencesRequest.test.js
```diff
@@ -0,0 +1,50 @@
...
+    const token = '123'
+    const request = new MetricsPreferencesRequest(token)
+
+    expect(request.authenticated).toBeTruthy()
+    expect(request.authToken).toEqual(token)
+    expect(request.baseUrl).toEqual('http://localhost:3000')
+  })
+})
+
+describe('metricsPreferencesRequest#transformResponse', () => {
```
macrouch, line 20: This is not a function on the `MetricsPreferencesRequest` class, it can just be handled by the test in the parent class

static/src/js/util/request/admin/metricsPreferencesRequest.js
```diff
@@ -0,0 +1,15 @@
+import Request from '../request'
+import { getEnvironmentConfig } from '../../../../../../sharedUtils/config'
+
+export default class MetricsPreferencesRequest extends Request {
```
macrouch, line 4: Everywhere else in the code this is referred to has Preferences Metrics, but this class is Metrics Preferences. Keep the naming consistent and switch this to Preferences before Metrics

bnp26: @macrouch The front end is referred to as Preferences Metrics, but we have another metrics stuff called Metrics Retrievals so I was going off that. Would you like me to change that to match up as well? The reason it's nice to have metrics first is that the spelling/alphabetical order you can see all the metrics together as apposed to them being set apart from one another when they do similar functionality.

macrouch: I'm just looking for consistency throughout the code. The component that uses it is "AdminPreferencesMetrics", the actions are "preferencesMetrics", the lambda is "adminGetPreferencesMetrics", but this request class is "metricsPreferencesRequest". 

If I wanted to find all the code related to this subject (for debugging/tracing/refactoring/etc.), I would do a project wide search for "preferencesMetrics", but that wouldn't return these files (making the above tasks more difficult).

static/src/js/containers/AdminPreferencesMetricsContainer/__tests__/AdminPreferencesMetricsContainer.test.js
```diff
@@ -0,0 +1,81 @@
...
+      }
+
+      expect(mapStateToProps(store)).toEqual(expectedState)
+    })
+  })
+})
+
+describe('AdminPreferencesMetricsContainer component', () => {
+  test('render AdminPreferencesMetrics with the correct props', () => {
+    setup()
```
macrouch, line 69:
```suggestion
    const { props } = setup()

    expect(props.onFetchAdminPreferencesMetrics).toHaveBeenCalledTimes(1)
    expect(props.onFetchAdminPreferencesMetrics).toHaveBeenCalledWith()

```

This verifies that the useEffect was called and that `onFetchAdminPreferencesMetrics` was called as expected

static/src/js/containers/AdminPreferencesMetricsContainer/__tests__/AdminPreferencesMetricsContainer.test.js
```diff
@@ -0,0 +1,81 @@
...
+
+const setup = () => {
+  const props = {
+    onFetchAdminPreferencesMetrics: jest.fn(),
+    preferences: {}
+  }
+
+  // https://testing-library.com/docs/example-react-router/
+  render(<AdminPreferencesMetricsContainer {...props} />, { wrapper: BrowserRouter })
+}
```
macrouch, line 30:
```suggestion

  return {
    props
  }
}
```

static/src/js/actions/admin/preferencesMetrics.js
```diff
@@ -0,0 +1,60 @@
...
+
+  const requestObject = new PreferencesMetricsRequest(authToken, earthdataEnvironment)
+
+  const response = await requestObject.all()
+    .then((responseObject) => {
+      const { data } = responseObject
+      const {
+        results
+      } = data
+      dispatch(setAdminPreferencesMetricsLoaded())
```
trevorlang, line 47: We should probably have some whitespace here

static/src/js/components/AdminPreferencesMetrics/AdminPreferencesMetricsList.jsx
```diff
@@ -0,0 +1,103 @@
...
+ */
+const createPreferencesTable = (preferences) => {
+  const tables = Object.keys(preferences).map((key) => {
+    const header = (
+      <thead key={`${key}_header`}>
+        <tr className="admin-preferences-metrics-list__table-row" key={`${preferences[key]}_header`}>
+          {
+            // Pulling out the values (not the count)
+            Object.values(preferences[key]).map(([field, count]) => (
+              <th aria-label={`${field} ${count} Header`} key={`${field}}`}>{field}</th>
```
trevorlang, line 24: I don't think you need this aria label here. The `th` element can be assumed to be a header. If it's for testing, I think table headers are automatically assigned an aria role that you should be able to select.

static/src/js/components/AdminPreferencesMetrics/AdminPreferencesMetricsList.jsx
```diff
@@ -0,0 +1,103 @@
...
+export const AdminPreferencesMetricsList = ({
+  preferencesMetrics
+}) => {
+  const { preferences, isLoading, isLoaded } = preferencesMetrics
+
+  return (
+    <div>
+      {
+        isLoading && (
+          <div aria-label="Preferences Metrics Spinner">
```
trevorlang, line 77: We should not be adding divs for the sole purpose of testing. We need to find another way to select the spinner if that's the goal of this element.

static/src/js/components/AdminPreferencesMetrics/AdminPreferencesMetricsList.scss
```diff
@@ -0,0 +1,28 @@
+.admin-preferences-metrics-list {
```
trevorlang, line 1: Double check if any of the styles in this file can be added with bootstrap instead of custom css.

static/src/js/components/AdminPreferencesMetrics/AdminPreferencesMetricsList.jsx
```diff
@@ -0,0 +1,103 @@
...
+      </thead>
+    )
+    const body = (
+      <tbody>
+        <tr className="admin-preferences-metrics-list__table-body" key={`${preferences[key]}_body`}>
+          {
+            // Pulling out the counts/percentages of each value
+            // value[0] is the field and value[1] is the count
+            Object.values(preferences[key]).map(([field, count]) => (
+              <td aria-label={`${field} ${count} Body`} key={`${field}_${count}`}>{count}</td>
```
trevorlang, line 37: Same comment about the aria label as above

static/src/js/routes/Admin/Admin.jsx
```diff
@@ -59,16 +60,24 @@ export class Admin extends Component {
                   |
                   {' '}
                   <PortalLinkContainer
-                    to="/admin/retrievalsMetrics"
+                    to="/admin/retrievals-metrics"
```
trevorlang, line 63: Is this url a hyphen or an underscore? I think the serverless.yml had is set to an underscore. Hyphen is more typical, I think.

static/src/js/components/AdminPreferencesMetrics/AdminPreferencesMetricsList.jsx
bnp26: This was specifically brought up by Matthew. He wanted to use an aria-label to more accurately pinpoint the field-values pairs we're checking. @macrouch

static/src/js/routes/Admin/Admin.jsx
macrouch: We seem to be using hyphens in React, and underscores in serverless. So think change is bringing things in line with the rest of the app

trevorlang: That checks out

bnp26: I mean, it's hyphen now. I think if they aren't they should be.

trevorlang (approved):

macrouch (approved):

PR #1803, opened 2024-09-24
EDSC-4079: error status code check change

daniel-zamora:
# Overview

### What is the feature?

When a user is unauthorized, either because they are not logged in or otherwise, and they select granule filters and attempt to downloaded, upon redirect to the project page then subsequent redirect to EDL then again back to the project, the filters they had selected can be dropped.

### What is the Solution?

Refactored the granule download button to send the user to a login page directly when they are trying to download granules while not logged in, instead of relying on failed subsequent calls on the project page that required authentication.

Granule download button.

# Testing

### Reproduction steps

In any environment,

1. Select some granules from download while not logged in.
2. Press the download button to progress to the project page.
3. Once redirected to EDL login, enter in credentials and submit (to be redirected back)
4. Verify your granule filters are intact(or that they are not there if fix is not in).

# Checklist

### DIFF
--- a/static/src/js/components/GranuleResults/GranuleDownloadButton.jsx
+++ b/static/src/js/components/GranuleResults/GranuleDownloadButton.jsx
@@ -8,6 +8,7 @@ import { Tooltip, OverlayTrigger } from 'react-bootstrap'
 import { getApplicationConfig, getEnvironmentConfig } from '../../../../../sharedUtils/config'
 
 import { commafy } from '../../util/commafy'
+import { isLoggedIn } from '../../util/isLoggedIn'
 import { stringify } from '../../util/url/url'
 import { locationPropType } from '../../util/propTypes/location'
 
@@ -33,7 +34,6 @@ export const GranuleDownloadButton = (props) => {
 
   const { disableDatabaseComponents } = getApplicationConfig()
   const { apiHost } = getEnvironmentConfig()
-  const isLoggedIn = authToken !== ''
 
   if (tooManyGranules) {
     return (
@@ -126,7 +126,7 @@ export const GranuleDownloadButton = (props) => {
     variant: 'full'
   }
 
-  if (!isLoggedIn) {
+  if (!isLoggedIn(authToken)) {
     const projectPath = `${window.location.protocol}//${window.location.host}/projects${stringifiedProjectParams}`
 
     return (
--- a/static/src/js/components/GranuleResults/GranuleDownloadButton.jsx
+++ b/static/src/js/components/GranuleResults/GranuleDownloadButton.jsx
@@ -108,40 +108,35 @@ export const GranuleDownloadButton = (props) => {
     }
   }
 
+  const stringifiedProjectParams = stringify({
+    ...params,
+    p,
+    pg
+  })
+
+  const downloadButtonProps = {
+    badge,
+    bootstrapVariant: 'primary',
+    className: 'granule-results-actions__download-all',
+    dataTestId: 'granule-results-actions__download-all-button',
+    disabled: granuleCount === 0 || initialLoading || disableDatabaseComponents === 'true',
+    icon: Download,
+    label: buttonText,
+    type: 'button',
+    variant: 'full'
+  }
+
   if (!isLoggedIn) {
-    const projectPath = `${window.location.protocol}//${window.location.host}/projects${stringify({
-      ...params,
-      p,
-      pg
-    })}`
+    const projectPath = `${window.location.protocol}//${window.location.host}/projects${stringifiedProjectParams}`
 
     return (
       <PortalLinkContainer
-        className="granule-results-actions__download-all"
+        {...downloadButtonProps}
         onClick={
           () => {
-            onAddProjectCollection(focusedCollectionId)
-            onChangePath(`${apiHost}/login?ee=${earthdataEnvironment}&state=${encodeURIComponent(projectPath)}`)
+            window.location.href = `${apiHost}/login?ee=${earthdataEnvironment}&state=${encodeURIComponent(projectPath)}`
           }
         }
-        to={
-          {
-            pathname: '/projects',
-            search: stringify({
-              ...params,
-              p,
-              pg
-            })
-          }
-        }
-        type="button"
-        badge={badge}
-        bootstrapVariant="primary"
-        dataTestId="granule-results-actions__download-all-button"
-        disabled={granuleCount === 0 || initialLoading || (disableDatabaseComponents === 'true')}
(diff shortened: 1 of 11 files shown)


static/src/js/components/GranuleResults/GranuleDownloadButton.jsx
```diff
@@ -30,6 +32,8 @@ export const GranuleDownloadButton = (props) => {
   } = props
 
   const { disableDatabaseComponents } = getApplicationConfig()
+  const { apiHost } = getEnvironmentConfig()
+  const isLoggedIn = authToken !== ''
```
trevorlang, line 36: I see we have a similar check in at least one other spot (static/src/js/components/SearchPanels/SearchPanels.jsx). There we check against an empty string or null. I'd consider creating a util to encapsulate this functionality.

static/src/js/components/GranuleResults/GranuleDownloadButton.jsx
```diff
@@ -104,37 +108,56 @@ export const GranuleDownloadButton = (props) => {
     }
   }
 
+  const stringifiedProjectParams = stringify({
+    ...params,
+    p,
+    pg
+  })
+
+  const downloadButtonProps = {
```
trevorlang, line 117: This might be a personal thing. I am not sure theres a hard line in the sand. But I feel like defining props as an object like this hurts readability. 

We generally have only done this sort of thing only when: a) we have conditional props that need to be omitted in certain cases, or b) when we want to forward an unknown props to a element.

static/src/js/components/GranuleResults/GranuleDownloadButton.jsx
```diff
@@ -104,37 +108,56 @@ export const GranuleDownloadButton = (props) => {
...
+    label: buttonText,
+    type: 'button',
+    variant: 'full'
+  }
+
+  if (!isLoggedIn) {
+    const projectPath = `${window.location.protocol}//${window.location.host}/projects${stringifiedProjectParams}`
+
+    return (
+      <PortalLinkContainer
```
trevorlang, line 133: I am a little confused why we are setting the href directly, rather than using the `href` prop. The purpose of the `PortalLinkContainer` is to make sure that, if the user is currently in a portal, they remain in the portal after clicking the link so make sure you double check theres no regression there. 

If we do end up using the `onClick`, and not setting a `href` or a `to`, I think we could just set a `Button` component here.

static/src/js/components/GranuleResults/GranuleDownloadButton.jsx
macrouch: I suggested this because we had so many props that were exactly the same between the two buttons. I was just trying to remove some of that duplication

trevorlang: or c) when we want to share the same props across conditionally rendered components...

:ill_see_myself_out:

static/src/js/components/GranuleResults/GranuleDownloadButton.jsx
trevorlang: Ignore this. I think I am on the same page now. I just looked back at the component and remembered we added some logic in there that runs after an `onClick`

trevorlang: I think I am still a little confused as to why were setting the location, rather than passing an `href`. Does that not accomplish what we want for some reason?

eudoroolivares2016: @trevorlang I'm not sure if this is what you mean but, I tested the branch with `  <Button
        badge
        href={`${apiHost}/login?ee=${earthdataEnvironment}&state=${encodeURIComponent(projectPath)}`}
        bootstrapVariant="primary"
        className="granule-results-actions__download-all"
        dataTestId="granule-results-actions__download-all-button"
        disabled={granuleCount === 0 || initialLoading || disableDatabaseComponents === 'true'}
        icon={Download}
        label={buttonText}
        type="button"
        variant="full"
      >
        {buttonText}
      </Button>`
      locally and that was successful It does break the look of the button etc so if we were to change to match the one in `static/src/js/components/SecondaryToolbar/SecondaryToolbar.jsx` with a Button there'd be additional work there

eudoroolivares2016 (approved):
Tested it locally using:
- [x] Does this work inside of a portal? (here is the redirected address which includes the portal url param http://localhost:8080/projects?portal=podaac&p=C1996881146-POCLOUD!C1996881146-POCLOUD&pg[1][a]=3252153467!3251508524!POCLOUD&pg[1][v]=t&pg[1][gsk]=-start_date&tl=1727718334!3!!
- [x] Does it work if you are signed in already?
- [x] Does this work if you are not signed in already

static/src/js/components/GranuleResults/GranuleDownloadButton.jsx
trevorlang: I accidentally thought that we were calling `onChangePath` for every `onClick` in a `PortalLinkContainer` but that appears to not be the case. If replacing with a `Button` works, it seems like it would be the preferred method. I'm not sure that the `PortalLinkContainer` is providing any additional behavior in this case. I'm surprised there are styling differences as the `PortalLinkContainer` just ultimately renders a `Button`. Not trying to overcomplicate the PR but I don't think we should be using more complicated components than we need to if possible.

daniel-zamora: I believe this removes an intermediate page from the redirect, and makes the user experience cleaner.

daniel-zamora:
@eudoroolivares2016 I tried using the button component in this way early development on this solution and found it failed to save the state properly, directing to the project page.  It was long enough ago that I do not remember the specifics. It may have been something else causing that.

static/src/js/components/GranuleResults/GranuleDownloadButton.jsx
daniel-zamora: The deduplication here was part of Mathew's feedback when pairing.  I am ambivalent about its use here as I don't experience the readability issue that you do.  It does make it more concise so if I had to choose I'd pick the way it currently is.

static/src/js/util/isLoggedIn.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Returns true if the authToken is not empty
+ * @return {boolean}
```
macrouch, line 3: This jsdoc should list the parameter `@param {Type} paramName Param description`

macrouch (approved):

PR #1809, opened 2024-10-10
EDSC-4108: Adds 'granule' type layers to GIBS tagging job

macrouch:
# Overview

### What is the feature?

When we parse the wv.json file to build our GIBS tags, we need to be processing 'granule' type layers the same way we do 'wmts' type layers.

I also made a change to validateToken to only request a single row from the database, an improvement over requesting many rows and only using the first row. We want to see if this improves performance issues we've been seeing in production.

GIBS tagging/imagery

# Testing

To fully test this process we need to run the code in production, because the collection concept-ids are only found in production

# Checklist

### DIFF
--- a/serverless/src/generateGibsTags/getSupportedGibsLayers.js
+++ b/serverless/src/generateGibsTags/getSupportedGibsLayers.js
@@ -36,7 +36,9 @@ export const getSupportedGibsLayers = async () => {
     const { conceptIds, projections, type } = currentLayer
 
     // Ignore non Web Map Tile Service layers
-    if (type !== 'wmts') {
+    // `granule` type layers are WMTS layers with special properties for Worldview. We need to
+    // process them the same way we do WMTS layers
+    if (type !== 'wmts' && type !== 'granule') {
       return
     }
 
--- a/serverless/src/util/authorizer/validateToken.js
+++ b/serverless/src/util/authorizer/validateToken.js
@@ -48,8 +48,8 @@ export const validateToken = async (jwtToken, earthdataEnvironment) => {
       } = verifiedJwtToken
 
       // Retrieve the authenticated users' access tokens from the database
-      const existingUserTokens = await dbConnection('user_tokens')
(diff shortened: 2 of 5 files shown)


eudoroolivares2016:
We should really add something in the PR description that explains what the db retrieval optimization is doing and briefly why.

serverless/src/util/authorizer/validateToken.js
```diff
@@ -48,8 +48,8 @@ export const validateToken = async (jwtToken, earthdataEnvironment) => {
       } = verifiedJwtToken
 
       // Retrieve the authenticated users' access tokens from the database
-      const existingUserTokens = await dbConnection('user_tokens')
-        .select([
+      const mostRecentToken = await dbConnection('user_tokens')
+        .first([
```
eudoroolivares2016, line 52: Good change we're doing the same thing but, we should offload to SQL to optimize

serverless/src/generateGibsTags/getSupportedGibsLayers.js
```diff
@@ -36,7 +36,9 @@ export const getSupportedGibsLayers = async () => {
     const { conceptIds, projections, type } = currentLayer
 
     // Ignore non Web Map Tile Service layers
-    if (type !== 'wmts') {
+    // `granule` type layers are WMTS layers with special properties for Worldview. We need to
+    // process them the same way we do WMTS layers
+    if (type !== 'wmts' && type !== 'granule') {
```
eudoroolivares2016, line 41:
```suggestion
    
if (!['wmts', 'granule'].includes(type)) {
  return
}
```
but, also this is nitpicky and it really does not matter either way if you think the conditionals are better keep and resolve the comment

eudoroolivares2016 (approved):
I also ran the database login locally with this branch and messed with ee params plus switching out the env in overridestatic and it worked fine

dpesall (approved):

PR #1819, opened 2024-10-31
Fix npm capitalization

friederbluemle:
Just a minor capitalization fix: `NPM` -> `npm`

npm should never be capitalized:
https://github.com/npm/cli#is-it-npm-or-npm-or-npm


macrouch:
Thanks for opening a PR, but we only accept PRs that add something of real value to the project. If you'd like to contribute check out our open issues. If you see one you'd like to work leave a comment on that issue and let us know.

PR #1823, opened 2024-11-05
EDSC-4186: Adding option to permanently hide tour in user preferences

dpesall:
# Overview

### What is the feature?

Added an option in user preferences to permanently hide the tour on page load, and added a checkbox to the first slide of the tour to update this preference from the tour itself.

### What is the Solution?

Updating the first slide of the tour and user preferences route to support the new preference option.

- Search route
- Preferences route

# Testing

### Reproduction steps

1. Load page while logged in and localStorage cleared
2. Verify that the tour loads
3. Check the box, close the tour, and go to user preferences
4. Verify user preferences were updated from default to your selected preference from the checkbox
5. Repeat these steps but this time uncheck the box

### Attachments

Screenshot 2024-11-05 at 2 11 33 PM
Screenshot 2024-11-05 at 2 11 58 PM

# Checklist

### DIFF
--- a/static/src/js/actions/preferences.js
+++ b/static/src/js/actions/preferences.js
@@ -64,7 +64,7 @@ export const setPreferencesFromJwt = (jwtToken) => (dispatch) => {
   }
 }
 
-export const updatePreferences = (data) => (dispatch, getState) => {
+export const updatePreferences = (data, showToast = true) => (dispatch, getState) => {
   const { formData: preferences } = data
 
   const state = getState()
@@ -90,10 +90,12 @@ export const updatePreferences = (data) => (dispatch, getState) => {
       dispatch(updateAuthTokenFromHeaders(headers))
       dispatch(setPreferences(newPreferences))
       dispatch(setIsSubmitting(false))
-      addToast('Preferences saved!', {
-        appearance: 'success',
-        autoDismiss: true
-      })
+      if (showToast) {
+        addToast('Preferences saved!', {
+          appearance: 'success',
+          autoDismiss: true
+        })
+      }
     })
     .catch((error) => {
       dispatch(setIsSubmitting(false))
--- a/static/src/js/actions/preferences.js
+++ b/static/src/js/actions/preferences.js
@@ -77,7 +77,6 @@ export const updatePreferences = (data) => (dispatch, getState) => {
   dispatch(setIsSubmitting(true))
 
   const requestObject = new PreferencesRequest(authToken, earthdataEnvironment)
-
   const response = requestObject.update({ preferences })
     .then((responseObject) => {
       const {
@@ -87,7 +86,6 @@ export const updatePreferences = (data) => (dispatch, getState) => {
       const {
         preferences: newPreferences
       } = dataObject
-
       dispatch(updateAuthTokenFromHeaders(headers))
       dispatch(setPreferences(newPreferences))
       dispatch(setIsSubmitting(false))
--- a/static/src/js/actions/preferences.js
+++ b/static/src/js/actions/preferences.js
@@ -86,6 +86,7 @@ export const updatePreferences = (data) => (dispatch, getState) => {
       const {
         preferences: newPreferences
       } = dataObject
+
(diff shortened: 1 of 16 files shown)


static/src/js/providers/TourContextProvider/TourContextProvider.jsx
```diff
@@ -11,13 +11,11 @@ const { disableSiteTour } = getApplicationConfig()
...
 const TourContextProvider = ({ children }) => {
-  const [runTour, setRunTour] = useState(false)
+  const [runTour, setRunTour] = useState(0)
   useEffect(() => {
     const isLocalhost = window.location.hostname === 'localhost'
     const overrideLocalhost = window.overrideLocalhost === true
-    const hasUserDisabledTour = localStorage.getItem('dontShowTour') === 'true'
-    const shouldShowTour = isSiteTourEnabled && !hasUserDisabledTour
-      && (!isLocalhost || overrideLocalhost)
+    const shouldShowTour = isSiteTourEnabled && (!isLocalhost || overrideLocalhost) ? 1 : 0
```
eudoroolivares2016, line 18: I don't like the magic number even if its a bool I don't understand what this change is for I'd prefer to just move the logic for determining if we are going to showTour into its own function that just returns true or false

static/src/js/components/Preferences/PreferencesForm.jsx
```diff
@@ -23,6 +23,8 @@ const PreferencesForm = (props) => {
     preferences: formDataProps
   } = preferences
 
+  console.log(preferences)
```
eudoroolivares2016, line 26: remove me

static/src/js/actions/preferences.js
```diff
@@ -87,7 +86,6 @@ export const updatePreferences = (data) => (dispatch, getState) => {
       const {
         preferences: newPreferences
       } = dataObject
-
```
eudoroolivares2016, line 90: I think spacing was fine before no need to compress it all

eudoroolivares2016 (commented):
We do need to also add some more tests such as in `static/src/js/components/Preferences/__tests__/PreferencesForm.test.js`

static/src/js/providers/TourContextProvider/TourContextProvider.jsx
```diff
@@ -11,13 +11,11 @@ const { disableSiteTour } = getApplicationConfig()
 const isSiteTourEnabled = disableSiteTour === 'false'
 
 const TourContextProvider = ({ children }) => {
-  const [runTour, setRunTour] = useState(false)
+  const [runTour, setRunTour] = useState(0)
   useEffect(() => {
     const isLocalhost = window.location.hostname === 'localhost'
```
eudoroolivares2016, line 16: Might be nice if we pulled some of these into `static/src/js/util/url/url.js` since they're just one liners and pretty specific to what we have to let Joyride know perhaps its not worth it

static/src/js/providers/TourContextProvider/TourContextProvider.jsx
dpesall: I don't love this solution myself, but the reason I went with 1 and 0 here, is because the showing of the tour is contingent on the user preference, but we can't access the redux store inside the context provider because at this point the redux store hasn't been initialized.

So the runTour context is set to a number (0 or 1) first, and if it is set to a zero or one, that's how we can tell we need to look at the redux store once it's been initialized which we do here:

`useEffect(() => {
    if (runTour === 1) {
      setRunTour(!getDefaultCheckboxValue())
    } else if (runTour === 0) {
      setRunTour(false)
    }
  }, [runTour, setRunTour])

  const shouldRunTour = () => runTour`
  
If we just return true/false from the `useContext`, we have no way to tell if that value came from the initial state. Because there's a situation where a logged in user has nothing in localstorage, so the `runTour` context defaults to true, but if they have the tour set to `dontshow` in their preferences, we run into issues with not knowing we need to update the tour context again. I can explain this more in detail in the PR walk through but that's the reason I went with this approach.

PR #1827, opened 2024-11-08
EDSC-4246: Show the EDD button when a Harmony order is complete_with_errors and has links

macrouch:
# Overview

### What is the feature?

Harmony orders can complete with errors but still have valid links. When this happens we are not showing the Earthdata Download download button.

### What is the Solution?

Change our logic to allow for complete_with_errors jobs that do have links to show the EDD button.

Harmony orders

# Testing

### Reproduction steps

This job has a bounding box that doesn't work for the a granule within the applied temporal. Ensure that both temporal and spatial subsetting are selected.

http://localhost:8080/projects?p=C2670138092-NSIDC_CPRD!C2670138092-NSIDC_CPRD&pg[1][v]=t&pg[1][gsk]=-start_date&pg[1][m]=harmony0&pg[1][cd]=f&pg[1][ets]=t&pg[1][ess]=t&q=ATL06&sb[0]=-50.99141%2C68.31574%2C-47.56861%2C69.91401&qt=2019-02-22T19%3A47%3A47.080Z%2C2019-04-11T10%3A04%3A08.643Z&ff=Customizable&tl=1553331766!3!!

Locally you will have to submit the job manually, and fetch updates manually after the job is completed to get your order into the correct state.

Once in complete_with_errors with other valid links you should see the "Download Files" button

### Attachments

Completed successfully
Screenshot 2024-11-08 at 2 40 38 PM

In Progress
Screenshot 2024-11-12 at 5 01 13 PM

Completed with errors but with links
Screenshot 2024-11-08 at 2 40 53 PM

# Checklist

### DIFF
--- a/static/src/js/components/OrderStatus/OrderStatusItem.jsx
+++ b/static/src/js/components/OrderStatus/OrderStatusItem.jsx
@@ -154,14 +154,15 @@ export class OrderStatusItem extends PureComponent {
       retrieval_collection_id: retrievalCollectionId
     } = collection
 
+    const orderStatus = aggregatedOrderStatus(orders)
+
     const [firstOrder = {}] = orders
     const {
-      state,
       type = ''
     } = firstOrder
 
     // If the order is Harmony and is still running or has no files, don't show the EDD link
-    const isDone = ['successful', 'complete_with_errors'].includes(state)
+    const isDone = !['creating', 'in progress'].includes(orderStatus)
     const notDoneOrEmpty = !isDone || downloadUrls.length === 0
     if (type.toLowerCase() === 'harmony' && notDoneOrEmpty) {
       return null
--- a/static/src/js/components/OrderStatus/OrderStatusItem.jsx
+++ b/static/src/js/components/OrderStatus/OrderStatusItem.jsx
@@ -689,15 +689,16 @@ export class OrderStatusItem extends PureComponent {
                       >
                         <DownloadFilesPanel
                           accessMethodType={accessMethodType}
+                          collectionIsCSDA={collectionIsCSDA}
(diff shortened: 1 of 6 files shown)


eudoroolivares2016 (approved):

macrouch:
Can we look at disabling the Download button while the job is running instead of hiding it?

static/src/js/components/OrderStatus/OrderStatusItem.jsx
```diff
@@ -154,14 +154,19 @@ export class OrderStatusItem extends PureComponent {
...
     const [firstOrder = {}] = orders
     const {
-      state,
       type = ''
     } = firstOrder
 
-    // If the order is Harmony and isn't successful, don't show the EDD link
-    if (type.toLowerCase() === 'harmony' && state !== 'successful') return null
+    // If the order is Harmony and is still running or has no files, don't show the EDD link
+    const isDone = !['creating', 'in progress'].includes(orderStatus)
```
daniel-zamora, line 165: should these be constants?

eudoroolivares2016: Theres a const here `sharedUtils/orderStatus.js` but, they aren't specific to Harmony these are apparently the different states that harmony can be in `https://github.com/nasa/harmony/blob/main/db/db.sql#L8`

macrouch: Yes, it should be constants, but I'm not really sure where to start. That orderStatus util @eudoroolivares2016 pointed out has a ton of string values that would be much better served as constants. This file is the same way. 

Rather than working that within this ticket and the scope exploding in size I want to write a new ticket with the goal of cleaning up string values like these.

daniel-zamora (approved):

PR #1836, opened 2024-11-21
EDSC-4251: Fixed progress bar for SWODLR jobs

bnp26:
# Overview

### What is the feature?

Fixed it so that complete SWODLR jobs now show the progress bar at 100%.

### What is the Solution?

Added logic to account for the order being a SWODLR order.

Progress bar in the downloads page.

# Testing

### Reproduction steps

- **Environment for testing:** sit
- **Collection to test with:** C2799438271-POCLOUD

1. create a SWODLR project and add 2 granules 
2. click "Download"
3. on the downloads page wait for the job to complete
4. observe the blue progress bar go to 100% when it does complete.

### Attachments

image
image
image

# Checklist


eudoroolivares2016:
Tested in SIT as feature branch on Friday Nov 22nd

static/src/js/components/OrderProgressList/__tests__/OrderProgressItem.test.js
```diff
@@ -1,57 +1,121 @@
...
 beforeEach(() => {
   jest.clearAllMocks()
 })
 
-Enzyme.configure({ adapter: new Adapter() })
+const setup = (type, overrideProps) => {
+  let orderOrigin = null
+  let props = {}
+
+  switch (type) {
```
macrouch, line 21: A switch statement feels overly complicated here. There are two different times orderOrigin gets set to the ESI order which is difficult to track down the reason. And after setting orderOrigin you just pull the order out to pass in as props. 

Instead of writing this logic, just pass in the order to want to set as props for each test

static/src/js/components/OrderProgressList/__tests__/OrderProgressItem.test.js
```diff
@@ -1,57 +1,121 @@
...
+  describe('Complete Swodlr Order', () => {
+    test('shows the correct order metadata', () => {
+      setup('SWODLR')
+
+      expect(screen.getByRole('heading', {
+        level: 5,
+        name: 'Order ID: e7efe743-f253-43e3-b017-74faa8bdfcf1'
+      })).toBeInTheDocument()
+
+      expect(screen.getByText('Complete')).toBeInTheDocument()
```
macrouch, line 65: We need better assertions than just 'Complete' shows up on the page. Where does it show up on the page, or under what context does it show up on the page?

static/src/js/components/OrderProgressList/__tests__/OrderProgressItem.test.js
```diff
@@ -1,57 +1,121 @@
...
+import { render, screen } from '@testing-library/react'
+
 import { OrderProgressItem } from '../OrderProgressItem'
 
-import { retrievalStatusPropsEsi } from '../../OrderStatus/__tests__/mocks'
+import {
+  retrievalStatusPropsEsi,
+  retrievalStatusPropsSwotOrder,
+  retrievalStatusPropsHarmonyOrder
+} from '../../OrderStatus/__tests__/mocks'
```
macrouch, line 11: It feels weird to borrow mocks from another component. Data needed for these tests should be located in the same place as these tests. 

If this test wasn't reusing data designed for another component we wouldn't need the additional logic to dig into the test data to pull out only the order this test needs.

static/src/js/components/OrderProgressList/__tests__/OrderProgressItem.test.js
```diff
@@ -1,57 +1,121 @@
...
 
-import { Badge, ProgressBar } from 'react-bootstrap'
+import { render, screen } from '@testing-library/react'
+
 import { OrderProgressItem } from '../OrderProgressItem'
 
-import { retrievalStatusPropsEsi } from '../../OrderStatus/__tests__/mocks'
+import {
+  retrievalStatusPropsEsi,
+  retrievalStatusPropsSwotOrder,
```
macrouch, line 9: Should be named with `Swodlr`, not `Swot`, to match the other names that use the access method type

static/src/js/components/OrderProgressList/__tests__/OrderProgressItem.test.js
eudoroolivares2016: @bnp26 One thing you can do is get the element and then do something like `element.parentElement` and use `within`

bnp26: That's bad practice to do that. React testing library doesn't like it when you use DOM like that. Better to add the appropriate role to the elements and look for them that way.

static/src/js/components/OrderProgressList/__tests__/OrderProgressItem.test.js
```diff
@@ -1,57 +1,115 @@
...
   }
 }
 
 describe('OrderProgressItem component', () => {
-  test('shows the correct order metadata', () => {
-    const { enzymeWrapper } = setup()
+  describe('Complete Swodlr Order', () => {
+    test('shows the correct order metadata', () => {
+      setup({
+        order: {
```
macrouch, line 35:
```suggestion
        order: retrievalStatusPropsSwodlrOrder
```

All of your mocks used in this file have this chain of `.retrieval.collections.byId[1].orders[0]`.

Mock data should only include the actual data for the mock, everything else is unnecessary and can lead to confusion later.

macrouch (approved):

PR #1837, opened 2024-11-25
EDSC-4166: Adding metric event for Temporal filtering

dpesall:
# Overview

### What is the feature?

Adding a metric event to be tracked whenever the user makes a change to the Temporal filter.

### What is the Solution?

Added a new metric action and event to send the payload to our metrics whenever the temporal filter is changed.

- Temporal Filter
- Metric Events

# Testing

### Reproduction steps

1. Open the Temporal filter
2. Set Start Date
3. Set End Date
4. Toggle and untoggle 'Recurring?'
5. Click Apply
6. Reopen and click Clear
7. Validate that all previous events were logged to google tag manager

# Checklist


static/src/js/middleware/metrics/events.js
```diff
@@ -215,6 +215,25 @@ export const spatialSelection = (action) => {
...
+  const { payload } = action
+  const {
+    type,
+    value
+  } = payload
+
+  dataLayer.push({
+    event: 'temporalFilter',
+    granuleFilterCategory: 'Temporal Filter',
+    granuleFilterEventAction: type,
```
eudoroolivares2016, line 232: names of these need to be updated

static/src/js/middleware/metrics/__tests__/events.test.js
```diff
@@ -407,6 +409,28 @@ describe('events', () => {
...
+          value: 'Some Value'
+        }
+      }
+
+      temporalFilter(action)
+
+      expect(dataLayerMock).toHaveBeenCalledTimes(1)
+      expect(dataLayerMock).toHaveBeenCalledWith({
+        event: 'temporalFilter',
+        granuleFilterCategory: 'Temporal Filter',
```
eudoroolivares2016, line 427: names of these update

static/src/js/components/AdminRetrievalsMetrics/AdminRetrievalsMetrics.jsx
```diff
@@ -76,6 +77,7 @@ export const AdminRetrievalsMetrics = ({
           <TemporalSelectionDropdown
             onChangeQuery={onChangeQuery}
             allowRecurring={false}
+            onMetricsTemporalFilter={onMetricsTemporalFilter}
```
eudoroolivares2016, line 80: The only issue with adding this here is that we might be polluting the metrics because if we (the admins use the temporalSelectionDropdown its going to be counted on GA. It would be a pretty small subset of the total usage though so if it adds significant complexity to not use this then perhaps our best coarse is to use it like this.
The only alternative way I could see this working in a simple manner here might be to pass a stub to the component but, that does seem pretty hacky
```suggestion
            onMetricsTemporalFilter={someFuncThatDoesNothing}
```

static/src/js/components/TemporalDisplay/TemporalSelectionDropdown.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-underscore-dangle */
```
eudoroolivares2016, line 1: I get moving this up reduces copying but, I think that when we ignore a `linting` rule its good to just have that next to where we are breaking the rule

macrouch: I agree, ignoring a linting rule for the whole file is an easy way to ignore the rule in the future without realizing it. Every time we want to ignore a rule we need to make a conscious decision to do it, so only use `eslint-disable-next-line`

static/src/js/components/AdminRetrievalsMetrics/AdminRetrievalsMetrics.jsx
macrouch: I think we could either have something in the event tell us which kind of temporal is being set. Or we could just ignore the admin pages. 

If you want to ignore the admin pages I wouldn't use a stub function, but just make the callback not required in TemporalSelectionDropdown, and only call it if it exists

eudoroolivares2016: I think ignoring the admin pages does make the most sense I remember we talked about doing that during the design review. Making the callback not required and only calling it if its there does make the most sense to me

static/src/js/containers/DatepickerContainer/__tests__/DatepickerContainer.test.js
```diff
@@ -280,4 +291,36 @@ describe('DatepickerContainer component', () => {
       expect(button).toHaveClass('rdtDisabled')
     })
   })
+
+  describe('when focusing the input', () => {
```
macrouch, line 295: I don't think we need these new tests. `describe('onTodayClick'`, `describe('onClearClick'` and `describe('when typing in the text field` should add assertions that `onMetricsTemporalFilter` was called as intended.

And `describe('when typing in the text field'` should have a new test where you click into the input and tab out of it and assert that `onMetricsTemporalFilter` was not called

static/src/js/components/TemporalDisplay/__tests__/TemporalSelectionDropdown.test.js
```diff
@@ -291,4 +309,321 @@ describe('TemporalSelectionDropdown component', () => {
...
+      await user.click(recurringCheckbox)
+    })
+
+    expect(onMetricsTemporalFilterSpy).toHaveBeenCalledTimes(0)
+  })
+
+  describe('setStartDate', () => {
+    beforeEach(() => {
+      TemporalSelectionDropdownMenu.mockImplementation(
+        ({ setStartDate, setEndDate }) => {
```
macrouch, line 435: This exact component is defined twice, here and the beforeEach in the describe 'setEndDate'. You can either:

1. Pull the whole component out and define it once at the top of the file. Then reference the component in both beforeEach blocks
2. Remove the end date buttons from this component and remove the start date buttons from the describe 'setEndDate' component

macrouch (approved):

static/src/js/components/TemporalDisplay/__tests__/TemporalSelectionDropdown.test.js
```diff
@@ -291,4 +349,219 @@ describe('TemporalSelectionDropdown component', () => {
 
     expect(onChangeQueryMock).toHaveBeenCalledTimes(1)
   })
+
+  test('onMetricsTemporalFilter is called on Apply', async () => {
+    const onMetricsTemporalFilterMock = jest.fn()
+    const onChangeQueryMock = jest.fn()
+    const user = userEvent.setup()
+
+    setup({
```
eudoroolivares2016, line 358:
```suggestion
    const {user } = setup({
```
We can do the `const user = userEvent.setup()` in the setup and just pull it out as a prop for all these

dpesall: This doesn't work because the `setup()` method in this test doesn't return `user`. In this file it returns:
`return render(
    <Provider store={store}>
      <TemporalSelectionDropdown {...props} />
    </Provider>
  )`
  
  This works in other test files though like `DatepickerContainer.test.js` where `setup()` returns:
  `return {
    props,
    user
  }`
  
  I think we could probably rewrite the setup method in this test, but that would involve rewriting all the other tests as well.

eudoroolivares2016 (approved):
I also tested the branch locally against GTM and the events looked good also it did not fire an event on the admin page using `DateTime` as expected

PR #1841, opened 2024-12-03
EDSC-4125: Recurring temporal slider doesn't work

daniel-zamora:
# Overview

### What is the feature?

On the granule temporal filter there are some issues that arise with the recurring toggle.  If start and end have the same year, the InputRange handlers overlap and are unmoveable.  When recurring is toggled on, you can still change year values with the date picker.  And when you scrub the InputRange handlers left or right, on each step change it will send a search request to CMR

### What is the Solution?

When the same year is selected for start and end via the datepicker, and recurring is toggled on, we max out the range from 1960 to w/e the end date is

When recurring is toggled on, the date picker will not allow you to traverse outside the bounds of a single year's worth of months(i.e. you cannot go to the previous month when on January or the next month after December) and the year is no longer displayed on the column header when in days mode.  It prevents traversing by disabling the navigation buttons.

CMR searches are now only sent in the onChangeComplete callback, not onChange

Datepicker, TemporalSelection, and GranuleFilterForm components

# Testing

### Reproduction steps

1. Select a Collection and filter granules by temporal filter
2. Select a start and end with the same year
3. Toggle recurring on.
4. Attempt to move the InputRange handle.

### Attachments

# Checklist


daniel-zamora:
in draft, just running tests

eudoroolivares2016:
I'm seeing a behavior on this branch where on the temporal filter (the top level collection filter I mean) the recurring slider if enabled is only able to include the years so it cannot actually be moved vs on the granule filter we can still slide back to previous or post times

static/src/js/components/TemporalSelection/TemporalSelection.jsx
```diff
@@ -29,6 +29,7 @@ import DatepickerContainer from '../../containers/DatepickerContainer/Datepicker
  * @param {Function} props.onSubmitEnd - Callback function to call when a submission ends
  * @param {Function} props.onSubmitStart - Callback function to call when a submission starts
  * @param {Function} props.onValid - Callback function to call when the entry is valid
+ * @param {Function} props.setFieldValue - Callback function provided by Formik.
```
eudoroolivares2016, line 32: Update this with the new function since the last commit

static/src/js/components/TemporalSelection/__tests__/TemporalSelection.test.js
```diff
@@ -23,6 +23,7 @@ function setup() {
     onSubmitEnd: jest.fn(),
     onInvalid: jest.fn(),
     onValid: jest.fn(),
+    setFieldValue: jest.fn(),
```
eudoroolivares2016, line 26: Think this is still from the previous commit

static/src/js/components/Datepicker/__tests__/Datepicker.test.js
```diff
@@ -215,4 +216,125 @@ describe('Datepicker component', () => {
...
+
+        // January state in days view
+        await waitFor(() => {
+          expect(currentMonthSwitch).toHaveTextContent('January')
+          expect(currentMonthSwitch).not.toHaveTextContent('2024')
+          expect(currentPrevNav).not.toBeVisible() // Hidden in January
+          expect(currentNextNav).toBeVisible()
+        })
+
+        // Navigate to December using next nav
```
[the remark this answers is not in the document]
daniel-zamora: linter will complain about awaiting in a loop

bnp26:
@daniel-zamora What do you think the difficulty of potentially converting Datepicker.jsx to a functional component?

trevorlang (commented):
Definitely seeing some nice improvement here. One thing I noticed though, if a user does not have any dates selected when checking "Recurring" they end up in the state where the start and end slider handles are overlapping and the range cannot be changed. It would be nice if instead, the entire range was selected.

trevorlang (commented):
I find it a little odd that once you select "Recurring?" the date inputs fill with what seems to be the beginning of the year and the current date/time. Any way we can keep them empty? 

Also, I'm just noticing that it looks like the "Apply" button is not disabling when the fields are empty, which it should be. Not sure if this is a bug or just something we need to improve. Looks like its the same in production. I don't want to increase the scope of this effort but that would be a great improvement. We can file a ticket and follow up with it if thats the best path forward.

dpesall (approved):

trevorlang (approved):

static/src/js/components/GranuleFilters/GranuleFiltersForm.jsx
```diff
@@ -528,18 +523,20 @@ export const GranuleFiltersForm = (props) => {
                 }
                 onSubmitStart={
                   (startDate, shouldSubmit) => {
+                    const { temporal: newTemporal } = values
+                    if (newTemporal.isRecurring) {
```
macrouch, line 527: Add a comment explaining why this logic is necessary

static/src/js/components/GranuleFilters/GranuleFiltersForm.jsx
```diff
@@ -528,18 +523,20 @@ export const GranuleFiltersForm = (props) => {
...
+                      const existingStartDate = moment(values.temporal.startDate)
```
macrouch, line 528: You already have `values.temporal` available in `newTemporal`

macrouch (changes requested):
Found another issue where the year is being changed.

1. Select a start and end date, the end date should not be 2025.
2. Check the recurring checkbox.
3. Pick a new end date.

The year in the recurring range will be changed to 2025.

Alternate:

1. Select a start and end date.
2. Check the recurring checkbox.
3. Drag the range slider's end year to be earlier than 2025.
4. Pick a new end date.

The year in the recurring range will be changed to 2025.

macrouch (approved):

PR #1844, opened 2024-12-13
EDSC-4076: Integrate a linter for react-testing-library before the number of those gets too large in EDSC

daniel-zamora:
# Overview

### What is the feature?

Adds react testing library eslint plugin to edsc

### What is the Solution?

Added edsc specific configs to the lint config, the shared linter config updated version, and updated all the tests that were experiencing new linter errors

Component tests that use react testing library.

# Testing

### Reproduction steps

npm run lint

### Attachments

# Checklist


.eslintrc
```diff
@@ -33,12 +37,45 @@
...
+    },
+    {
+      "files": [
+        "tests/**/*.js",
+        "serverless/**/*",
+        "static/src/js/components/**/[^_]*/*.jsx",
+        "static/src/js/**/*.js"
+      ],
+      "rules": {
+        "testing-library/await-async-events": "off",
```
eudoroolivares2016, line 49: Should these be done in the linter package? otherwise EDSC might have a different setup than say MMT. Some of these feel like they aren't bad to ignore though I understand that it would make this PR even larger some comments on why we're doing these would be good to have though

eudoroolivares2016 (commented):
We do need to figure out why code-cov is missing some files

.eslintrc
daniel-zamora: I think that is the point of the eslintrc, not all repos will want this, we need to exclude tests that haven't been converted from enzyme etc.

macrouch: Add a comment explaining why these rules are being ignored

static/src/js/components/CollapsePanel/__tests__/CollapsePanel.test.jsx
```diff
@@ -28,8 +28,8 @@ const setup = () => {
 
 describe('CollapsePanel component', () => {
   test('renders itself correctly', () => {
-    const { container } = setup()
-    expect(container.firstChild).toHaveClass('collapse-panel test-wrap-class')
+    setup()
```
macrouch, line 31: If no other tests are using the container, don't set it on line 19 and return it on line 22

static/src/js/components/EDSCImage/__tests__/EDSCImage.test.jsx
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable testing-library/no-node-access */
```
macrouch, line 1: Why are these rules ignored? Does the eslintrc rules cover them, or do they need to be addressed?

If they do have to be ignored, we should use the `// eslint-disable-next-line` syntax only on the lines that need it. In general this means we don't accidentally write code in a file that gets ignored but could have been fixable

static/src/js/components/Facets/Facets.jsx
```diff
@@ -55,6 +55,7 @@ const Facets = (props) => {
     featuresFacet.children.push({
       applied: featureFacets.customizable,
       title: 'Customizable',
+      'data-testid': 'customizable-info',
```
macrouch, line 58: Is this being used in the tests?

static/src/js/components/GranuleResults/GranuleResultsDownloadNotebookButton.jsx
```diff
@@ -101,8 +103,15 @@ export const GranuleResultsDownloadNotebookButton = ({
           <Dropdown.Menu
             ref={dropdownMenuRef}
             className="granule-results-download-notebook-button__menu"
+            data-testid="dropdown-menu"
```
macrouch, line 106: We should avoid using test ids. In this example I don't think it is necessary

```
        const dropdownMenu = screen.getByTestId('dropdown-menu')
        const downloadButton = within(dropdownMenu).getByRole('button', { name: 'Download Notebook' })
```

Does `screen.getByRole('button', { name: 'Download Notebook' })` work without needing the test id?

static/src/js/components/GranuleResults/__tests__/GranuleResultsFocusedMeta.test.jsx
```diff
@@ -870,14 +870,17 @@ describe('GranuleResultsFocusedMeta component', () => {
               await user.click(modalPrev)
             })
 
+            let images
+            let pagination
             await waitFor(() => {
-              const images = within(modal).queryAllByTestId('mock-edsc-image')
-              const pagination = within(modal).queryByText('2/2')
+              images = within(modal).queryAllByTestId('mock-edsc-image')
+              pagination = within(modal).queryByText('2/2')
```
macrouch, line 877: I think this line should also be moved below the `waitFor`

static/src/js/components/GranuleResults/__tests__/GranuleResultsItem.test.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable testing-library/no-unnecessary-act */
```
macrouch, line 1: Is this covered by eslintrc? If not, use `// eslint-disable-next-line` instead of disabling for the whole file

static/src/js/components/Panels/PanelItem.jsx
```diff
@@ -135,6 +135,9 @@ export class PanelItem extends Component {
       if (scrollable) {
         return (
           <SimpleBar
+            data-testid="simplebar-content-wrapper"
```
macrouch, line 138: This doesn't seem necessary with both a role and a label on the component

static/src/js/components/ProjectCollections/skeleton.jsx
```diff
@@ -51,22 +51,25 @@ export const projectHeader = [
     top: 6,
     height: 12,
     width: 80,
-    radius: 2
+    radius: 2,
+    'data-testid': 'skeleton-item'
```
macrouch, line 55: Are these being used?

static/src/js/components/TemporalDisplay/__tests__/TemporalSelectionDropdown.test.jsx
```diff
@@ -100,19 +100,19 @@ describe('TemporalSelectionDropdown component', () => {
...
     setup()
 
-    expect(screen.queryByRole('button')).toBeInTheDocument()
+    expect(screen.getByRole('button')).toBeInTheDocument()
     expect(screen.queryByLabelText('Start')).not.toBeInTheDocument()
   })
 
   test('when clicked toggles the state of show ', async () => {
-    const user = userEvent.setup()
+    const { click } = userEvent.setup()
```
macrouch, line 108: Not sure why this change was made. But one thing we've done in other repos is call `userEvent.setup()` first in the `setup` function, then return the `user` object. So the test ends up looking like this `const { user } = setup({})`. That avoids having `userEvent.setup()` called 18 times in this file

.eslintrc
trevorlang: I could use a little background as to why all of these rules are being ignored. A comment explaining might be enough 😄

static/src/js/components/ProjectCollections/skeleton.jsx
daniel-zamora: Looks like ProjectHeader line 78 uses this.

static/src/js/components/TemporalDisplay/__tests__/TemporalSelectionDropdown.test.jsx
daniel-zamora: Hrmm, when I ran the linter it complained that this needed destructuring, but now it doesn't. I will change them back.

daniel-zamora: Ahh, got it to complain again. 

```/Users/dpzamora/lab/cmr/earthdata-search/static/src/js/components/TemporalDisplay/__tests__/TemporalSelectionDropdown.test.jsx
  108:11  error  `user` is not a recommended name for `render` returned value. Instead, you should destructure it, or name it using one of: `view`, or `utils`  testing-library/render-result-naming-convention````

daniel-zamora: I'll try it your way and see what happens.

macrouch (approved):

eudoroolivares2016 (approved):

PR #1855, opened 2025-02-07
EDSC-4346: Make the entire map to be visible when I select data

dpesall:
# Overview

### What is the feature?

- Current behavior allows granules to be partially obscured by the side panel since it overlaps the map.

### What is the Solution?
- Updating the map behavior to be rendered to the right of the panel rather than underneath.

- EDSC Map
- Side Panels

# Testing

### Reproduction steps

1. On the EDSC home page, select a granule in a collection.
2. Verify that the entire granule swath is visible.
3. Drag the side bar back and forth and ensure the map width updates correctly.
4. Collapse/expand the side panel and ensure the map width updates correctly.

# Checklist


trevorlang (changes requested):
I have some concerns about the rendering performance when changing the width of the panel. When comparing the changes to the old panel, it is significantly slower/jankier. I found a option on `invalidateSize`  called `debounceMoveend` which helps a little but still not the best. We need to do everything we can to try and reach the level of performance of the current implementation.

I noticed you're dynamically setting a variable to control the left and the width with a css variable. A few things there:

1. We do not use css variables this way anywhere else in the app, so while valid, I don't think we should introduce a new pattern if we don't need to. It would be preferable to set the dynamic css with the `style` attribute on the element like we do elsewhere in the app.
2. You do not need to set both `left` and `width` properties on the the `.map` element. Setting only width would be be preferable. 
3. Setting `will-change: width` on the `.map` container could improve performance.

trevorlang:
Playing with things locally, I was able to achieve much better performance using by setting the `width` style property directly on the element as the panel is resized. There might be further improvements to make but that seemed to help a lot.

trevorlang:
This PR can be closed now

PR #1864, opened 2025-02-20
EDSC-4409: Remove the place labels from Earthdata Search

dpesall:
# Overview

### What is the feature?

Removing the Place Labels from the map on Earthdata Search.

### What is the Solution?

- Removed the option to add/remove Place Labels on the map so it longer shows place labels.
- Removed Place Labels as an option in User Preferences.

- EDSC Map
- Leaflet Toolbar
- User Preferences

# Testing

1. Navigate to EDSC home page
2. Verify that Place Labels are not showing anywhere on the map
3. Verify that the Leaflet Toolbar does not have the 'Place Labels' checkbox available beneath 'Borders and Roads' and 'Coastlines'
4. Navigate to User Preferences
5. Verify that 'Place Labels' are no longer present as an option under 'Overlay Layers'

# Checklist


static/src/js/util/url/mapEncoders.js
```diff
@@ -173,8 +171,7 @@ export const decodeMap = (params) => {
   if (overlaysParam) {
     decodedOverlays = {
       referenceFeatures: overlaysParam.split(',').indexOf('referenceFeatures') !== -1,
-      coastlines: overlaysParam.split(',').indexOf('coastlines') !== -1,
-      referenceLabels: overlaysParam.split(',').indexOf('referenceLabels') !== -1
```
eudoroolivares2016, line 177: If we do keep this change we'll want to make sure we update https://wiki.earthdata.nasa.gov/display/EDSC/Earthdata+Search+URL+Parameters when it hits prod

PR #1865, opened 2025-02-21
EDSC-4409: Remove the place labels from Earthdata Search

dpesall:
# Overview

### What is the feature?

Removing the Place Labels from the map on Earthdata Search.

### What is the Solution?

- Removed the option to add/remove Place Labels on the map so it longer shows place labels.
- Removed Place Labels as an option in User Preferences.

- EDSC Map
- Leaflet Toolbar
- User Preferences

# Testing

1. Navigate to EDSC home page
2. Verify that Place Labels are not showing anywhere on the map
3. Verify that the Leaflet Toolbar does not have the 'Place Labels' checkbox available beneath 'Borders and Roads' and 'Coastlines'
4. Navigate to User Preferences
5. Verify that 'Place Labels' are no longer present as an option under 'Overlay Layers'

# Checklist

### DIFF
--- a/static/src/js/containers/MapContainer/MapEvents.jsx
+++ b/static/src/js/containers/MapContainer/MapEvents.jsx
@@ -1,6 +1,11 @@
-import { useLayoutEffect } from 'react'
+import React, {
+  useState,
+  useCallback,
+  useLayoutEffect
+} from 'react'
 import PropTypes from 'prop-types'
 import { useMap, useMapEvents } from 'react-leaflet'
+import { Overlay, Tooltip } from 'react-bootstrap'
 
 const MapEvents = (props) => {
   const map = useMap()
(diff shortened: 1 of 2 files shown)


eudoroolivares2016 (approved):
@dpesall Addressed my questions offline. Feature looks good locally

static/src/js/containers/MapContainer/MapEvents.jsx
```diff
@@ -10,9 +15,44 @@ const MapEvents = (props) => {
...
+    setShowTooltip(true)
+    setTooltipTarget(event.currentTarget)
+  }, [])
+
+  const handleMouseOut = useCallback(() => {
+    setShowTooltip(false)
+    setTooltipTarget(null)
+  }, [])
+
   useLayoutEffect(() => {
```
trevorlang, line 31: Since we've added some event listeners, we need to make sure to provide a cleanup callback function in a return statement at the end of the function. 

Should end up looking like:
```
return () => {
    control.removeEventListener('mouseover', handleMouseOver)
    control.removeEventListener('mouseout', handleMouseOut)
}
```

trevorlang (changes requested):

static/src/js/containers/MapContainer/MapEvents.jsx
```diff
@@ -23,6 +62,19 @@ const MapEvents = (props) => {
       attributionElement.innerHTML = '* © <a href="http://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors'
       layersControl.appendChild(attributionElement)
     }
+
+    // Return cleanup function to remove event listeners when component unmounts
+    return () => {
```
trevorlang, line 67: This is a small thing, and since this is going to be short-lived Im not sure its even worth fixing, but It would be better to do your looping and finding in the scope above and saving it to a variable that could be used in the return function, rather than duplicating the logic down here.

trevorlang (approved):
One last small comment. Not sure if its worth addressing now but I wanted to call it out.

PR #1868, opened 2025-02-24
EDSC-3560: Adding collection sort key parameter into URL

william-valencia:
# Overview

### What is the feature?

When the user clicks on a non default collection sort key (or if preferences are set, not preferred sort key) , they key shows up as a parameter in the URL

### What is the Solution?

Updated code to add collection sort key parameter to the URL and have EDSC use collection sort key parameter in the URL if present

Collection Sorting on EDSC and the URL's shown on the browser. Also the user preferred sort key will be used if set.

# Testing

### Reproduction steps

Open a browser and navigate to EDSC. While logged out, select a different Collection Sort Key ie Relevance, Start Date, or End Date. Ensure that a corresponding query parameter gets added to the URL:
ie:
csk=-score (Relevance)
csk=start_date (Start Date)
csk=-ongoing (Ongoing)

Also ensure refreshing the page with the specified query parameter keeps the sort order as it should be.

Then login. Check your user preferences and make note of your Collection Sort preference. Go the the EDSC homepage and make sure the collections are sorted based on your preference. 

Then select a new preference and repeat. Make sure the selected preference is how the collections get sorted initially on the home page and that your selected preference is highlighted as the active sorting option. Make sure your preference does not show up as a query parameter.

Selecting a new preference from the drop down, however will add the proper query parameter to the URL and the collections get sorted with the new parameter.

Refreshing the page with a proper csk parameter set will also yield the expected results. (proper collection sorting and correctly highlighted text)

### Attachments
Screenshot 2025-03-04 at 6 40 46 PM

# Checklist


static/src/js/selectors/preferences.js
```diff
@@ -46,3 +47,21 @@ export const getGranuleSortPreference = createSelector(
...
+ * @param {Object} state Current state of Redux
+ */
+export const getNondefaultCollectionSortKey = (state) => {
+  const { collectionSearchResultsSortKey } = getApplicationConfig()
+
+  const { query } = state
+  const { collection } = query
+  const { sortKey } = collection
+
+  return sortKey[0] !== collectionSearchResultsSortKey ? sortKey[0] : null
```
eudoroolivares2016, line 66:
```suggestion
const { collectionSearchResultsSortKey: defaultSortKey } = getApplicationConfig()

  const { query } = state
  const { collection } = query
  const { sortKey } = collection
  const [collectionSortKey] = sortKey

  return collectionSortKey !== defaultSortKey ? collectionSortKey : null
```
Just my preference I don't think we are perfectly consistent on using the es6 arr destructure

eudoroolivares2016 (changes requested):

static/src/js/components/SearchPanels/SearchPanels.jsx
```diff
@@ -284,9 +292,12 @@ class SearchPanels extends PureComponent {
 
     const setCollectionSort = (value) => {
       const sortKey = [value]
+      const paramCollectionSortKey = value
+      console.log(`setCollectionSort: ${sortKey}`)
```
eudoroolivares2016, line 296: rm

static/src/js/components/SearchPanels/__tests__/SearchPanels.test.jsx
```diff
@@ -154,9 +154,9 @@ describe('SearchPanels component', () => {
 
       expect(collectionResultsPanelProps.activeSort).toBe('')
       expect(collectionResultsPanelProps.sortsArray[0].label).toBe('Relevance')
-      expect(collectionResultsPanelProps.sortsArray[0].isActive).toBe(true)
+      expect(collectionResultsPanelProps.sortsArray[0].isActive).toBe(false)
```
eudoroolivares2016, line 157: The reason this is failing on the remote is because in the `static.config.json` we set the `collectionSearchResultsSortKey` as `-score` we should either keep this using Relevance or otherwise just mock the function call. We should add a few new tests for the new logic we have introduced (the logic for determining the active sort key) or otherwise move that logic to a util and test it there then just ensure that the final response here is correct

macrouch: We should be mocking any config file values used in a test so that this kind of failure doesn't happen in the future.

eudoroolivares2016: Actually I was wrong so this is actually being passed in the setup method. We had ran into that problem I described about and I correctly thought it was the same issue here. Setup is passing -score so that does not need to be mocked

static/src/js/components/SearchPanels/SearchPanels.jsx
william-valencia: removed

static/src/js/actions/urlQuery.js
```diff
@@ -48,8 +48,11 @@ export const updateStore = ({
   const { router } = state
   const { location } = router
   const { pathname } = location
+  const { collection } = query
+  const { sortOrder } = collection
 
-  const collectionSortPreference = getCollectionSortPreference(state)
+  // Use sortOrder if defined
+  const collectionSortPreference = sortOrder || getCollectionSortPreference(state)
```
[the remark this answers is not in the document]
william-valencia: user preferences are now being used.

static/src/js/selectors/preferences.js
william-valencia: Updated

static/src/js/components/SearchPanels/SearchPanels.jsx
```diff
@@ -180,10 +180,18 @@ class SearchPanels extends PureComponent {
 
     const {
       pageNum: collectionsPageNum = 1,
-      sortKey: collectionsSortKey = collectionSortKeys.scoreDescending
+      paramCollectionSortKey: urlCollectionsSortKey
     } = collectionQuery
 
-    const [activeCollectionsSortKey = collectionSortKeys.scoreDescending] = collectionsSortKey
+    const { collectionSort: userPrefCollectionSortKey } = preferences
```
eudoroolivares2016, line 186: Since the test for this component is `static/src/js/components/SearchPanels/__tests__/SearchPanels.test.jsx` already just passing in state `collectionQuery` we having things passing since we are maintaining the default behavior but, we should add a test that is passing this new one into the component

eudoroolivares2016 (approved):

eudoroolivares2016:
Tested feature locally including with a test with an empty user preferences

static/src/js/util/collections.js
```diff
@@ -4,6 +4,23 @@ import { getApplicationConfig } from '../../../../sharedUtils/config'
 import { tagName } from '../../../../sharedUtils/tags'
 import { autocompleteFacetsMap } from './autocompleteFacetsMap'
 import { withAdvancedSearch } from './withAdvancedSearch'
+import { getCollectionSortPreference } from '../selectors/preferences'
+
+/**
+ * Translate the sortKey to the actual default sort key
+ * if it is set to 'default'. If sortKey is undefined then
```
eudoroolivares2016, line 11:
```suggestion
 * if sortkey is set to 'default' or is undefined then
```

static/src/js/selectors/preferences.js
```diff
@@ -46,3 +47,24 @@ export const getGranuleSortPreference = createSelector(
...
+ * a different collection sorting key.
+ * @param {Object} state Current state of Redux
+ */
+export const getCollectionSortKeyParameter = (state) => {
+  const { query } = state
+  const { collection } = query
+  const { paramCollectionSortKey } = collection
+  const userPrefSortKey = translateDefaultCollectionSortKey(getCollectionSortPreference(state))
+
+  if (paramCollectionSortKey === userPrefSortKey) {
```
eudoroolivares2016, line 65:
```suggestion
// Do not show url parameter if preference matches the current query
  if (paramCollectionSortKey === userPrefSortKey) {
```

eudoroolivares2016: We'll update this later when we change it so that it would show up for all user preferences

macrouch (commented):
Functionality looks good, but the code is hard to follow. 

You put a function in the selectors directory `getCollectionSortKeyParameter`. You are using it like a selector (from `UrlQueryContainer.jsx`), but it isn't a selector. (Selectors use the `createSelector` function from `reselect`). 

Then you are using a selector (`getCollectionSortPreference`) from outside of redux in the `collections.js` util file. You are also using that selector incorrectly from `getCollectionSortKeyParameter` (see the `createSelector` syntax.

So changes:

1. `getCollectionSortKeyParameter` should use `createSelector`, it should call `getCollectionSortPreference` in the same way that `getCollectionSortPreference` calls `getPreferences`
2. `getCollectionSortPreference` should not be called from `collections.js`. Instead, pull `preferences` out of `state` on line 40, and pull `collectionSort` out of `preferences` and pass that to `translateDefaultCollectionSortKey`. This is the same way that `SearchPanels.jsx` uses it

eudoroolivares2016 (approved):

static/src/js/components/SearchPanels/SearchPanels.jsx
```diff
@@ -180,10 +181,16 @@ class SearchPanels extends PureComponent {
 
     const {
       pageNum: collectionsPageNum = 1,
-      sortKey: collectionsSortKey = collectionSortKeys.scoreDescending
+      paramCollectionSortKey: urlCollectionsSortKey
```
daniel-zamora, line 184: is there a reason we are aliasing here to urlCollectionsSortKey?

william-valencia: Yeah, look on line 294. We are updating the value of paramCollectionSortKey.

daniel-zamora (approved):

PR #1870, opened 2025-02-26
EDSC-4421: Fixing styling discrepancies from bootstrap upgrade 

dpesall:
# Overview

### What is the feature?
Fixing some minor styling errors that resulted from upgrading our version of bootstrap.

### What is the Solution?

- Corrected alignment of subscription action buttons
- Added ellipses to facets which are cut off due to long names
- Corrected font size for Additional Filters text to prevent wrapping
- Fixed `x` button on Shape File modal

- EDSC Home Page
- Shape File Modal
- Subscriptions Page

# Testing

### Reproduction steps

1. Navigate to EDSC home page
2. Verify that the visuals changed in this ticket have been corrected
3. Create a subscription
4. Navigate to Subscriptions page
5. Verify that the action buttons are aligned properly

### Attachments

Screenshot 2025-02-26 at 2 14 56 PM
Screenshot 2025-02-26 at 2 15 15 PM
Screenshot 2025-02-26 at 2 16 23 PM

# Checklist


static/src/js/components/Sidebar/SidebarFiltersList.scss
```diff
@@ -2,4 +2,11 @@
...
   list-style: none;
-}
+
+  &__small-checkbox-text {
+    .form-check-label {
+      font-size: 0.825rem;
+      line-height: 1.2;
+    }
+  }
+}
```
eudoroolivares2016, line 12: newline

static/src/js/components/SubscriptionsList/SubscriptionsListTable.scss
```diff
@@ -13,6 +13,11 @@
 
   &__actions {
     text-align: center;
+    div {
```
eudoroolivares2016, line 16: spaces between these

static/src/js/components/ShapefileUploadModal/ShapefileUploadModal.scss
```diff
@@ -26,6 +26,10 @@
     margin-right: calc(bootstrap.$spacer / 2);
   }
 
+  .btn-close {
```
macrouch, line 29: I don't understand this change. this should be core functionality of react-bootstrap, what are we doing to break it?

dpesall: From what I could find, the bootstrap upgrade likely changed how the close button is positioned within the stacking context which caused this issue. Why it's only on this modal, I'm not sure.

macrouch: When we upgraded bootstrap, in the EDSCModal we changed from rendering our own close button to using the react-bootstrap `closeButton` prop. It looks like we already used that prop on ShapefileUploadModal before the upgrade.

react-bootstrap docs specify that `onHide` is called when the `closeButton` is clicked, so we are doing something that breaks that base functionality. The fix for that shouldn't be to override the library's styles, we should find what we are doing to break it and fix that.

macrouch (commented):
I was able to get the facet length issue fixed with only including `overflow: hidden;` to FacetsItem.scss on `&__title-container`. I didn't need to change any min/max-width values

I was able to fix the sidebar filter wrapping text with only adding `font-size: 0.825rem;` to SidebarFiltersItem.scss on `&__body`

Using `&__header` is definitely better than using `.btn-close` in ShapefileUploadModal.scss. The probem/difference with the Shapefile modal and EDSCModal seems to be the `&:before` css where we are putting a dashed border inset on the modal. 

I was able to fix the Subscription list icon alignment by adding the class `align-middle` to the `EDSCIcon` call on line 116 of SubscriptionsListTable.jsx

trevorlang:
> I was able to get the facet length issue fixed with only including `overflow: hidden;` to FacetsItem.scss on `&__title-container`. I didn't need to change any min/max-width values
> 
> I was able to fix the sidebar filter wrapping text with only adding `font-size: 0.825rem;` to SidebarFiltersItem.scss on `&__body`
> 
> Using `&__header` is definitely better than using `.btn-close` in ShapefileUploadModal.scss. The probem/difference with the Shapefile modal and EDSCModal seems to be the `&:before` css where we are putting a dashed border inset on the modal.
> 
> I was able to fix the Subscription list icon alignment by adding the class `align-middle` to the `EDSCIcon` call on line 116 of SubscriptionsListTable.jsx

I agree with matthew. I noticed that this is using the bootstrap `Modal` directly and not using our `EDSCModalContainer`, maybe swapping it over would fix the issue...

static/src/js/components/ShapefileUploadModal/ShapefileUploadModal.scss
trevorlang, line 1: @dpesall / @eudoroolivares2016, I played around with this and a better way to fix it would be to add the following in the `&:before` on line 37

```
&:before {
    ...
    user-select: none;

    .is-dragging & {
        ....
    }
}
```

This will prevent that pseudo element from intercepting the click events 😄

trevorlang (approved):

eudoroolivares2016 (approved):

PR #1872, opened 2025-03-05
EDSC-4433: Update NASA official Make this a configurable value

eudoroolivares2016:
# Overview

### What is the feature?

Updates the NASA official to the current official @littleidiot40  makes it a configurable value on Bamboo

### What is the Solution?

Updates the footer on the page as well as the `edsc` portal since it is being set for the attribution text
Removing the /paritals folders they are no longer being used since webpack; makes some fixes to the fallback page on EDSC

Footer component information, edsc portal, fallback page

# Testing

### Reproduction steps

- **Environment for testing:*ANY*
- **Collection to test with:*NA*

1. Ensure that the footer has been updated on all the pages
2. Ensure footer is updated on different portals
3. Disable javascript on the page ensure the official is correct and that the colors on the links are now visible 

### Attachments
Fix the colors on the fallback page
image

Can override the value in SIT ie. I set it in bamboo to this string
image

# Checklist

- [NA] I have added automated tests that prove my fix is effective or that my feature works
- [NA] I have commented my code, particularly in hard-to-understand areas
- [NA] I have made corresponding changes to the documentation


macrouch:
Put this name into the config file and overwrite it during deployments from a bamboo variable

eudoroolivares2016:
@macrouch I can look into how that would work but, we'd have to read the static.config.json from the portal config

static/src/js/actions/__tests__/urlQuery.test.js
```diff
@@ -206,7 +206,7 @@ describe('updateStore', () => {
               }
             },
             footer: {
-              attributionText: 'NASA Official: Stephen Berrick',
```
eudoroolivares2016, line 209: We don't have to update these I was just doing it for consistency I can go and make these test values or just leave them if people prefer

static/src/js/util/__tests__/portals.test.js
```diff
@@ -4,6 +4,91 @@ import * as getApplicationConfig from '../../../../../sharedUtils/config'
 // eslint-disable-next-line import/no-unresolved
 import availablePortals from '../../../../../portals/availablePortals.json'
 
+jest.mock('../../../../../portals/availablePortals.json', () => ({
```
eudoroolivares2016, line 7: Was using the real portals before so changes would break the test now its using a mock

bin/deploy-bamboo.sh
```diff
@@ -6,6 +6,15 @@ set -eux
 # Deployment configuration/variables
 ####################################
 
+# Read in edsc portal configuration
+edscPortal=$(cat portals/edsc/config.json)
+
+# Update key for edsc portal
+edscPortal=$(jq --arg newValue "$bamboo_NASA_ATTRIBUTION_TEXT" '.footer.attributionText = $newValue' <<< "$edscPortal")
```
eudoroolivares2016, line 13: This has to be wrapped this way because the attribution has a space in it

portals/edsc/config.json
```diff
@@ -9,7 +9,7 @@
     }
   },
   "footer": {
-    "attributionText": "NASA Official: Stephen Berrick",
+    "attributionText": "NASA Official: Doug Newman",
```
macrouch, line 12: Don't put a name here. The value will be set during deployments and we don't want another PR to change this name again later

static/src/js/util/__tests__/portals.test.js
macrouch: I don't understand this comment, but you've added this huge wall of text in multiple files now

static/src/js/actions/__tests__/urlQuery.test.js
```diff
@@ -10,6 +10,121 @@ import * as urlQuery from '../urlQuery'
 
 const mockStore = configureMockStore([thunk])
 
+jest.mock('../../../../../portals/availablePortals.json', () => ({
+  airmoss: {
```
macrouch, line 14: If you need to mock values because we are testing for them, you should use fake values, not the real portals. Using real values can lead to confusion later if the portals change and the values are no longer synced. I realize this is not how we did it with the availablePortalsMock.json, but we should improve and do it better than we used to

eudoroolivares2016: I agree! Good callout I did need to use `virtual mocks` in one place since the png file that jest tries to get does not actually exist on the disk

portals/edsc/config.json
```diff
@@ -9,7 +9,7 @@
     }
   },
   "footer": {
-    "attributionText": "NASA Official: Stephen Berrick",
+    "attributionText": "",
```
macrouch, line 12: I think we should include a value here. Not a real name we'll have to change in the future, but some text. Without something that block of text will never show up in development, which could lead to headaches in the future because we didn't handle something until we noticed it in a production environment.

static/src/js/actions/__tests__/urlQuery.test.js
```diff
@@ -10,6 +10,121 @@ import * as urlQuery from '../urlQuery'
 
 const mockStore = configureMockStore([thunk])
 
+jest.mock('../../../../../portals/availablePortals.json', () => ({
```
macrouch, line 13: Is this same mock defined in 3 different places?

portals/edsc/config.json
eudoroolivares2016: I think we should just default to Doug the tests now don't have those so it shouldn't be confusing. It will just get overwritten with each deployment

static/src/js/actions/__tests__/urlQuery.test.js
eudoroolivares2016: Moved this to update the jest config

macrouch (approved):

dpesall (approved):

PR #1881, opened 2025-03-20
EDSC-3620: Commafying the facet item totals with explicit check for undefined count

william-valencia:
# Overview

### What is the feature?

Commafying the facet item totals

### What is the Solution?

Calling commafy on the total only if facet.count is defined.

Facets and the Facet modals

# Testing

### Reproduction steps

Navigate to EDSC and view the Facet lists. Make sure any facets over 999 have commas.
Make sure any facets with 0 such as NISAR are the same color as other totals.

### Attachments

Screenshot 2025-03-19 at 9 52 08 PM

# Checklist

### DIFF
--- a/static/src/js/components/Facets/FacetsItem.jsx
+++ b/static/src/js/components/Facets/FacetsItem.jsx
@@ -145,7 +145,7 @@ class FacetsItem extends Component {
               )
             }
           </div>
-          { (!applied || !children) && (facet.count !== undefined) && <span className="facets-item__total">{commafy(facet.count)}</span> }
+          { (!applied || !children) && (facet.count != null) && <span className="facets-item__total">{commafy(facet.count)}</span> }
         </label>
         { children && <ul className="facets-list">{children}</ul> }
       </li>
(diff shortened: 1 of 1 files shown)


static/src/js/components/Facets/FacetsItem.jsx
```diff
@@ -145,7 +145,7 @@ class FacetsItem extends Component {
               )
             }
           </div>
-          { (!applied || !children) && facet.count && <span className="facets-item__total">{commafy(facet.count)}</span> }
+          { (!applied || !children) && (facet.count !== undefined) && <span className="facets-item__total">{commafy(facet.count)}</span> }
```
eudoroolivares2016, line 148:
```suggestion
          { (!applied || !children) && (facet.count facet.count != null) && <span className="facets-item__total">{commafy(facet.count)}</span> }
```
The current change does fix the edge case which should not happen in the first place but, is a state we can currently enter due to a bug on the API. This suggested solution above would cover if the `facet.count` is `undefined` or `null` by loose equality

eudoroolivares2016 (approved):

dpesall (approved):

PR #1887, opened 2025-03-29
EDSC-4418: User preferences are accounted for on the open layers map

eudoroolivares2016:
# Overview

### What is the feature?

Fixes a bug in production today, where set preference values **always** override the map-view even when a url containing `mapView` configurations is passed. Renames `blueMarble` to `worldImagery` to give attribution to the new map we are using

### What is the Solution?

We should only set the Map state in redux **IF** when there is no map view configurations in the url i.e. the user didn't select any changes to the map. Its only when the user is logged in and they are using the default map settings and has preferences that we should return the map with their preferences set.

Updates bluemarble in the code, preferences schema, and does a database migration to update users with that value set to `worldImagery`

Database `users` table, preferences, ol map

# Testing

### Reproduction steps

- **Environment for testing:*Any*
- **Collection to test with:*NA*

1. Ensure that the database migration works by running `npm run invoke-local migrateDatabase` locally and ensure that users with `blueMarble` set in their `site_preferences` column on the `users` table have the baseLayer field updated to `worldImagery`. 

2. In the migrateDatabase lambda handler change the code to use `down` to test that the database migration reversion also works. Ensure that the values for the users revert back to `blueMarble`

3. Ensure that when you are logged out and go to `http://localhost:8080/search` it takes you to the map with the base settings (lat long 0,0 etc)

4. Ensure that when you move the map, zoom, change projections etc while logged out the url is updating

5. Ensure that when you refresh the page or paste it into a new browser the urls are read and the same map settings carry forward from the url

6. Now log into EDSC and ensure you have some preferences set. Ensure that if you navigate to `http://localhost:8080/search` those preferences are used. Ensure that the url does **not** contain those values it should just be `/search`

7. Make map changes move around, zoom, etc and then refresh or share that url to a new browser. Ensure that the url is getting updated. Then on a refresh or new tab ensure that those settings are being used in the map

### Attachments

Top level /search going to my preferences area and zoom level
image# Checklist


migrations/1743133132745_change-blue-marble-to-world-imagery.js
```diff
@@ -0,0 +1,31 @@
+exports.shorthands = undefined
+
+exports.up = async (pgm) => {
+  // Update the value in the JSONB column
```
eudoroolivares2016, line 4:
```suggestion
  // Update the value in the jsonb column
```

schemas/sitePreferencesSchema.json
```diff
@@ -151,14 +151,14 @@
         "baseLayer": {
           "type": "string",
           "title": "Base Layer",
-          "default": "blueMarble",
+          "default": "worldImagery",
```
eudoroolivares2016, line 154: Ensure that `blueMarble` is temporarily valid so the user will never see an error message on the preferences page

static/src/js/actions/__tests__/mocks.js
```diff
@@ -1 +1 @@
...
+export const testJwtToken = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6MSwidXNlcm5hbWUiOiJ0ZXN0dXNlciIsInByZWZlcmVuY2VzIjp7Im1hcFZpZXciOnsiem9vbSI6NCwibGF0aXR1ZGUiOjM5LCJiYXNlTGF5ZXIiOiJ3b3JsZEltYWdlcnkiLCJsb25naXR1ZGUiOi05NSwicHJvamVjdGlvbiI6ImVwc2c0MzI2Iiwib3ZlcmxheUxheWVycyI6WyJyZWZlcmVuY2VGZWF0dXJlcyIsInJlZmVyZW5jZUxhYmVscyJdfSwicGFuZWxTdGF0ZSI6ImRlZmF1bHQiLCJncmFudWxlU29ydCI6ImRlZmF1bHQiLCJjb2xsZWN0aW9uU29ydCI6ImRlZmF1bHQiLCJncmFudWxlTGlzdFZpZXciOiJkZWZhdWx0IiwiY29sbGVjdGlvbkxpc3RWaWV3IjoiZGVmYXVsdCJ9LCJ1cnNQcm9maWxlIjp7ImZpcnN0X25hbWUiOiJ0ZXN0In0sImVhcnRoZGF0YUVudmlyb25tZW50IjoicHJvZCIsImlhdCI6MTY3MzY0NjY1NH0.Z_qDm4aWpSRMxN4tpcyLJ-ZvZP-mLzFlzINMDR03VLA'
```
eudoroolivares2016, line 1: The reason this gets updated is because if you go on https://jwt.io/ this has blueMarble in it so it needs to be updated

static/src/js/components/Preferences/PreferencesRadioField.jsx
```diff
@@ -38,10 +38,27 @@ class PreferencesRadioField extends Component {
...
-      enumNames,
-      description
+      description,
+      title,
+      enumNames
     } = schema
 
+    let parsedFormData = formData
+    // TODO remove this in a subsequent release
+    if (title === 'Base Layer') {
```
eudoroolivares2016, line 48: When we do the database migration on deployment, if a user is logged in already they have an active jwt token which will contain the old value `blueMarble`. This will be fixed when the user logs out and back in or when the cookie expires. In the meantime we don't want to show blueMarble as an option but, we also need it in the schema so that the user does not see an error message
image

static/src/js/components/Preferences/PreferencesForm.jsx
```diff
@@ -41,10 +42,13 @@ const PreferencesForm = (props) => {
     radio: PreferencesRadioField
   }
 
-  const validate = (formDataObject, errors) => {
-    // Projections that aren't geographic have a zoom limit of 4
-    if (formDataObject.mapView.projection !== 'epsg4326' && formDataObject.mapView.zoom > 4) {
-      errors.mapView.zoom.addError('should be less than or equal to 4')
+  //  Notes: - The validate() function must always return the errors object received as second argument
+  //  The validate() function is called after the JSON schema validation.
+  const customValidate = (formDataObject, errors) => {
```
eudoroolivares2016, line 47: This is currently not working in prod today because the prop at some point was renamed by the package we are using. This fixes it and give the new values what the maximum is supposed to be for open layers

static/src/js/components/Preferences/PreferencesRadioField.jsx
```diff
@@ -38,10 +38,27 @@ class PreferencesRadioField extends Component {
...
       enum: values,
-      enumNames,
-      description
+      description,
+      title,
+      enumNames
     } = schema
 
+    let parsedFormData = formData
+    // TODO remove this in a subsequent release
```
macrouch, line 47: Write the ticket to do this work and leave the ticket number in this TODO comment

tests/e2e/projects/harmony_with_mbr/harmony_with_mbr.spec.js
```diff
@@ -87,6 +84,7 @@ test.describe('Harmony with MBR', () => {
 
           // Expect the URL to be updated with `ess=t`
           await expect(page).toHaveURL(/pg\[1\]\[ess\]=t/)
+          await page.waitForURL(/zoom=3/, { timeout: 3000 })
```
macrouch, line 87: Is this a different zoom level than before? If it is we need new screenshots for these tests

eudoroolivares2016: I had to bring down the pixel diff to get the `--update-snapshots` to work those should now we updated

static/src/js/components/Preferences/PreferencesRadioField.jsx
eudoroolivares2016: We should do this once only in the `static/src/js/actions/preferences.js` at this level it will push down to other components including this one

macrouch (approved):

static/src/js/components/Preferences/PreferencesForm.jsx
macrouch: We should have a test for this method

dpesall (approved):

PR #1899, opened 2025-04-22
EDSC-4771: Remove node-forge and replace with a smaller library

eudoroolivares2016:
# Overview

### What is the feature?

Fixes a few bugs for saving echo-forms, shapefiles, and processing multi-shapefiles

Removes `node-forge` and replaces with `md5` package. Replaced the echo-forms with an md5 encoding 

### What is the Solution?

For the hash of the echo-form/saved access configuration we are replacing that with a base64 encoding this should be sufficient for our purposes. We are removing node forge and adding in md5 which uses the same cryptographic hash but, the package is significantly lighter weight.

In production today, echo-forms aren't being saved correctly because the formDigest was not being passed correctly. Similarly when we are saving shapefiles, the ogre response that converts them to geoJson had a hash in the name of the file and we were not parsing partial shapefiles correctly as the name is nested in a `properties` object

We are clearing out the saved access configurations table of data because the values are not useful given that they were not being saved correctly. Additionally we are now passing and saving the orderoptions revision-id in the database so that we can if needed, get the echo form data base from CMR. Because that was not in there we cannot fix the existing data in the saved access configuration

Saved access configurations, shapefile saving, processing partial shapefiles

# Testing

### Reproduction steps

- **Environment for testing:*Prod*
- **Collection to test with:*See steps below*

1. Submit this harmony order including the shapefile `tests/fixtures/shapefiles/multiple_shapes.geojson` to http://localhost:8080/projects?p=C2596864127-NSIDC_CPRD!C2596864127-NSIDC_CPRD&pg[1][a]=3444063768!3445031488!3445126880!3445086504!NSIDC_CPRD&pg[1][v]=t&pg[1][gsk]=-start_date&pg[1][m]=harmony0&pg[1][cd]=f&pg[1][ets]=t&pg[1][ess]=t&line[0]=31%2C-15%2C36%2C-17%2C41%2C-15&ff=Customizable&sf=4517239960&sfs[0]=2&csk=-usage_score&tl=1642413449.847!4!!&lat=-8.415327435948422&long=46.466170630743946&zoom=4.252758534379918 and choose one of the shapes on the map to subset from
ensure that it saves the shapefile **and** a child for the subset shapefile in the same table

2. Ensure that the database migration works note it is only a forward step we cannot restore the saved access configration data `npm run invoke-local migrateDatabase`

3. Use `C179031504-LARC` for echo-order and `C1396790070-NSIDC_ECS` ensure that when you download data from it after entering in form data (For this specific collection use the Ftp push option) ensure that going back to this download from your saved downloads and then back to the project, the entries in the form show back up and if you make changes they are updated when you return. Ensure that the database entry in your saved access configurations table includes the revision-id for the order option associated to this collection (it will just be the most recent revision-id)

4. Using `C1996881146-POCLOUD` and subsetting with a geojson containing multiple shapes I am able to successfully get back the job from Harmony the same way it works in production

5. Ensure the form is saved correctly for a opendap collection (THIS IS IN UAT http://localhost:8080/projects?p=!C1224234390-E2E_18_4&pg[1][v]=t&pg[1][m]=opendap&pg[1][cd]=f&pg[1][of]=nc4&q=C1224234390-E2E_18_4&tl=1388190603.043!5!!) 

6. Ensure that forms are saved correctly for a harmony collection and opendap i.e. going back to Harmony after works as expected

### Attachments
image

# Checklist


serverless/src/getSavedAccessConfigs/handler.js
```diff
@@ -48,6 +48,7 @@ const getSavedAccessConfigs = async (event, context) => {
         collection_id: collectionId,
         access_method: accessMethod
       } = config
+      console.log(`Found savedAccessConfigs for ${accessMethod.type} collectionIds: ${collectionIds}`)
```
macrouch, line 51: Add a test for this to make sure we don't remove or change it by accident. use `spyOn(console, 'log')` and do `toHaveBeenCalledTimes()` and `toHaveBeenCalledWith()`

macrouch (approved):

dpesall (approved):

trevorlang (approved):

PR #1900, opened 2025-04-22
EDSC-4465: Remove "retrievalHistory" redux code

dpesall:
# Overview

### What is the feature?

- Removing redux code for `retrievalHistory` logic, as all the logic using the stored variables is within the same component.

### What is the Solution?

- Refactored DownloadHistory.jsx, DownloadHistoryContainer.jsx
-  Removed redux code
- Rewrote enzyme tests into RTL

- Download History page

# Testing

### Reproduction steps

1. Select and download granule to download
2. Navigate to the Download History page
3. Verify the download page works as expected and populates the new download
4. Verify the `x` button removes the entry
5. Refresh the page and verify that it shows the correct entries

# Checklist


static/src/js/components/DownloadHistory/__tests__/DownloadHistory.test.jsx
macrouch, line 1: When converting tests we should take the opportunity to bring in patterns we've learned since writing the original test.

We like to use describe blocks to describe the input of the test, and the test block to describe the output.
So the first test is 
```
    test('renders a spinner when retrievals are loading', () => {
```
That should change to 
```
    describe('when retrievals are loading' , () => {
      test('renders a spinner', () => {
```

For the setup function, use the `const setup = (props) => {` syntax.

Don't use test ids, when finding an element our tests should behave like users. The user is looking for a link, so use `screen.getByRole('link', { name: '1 collection' })`, which also tests the text content, so you don't need that extra assertion.

static/src/js/components/DownloadHistory/__tests__/DownloadHistory.test.jsx
```diff
@@ -125,49 +142,28 @@ describe('DownloadHistory component', () => {
...
-      expect(helmet.childAt(1).props().name).toEqual('title')
-      expect(helmet.childAt(1).props().content).toEqual('Download Status & History')
-      expect(helmet.childAt(2).props().name).toEqual('robots')
-      expect(helmet.childAt(2).props().content).toEqual('noindex, nofollow')
-      expect(helmet.childAt(3).props().rel).toEqual('canonical')
-      expect(helmet.childAt(3).props().href).toEqual('https://search.earthdata.nasa.gov/downloads')
+      expect(screen.getByRole('heading', { level: 2 })).toHaveTextContent('Download Status & History')
     })
 
     test('renders links correctly when portals were used to place an order', () => {
```
macrouch, line 165: None of the logic described in this test is part of DownloadHistory, and should be tested within PortalLinkContainer, this test should be removed.

static/src/js/containers/DownloadHistoryContainer/__tests__/DownloadHistoryContainer.test.js
```diff
@@ -1,89 +1,87 @@
...
 
-    mapDispatchToProps(dispatch).onFetchRetrievalHistory('prod')
+describe('DownloadHistoryContainer component', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
 
-    expect(spy).toBeCalledTimes(1)
-    expect(spy).toBeCalledWith('prod')
+    // Mock the RetrievalRequest methods
+    RetrievalRequest.mockImplementation(() => ({
```
macrouch, line 31: Use `nock` to mock network requests, you don't need to mock the request classes

static/src/js/containers/DownloadHistoryContainer/__tests__/DownloadHistoryContainer.test.js
```diff
@@ -1,89 +1,87 @@
...
-    const store = {
-      earthdataEnvironment: 'prod',
-      retrievalHistory: {
-        history: {},
-        isLoading: false,
-        isLoaded: false
-      }
-    }
+    await waitFor(() => {
+      expect(screen.getByTestId('download-history')).toBeInTheDocument()
```
macrouch, line 59: Don't use test ids

static/src/js/containers/DownloadHistoryContainer/__tests__/DownloadHistoryContainer.test.js
```diff
@@ -1,89 +1,87 @@
...
+    })
 
-    const expectedState = {
-      earthdataEnvironment: 'prod',
-      retrievalHistory: {},
-      retrievalHistoryLoading: false,
-      retrievalHistoryLoaded: false
-    }
+    const { calls } = DownloadHistory.mock
```
macrouch, line 62:
```
    expect(DownloadHistory).toHaveBeenCalledTimes(#)
    expect(DownloadHistory).toHaveBeenCalledWith(...)
// or this one
    expect(DownloadHistory).toHaveBeenLastCalledWith(...)
```

static/src/js/containers/DownloadHistoryContainer/DownloadHistoryContainer.jsx
```diff
@@ -1,71 +1,87 @@
...
-    onFetchRetrievalHistory()
+    try {
+      const requestObject = new RetrievalRequest(authToken, earthdataEnvironment)
+      const response = await requestObject.all()
+      const { data } = response
+
+      setRetrievalHistory(data)
+      setRetrievalHistoryLoaded(true)
+    } catch (error) {
+      console.error('Error fetching retrieval history:', error)
```
macrouch, line 31: console.error in a user's browser does not help us find and fix problems, errors need to be dispatched to `handleError` like they previously were. That method handles showing the user a problem and reporting the error to a lambda where we can see the log.

static/src/js/actions/retrieval.js
```diff
@@ -188,7 +187,6 @@ export const deleteRetrieval = (id) => (dispatch, getState) => {
     const requestObject = new RetrievalRequest(authToken, earthdataEnvironment)
```
macrouch, line 187: This whole deleteRetrieval method should be removed

macrouch (changes requested):

static/src/js/components/DownloadHistory/DownloadHistory.jsx
```diff
@@ -16,144 +16,131 @@ import PortalLinkContainer from '../../containers/PortalLinkContainer/PortalLink
...
-                    })
-                  }
-                </tbody>
-              </Table>
-            ) : (
-              <p>No download history to display.</p>
-            )
+                            }
+                          >
+                            {retrievalDescription(collections)}
```
trevorlang, line 116: Im tempted to say we should memoize the result of this against the retrievalHistory so we dont have to recalculate it on every render.

static/src/js/components/DownloadHistory/DownloadHistory.jsx
```diff
@@ -16,144 +16,131 @@ import PortalLinkContainer from '../../containers/PortalLinkContainer/PortalLink
...
+                          </PortalLinkContainer>
+                        </td>
+                        <td className="download-history-table__ago">
+                          <TimeAgo date={createdAt} />
+                        </td>
+                        <td className="download-history-table__actions">
+                          <Button
+                            className="download-history__button download-history__button--remove"
+                            onClick={() => handleRemove(id)}
```
trevorlang, line 125: We do this alot but it actually creates a new function on every render. If this was moved to a useCallback like this, it would avoid that:

```
const handleRemoveClick = useCallback((id) => () => handleRemove(id), [handleRemove])

...

<Button
  onClick={handleRemoveClick(id)}
  ...
/>
```

static/src/js/components/DownloadHistory/DownloadHistory.jsx
```diff
@@ -16,144 +16,131 @@ import PortalLinkContainer from '../../containers/PortalLinkContainer/PortalLink
...
-                            <PortalLinkContainer
-                              portalId={portalId}
-                              to={
-                                {
-                                  pathname: `/downloads/${id}`,
-                                  search: stringify({ ee: earthdataEnvironment === deployedEnvironment() ? '' : earthdataEnvironment })
-                                }
+      {
+        retrievalHistoryLoaded && (
+          retrievalHistory.length > 0 ? (
```
trevorlang, line 83: I think a turnary makes this hard to follow. Something more explicit like this might work better:

```
{
retrievalHistoryLoaded && retrievalHistory.length > 0 && (
...
)
retrievalHistoryLoaded && retrievalHistory.length === 0 && (
...
)
}
```

static/src/js/components/DownloadHistory/DownloadHistory.jsx
trevorlang, line 1: Not a change you caused but since I noticed. the propType for `retrievalHistory` should probably be a little more specific

static/src/js/containers/DownloadHistoryContainer/DownloadHistoryContainer.jsx
```diff
@@ -1,71 +1,101 @@
...
-    return (
-      <DownloadHistory
-        earthdataEnvironment={earthdataEnvironment}
-        retrievalHistory={retrievalHistory}
-        retrievalHistoryLoading={retrievalHistoryLoading}
-        retrievalHistoryLoaded={retrievalHistoryLoaded}
-        onDeleteRetrieval={onDeleteRetrieval}
-      />
-    )
+  const handleDeleteRetrieval = async (retrievalId) => {
```
trevorlang, line 45: This should be a `useCallback` so its not redefined on every render

static/src/js/containers/DownloadHistoryContainer/DownloadHistoryContainer.jsx
```diff
@@ -1,71 +1,101 @@
...
+  const [retrievalHistoryLoaded, setRetrievalHistoryLoaded] = useState(false)
 
-export const mapDispatchToProps = (dispatch) => ({
-  onFetchRetrievalHistory: (earthdataEnvironment) => {
-    dispatch(actions.fetchRetrievalHistory(earthdataEnvironment))
-  },
-  onDeleteRetrieval:
-    (retrievalId) => dispatch(actions.deleteRetrieval(retrievalId))
-})
+  const fetchRetrievalHistory = async () => {
```
trevorlang, line 20: This should be a `useCallback` so its not redefined on every render

static/src/js/containers/DownloadHistoryContainer/DownloadHistoryContainer.jsx
```diff
@@ -1,71 +1,101 @@
...
+      })
+    }
   }
-}
 
-DownloadHistoryContainer.defaultProps = {
-  retrievalHistory: []
+  useEffect(() => {
+    fetchRetrievalHistory()
+  }, [earthdataEnvironment, authToken])
```
trevorlang, line 71: A few questions on this one:
- Are these necessary in the dependency array? 
- Should we prevent the function from firing if they are not defined?

`fetchRetrievalHistory` might need to be added here if its made a `useCallback`

static/src/js/containers/DownloadHistoryContainer/__tests__/DownloadHistoryContainer.test.js
```diff
@@ -1,89 +1,250 @@
...
-  const enzymeWrapper = shallow(<DownloadHistoryContainer {...props} />)
+const mockStore = configureStore([])
 
-  return {
-    enzymeWrapper,
-    props
-  }
-}
+describe('DownloadHistoryContainer component', () => {
+  let store
```
trevorlang, line 38: This `store` should be created in a setup function

static/src/js/containers/DownloadHistoryContainer/__tests__/DownloadHistoryContainer.test.js
```diff
@@ -1,89 +1,250 @@
...
-Enzyme.configure({ adapter: new Adapter() })
+jest.mock('../../../util/addToast', () => ({
+  addToast: jest.fn()
+}))
+
+jest.mock('../../../actions/errors', () => ({
+  handleError: jest.fn(() => ({ type: 'HANDLE_ERROR' }))
+}))
+
+let lastPassedOnDeleteRetrieval = null
```
trevorlang, line 25: Instead of using this variable to track the last retrieval, you can import `DownloadHistory` which will be your mocked component and then use `toHaveBeenCalledTimes` and `toHaveBeenCalledWith` to determine if its called the right number of times and with the correct values.

static/src/js/containers/DownloadHistoryContainer/__tests__/DownloadHistoryContainer.test.js
```diff
@@ -1,89 +1,250 @@
...
+        </Provider>
+      )
+
+      await waitFor(() => {
+        expect(lastPassedOnDeleteRetrieval).toEqual(expect.any(Function))
+      })
+
+      await act(async () => {
+        if (!lastPassedOnDeleteRetrieval) {
+          throw new Error('onDeleteRetrieval callback was not captured by mock')
```
trevorlang, line 203: We shouldnt be throwing errors from tests. We need to make sure that they fail by not meeting `expect`s.

trevorlang (changes requested):

static/src/js/containers/DownloadHistoryContainer/__tests__/DownloadHistoryContainer.test.js
trevorlang: Scratch that, kinda. I needed to take a deeper look a this but I think I see whats going on. I think rather than using this variable to track the most recent call, the more "testing-library" way to do this would be to return a `button` in the mocked component that has an `onClick` which calls `onDeleteRetrieval`. Then in your tests, you can use `userEvent` to click on the button and then call `toHaveBeenCalledWith()` to make sure `onDeleteRetrieval` is called with what we expect.

trevorlang (changes requested):

static/src/js/containers/DownloadHistoryContainer/__tests__/DownloadHistoryContainer.test.js
macrouch: I don't think we need this store at all, if you look at our other container tests we import the named export of the component, not the default export. You are trying to test redux here functionality here and we only want to test our component code.

static/src/js/containers/DownloadHistoryContainer/__tests__/DownloadHistoryContainer.test.js
```diff
@@ -1,89 +1,250 @@
...
+          <MemoryRouter>
+            <DownloadHistoryContainer />
+          </MemoryRouter>
+        </Provider>
+      )
 
-    expect(spy).toBeCalledTimes(1)
-    expect(spy).toBeCalledWith('retrievalId')
+      await waitFor(() => {
+        expect(DownloadHistory).toHaveBeenLastCalledWith(
```
macrouch, line 76: Everywhere you test a mock you have to test how many times it has been called (`toHaveBeenCalledTimes()`) along with it being called with the correct parameters. That tells you how many times the component has been rendered, which can be an indicator of problems

static/src/js/containers/DownloadHistoryContainer/DownloadHistoryContainer.jsx
```diff
@@ -1,71 +1,101 @@
...
-    PropTypes.shape({})
-  ),
-  retrievalHistoryLoading: PropTypes.bool.isRequired,
-  retrievalHistoryLoaded: PropTypes.bool.isRequired,
-  onFetchRetrievalHistory: PropTypes.func.isRequired,
-  onDeleteRetrieval: PropTypes.func.isRequired
+  dispatchHandleError: PropTypes.func.isRequired
 }
 
+const mapStateToProps = (state) => ({
```
macrouch, line 90: Every other container puts `mapStateToProps` and `mapDispatchToProps` before the component, please follow existing patterns in the code

static/src/js/containers/DownloadHistoryContainer/__tests__/DownloadHistoryContainer.test.js
```diff
@@ -1,89 +1,250 @@
...
+            retrievalHistory: initialHistory
+          }),
+          expect.anything()
+        )
+      })
+    })
+  })
+
+  describe('mapStateToProps', () => {
+    test('maps state to props correctly', () => {
```
macrouch, line 233: This is not testing `mapStateToProps` from the component, you have to import it to test it. 

Also missing tests for `mapDispatchToProps`

macrouch (changes requested):

static/src/js/containers/DownloadHistoryContainer/__tests__/DownloadHistoryContainer.test.js
```diff
@@ -1,89 +1,368 @@
...
+  )
+}))
 
-describe('mapDispatchToProps', () => {
-  test('onFetchRetrievalHistory calls actions.fetchRetrievalHistory', () => {
-    const dispatch = jest.fn()
-    const spy = jest.spyOn(actions, 'fetchRetrievalHistory')
+const setup = (props) => render(
+  <MemoryRouter>
+    <DownloadHistory {...props} />
```
macrouch, line 35: This test is supposed to be testing DownloadHistoryContainer, not DownloadHistory

static/src/js/containers/DownloadHistoryContainer/__tests__/DownloadHistoryContainer.test.js
```diff
@@ -1,89 +1,368 @@
...
-    enzymeWrapper,
-    props
-  }
-}
+jest.mock('../../../containers/PortalLinkContainer/PortalLinkContainer', () => ({
+  __esModule: true,
+  default: ({ children, to, portalId }) => (
+    <a
+      href={`/downloads/${to.pathname.split('/').pop()}${to.search}`}
+      data-testid="portal-link"
```
macrouch, line 25: Remove these two data attributes

static/src/js/containers/DownloadHistoryContainer/__tests__/DownloadHistoryContainer.test.js
```diff
@@ -1,89 +1,368 @@
...
+          retrievalHistoryLoaded: true,
+          onDeleteRetrieval: jest.fn()
+        })
+
+        expect(screen.getByRole('table')).toBeInTheDocument()
+        expect(screen.getByRole('link', { name: 'Collection Title and 1 other collection' })).toHaveAttribute('href', '/downloads/8069076')
+      })
+    })
+
+    describe('when portals were used to place an order', () => {
```
macrouch, line 143: I made this comment earlier, but none of the logic described in this test is part of DownloadHistoryContainer, and should be tested within PortalLinkContainer, this test should be removed.

static/src/js/containers/DownloadHistoryContainer/DownloadHistoryContainer.jsx
```diff
@@ -1,69 +1,105 @@
...
+  earthdataEnvironment: state.earthdataEnvironment
 })
 
 export const mapDispatchToProps = (dispatch) => ({
-  onFetchRetrievalHistory: (earthdataEnvironment) => {
-    dispatch(actions.fetchRetrievalHistory(earthdataEnvironment))
-  },
-  onDeleteRetrieval:
-    (retrievalId) => dispatch(actions.deleteRetrieval(retrievalId))
+  dispatchHandleError: (errorConfig) => dispatch(handleError(errorConfig))
```
macrouch, line 21: `dispatch(actions.handleError())` to follow the pattern of how we dispatch actions

static/src/js/containers/DownloadHistoryContainer/__tests__/DownloadHistoryContainer.test.js
```diff
@@ -1,89 +1,368 @@
...
+  beforeEach(() => {
+    jest.spyOn(errorsModule, 'handleError').mockImplementation(
+      (errorConfig) => ({
+        type: 'HANDLE_ERROR',
+        payload: errorConfig
       })
+    )
+  })
+
+  test('maps dispatch to props correctly', () => {
```
macrouch, line 355: This needs to be more descriptive, mapDispatchToProps tests are testing individual functions that have specific results. See any of the other tests as an example

```
  test('onChangeMap calls actions.changeMap', () => {
    const dispatch = jest.fn()
    const spy = jest.spyOn(actions, 'changeMap')

    mapDispatchToProps(dispatch).onChangeMap({ mock: 'data' })

    expect(spy).toHaveBeenCalledTimes(1)
    expect(spy).toHaveBeenCalledWith({ mock: 'data' })
  })
```

macrouch (changes requested):

eudoroolivares2016 (approved):

static/src/js/components/DownloadHistory/DownloadHistory.jsx
```diff
@@ -16,144 +16,140 @@ import PortalLinkContainer from '../../containers/PortalLinkContainer/PortalLink
...
+        <title>Download Status &amp; History</title>
+        <meta name="title" content="Download Status &amp; History" />
+        <meta name="robots" content="noindex, nofollow" />
+        <link rel="canonical" href={`${edscHost}/downloads`} />
+      </Helmet>
+      <h2 className="route-wrapper__page-heading">Download Status & History</h2>
+      {
+        (retrievalHistoryLoading && !retrievalHistoryLoaded) && (
+          <Spinner
+            dataTestId="spinner"
```
macrouch, line 83: Don't use test ids

static/src/js/components/DownloadHistory/__tests__/DownloadHistory.test.jsx
```diff
@@ -1,255 +1,226 @@
...
+jest.mock('../../../../../../sharedUtils/config', () => ({
+  getEnvironmentConfig: jest.fn()
+}))
+
+jest.mock('../../../containers/PortalLinkContainer/PortalLinkContainer', () => ({
+  __esModule: true,
+  default: ({ children, to, portalId }) => (
+    <a
+      href={`/downloads/${to.pathname.split('/').pop()}${to.search}`}
+      data-testid="portal-link"
```
macrouch, line 24: Don't use test ids

static/src/js/components/DownloadHistory/__tests__/DownloadHistory.test.jsx
```diff
@@ -1,255 +1,226 @@
...
-        retrievalHistoryLoaded: true,
-        onDeleteRetrieval: jest.fn()
-      })
-
-      expect(enzymeWrapper.find(Table).length).toBe(1)
-      expect(enzymeWrapper.find('tbody tr').length).toBe(1)
-      expect(enzymeWrapper.find(PortalLinkContainer).prop('to')).toEqual({
-        pathname: '/downloads/8069076',
-        search: ''
+    describe('when portals were used to place an order', () => {
```
macrouch, line 144: This logic is part of PortalLinkContainer, not DownloadHistory, you don't need to test it here

static/src/js/components/DownloadHistory/__tests__/DownloadHistory.test.jsx
macrouch: I'd just mock PortalLinkContainer and test that you are giving it the correct props. That will simplify the test setup quite a bit

```
expect(PortalLinkContainer).toHaveBeenCalledTimes()
expect(PortalLinkContainer).toHaveBeenCalledWith()
```

static/src/js/containers/DownloadHistoryContainer/__tests__/DownloadHistoryContainer.test.js
```diff
@@ -1,89 +1,108 @@
...
-  mapStateToProps
-} from '../DownloadHistoryContainer'
-import { DownloadHistory } from '../../../components/DownloadHistory/DownloadHistory'
+import { render, screen } from '@testing-library/react'
+import { MemoryRouter } from 'react-router-dom'
+import { Provider } from 'react-redux'
+import configureStore from 'redux-mock-store'
 
-Enzyme.configure({ adapter: new Adapter() })
+import DownloadHistoryContainer, {
```
macrouch, line 8: We don't test the default exports of our containers, we test the named export. We do this in every container test so that we don't have to setup a full redux store, we just give it the props we want and test our own code

```
import {
  DownloadHistoryContainer,
  mapDispatchToProps,
  mapStateToProps
} from '../DownloadHistoryContainer'
```

static/src/js/containers/DownloadHistoryContainer/__tests__/DownloadHistoryContainer.test.js
```diff
@@ -1,89 +1,108 @@
...
-          created_at: '2019-08-25T11:58:14.390Z',
-          collections: [{}]
-        }],
-        retrievalHistoryLoading: false,
-        retrievalHistoryLoaded: false,
-        onFetchRetrievalHistory: jest.fn(),
-        onDeleteRetrieval: jest.fn()
-      })
+  describe('mapDispatchToProps', () => {
+    test('should dispatch the handleError action with the correct error configuration', () => {
```
macrouch, line 85: This needs to match every mapDispatchToProps test in the repo 

```
  test('dispatchHandleError calls actions.handleError', () => {
    const dispatch = jest.fn()
    const spy = jest.spyOn(actions, 'handleError')

    mapDispatchToProps(dispatch).dispatchHandleError({ mock: 'data' })

    expect(spy).toHaveBeenCalledTimes(1)
    expect(spy).toHaveBeenCalledWith({ mock: 'data' })
  })
```

macrouch (changes requested):

static/src/js/components/DownloadHistory/DownloadHistory.jsx
```diff
@@ -59,6 +59,15 @@ export const DownloadHistory = ({
     }
   }, [onDeleteRetrieval])
 
+  const handleRemoveClick = useCallback((id) => () => handleRemove(id), [handleRemove])
+
+  // Memoize the descriptions for all items in the history
+  const memoizedDescriptions = useMemo(() => retrievalHistory.reduce((acc, retrieval) => {
```
trevorlang, line 65: Minor thing but instead of calling this `memoizedDescriptions` which we dont really do for other memoized data, it could just be called `retrievalDescriptions` or something...

```suggestion
  const retrievalDescriptions = useMemo(() => retrievalHistory.reduce((acc, retrieval) => {
```

trevorlang (approved):
I am good with this once @macrouch give his approval

macrouch (approved):

PR #1911, opened 2025-05-09
EDSC-4489: Generate metrics reports

macrouch:
# Overview

### What is the feature?

Generate various metrics on a weekly schedule, by UAT and PROD releases

None

# Testing

This will run on a cron schedule, and update [this file](https://github.com/nasa/earthdata-search/blob/metrics/metrics.json), I've ran the job to populate that file with the current releases

# Checklist


bin/metrics/updateReport.mjs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// Add each value in reports to the metrics-branch/metrics.json file
+const metricsFilePath = 'metrics-branch/metrics.json'
+const metricsFileContent = fs.readFileSync(metricsFilePath, 'utf8')
+const metricsFile = JSON.parse(metricsFileContent)
+
+const updatedMetricsFile = {
+  ...metricsFile,
+  [version]: {
```
macrouch, line 41: Use only the sprint number for this version, add the full release number inside this object

macrouch (commented):
Delete the performance playwright tests

trevorlang (approved):
This is looking good. 

It's maybe a minor thing but if this were to be something that we wanted others to adopt, I feel like the pattern where youve used the sprint name as the key in the report might be too specific to our case and maybe a timestamp might work better for a key, or maybe an array. You'd need some logic to loop through and update the right thing when collecting from a subsequent env or when performing a backport. Or maybe theres merit to @abbottry's idea to generate a sqlite db with the data instead. Either way, it might make it easier/more reliable when trying to view in another metrics application we might expose this too.

Totally aware I am probably overthinking this at this stage and it could just be something we keep in mind if we want to share or iterate on this.

bin/metrics/parseBuildFile.mjs
```diff
@@ -0,0 +1,74 @@
...
+
+    const lineParts = line.split(' ').filter(Boolean)
+
+    if (lineParts[0]?.startsWith('static/dist')) {
+      numFiles += 1
+
+      const [
+        name,
+        size,
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
mandyparson, line 33: any reason why we can't disable the entire file for no unused vars?

bin/metrics/updateReport.mjs
```diff
@@ -0,0 +1,62 @@
...
+  })
+} catch (error) {
+  console.log('No reports found in metrics-reports directory.', error)
+
+  // If no reports are found, exit the script
+  process.exit(0)
+}
+
+// Add each value in reports to the metrics-branch/metrics.json file
+const metricsFilePath = 'metrics-branch/metrics.json'
```
mandyparson, line 35: Do we need to add an initial metrics.json to this PR so that the below functions work?

bin/metrics/parseBuildFile.mjs
macrouch: Disabling an eslint rule on the entire file is a bad idea, it can be very easy to break that rule unintentionally if it is ignored on the whole file. Disabling an eslint rule at all should be a last resort, and done on a single line to ensure you are only ignoring the specific issue. 

You could use `eslint-disable` followed by an `eslint-enable` to catch just a block of code, seen here (https://github.com/nasa/earthdata-search/blob/main/static/src/js/components/CustomToggle/CustomToggle.jsx#L67), but I lean towards just each line being disabled

mandyparson (approved):

eudoroolivares2016 (approved):

PR #1914, opened 2025-05-15
EDSC-4138: Adding new sort key "Recent Version" for collections

dpesall:
# Overview

### What is the feature?

- Users can now sort collections by the most recent version. So for collections that share the same name, the newest version will appear before the others.
- When sorting by a key other than Recent Version, collections will secondarily be sorted by Recent Version after being sorted by the selected sort key.

### What is the Solution?

- Added new sort key to collection sort dropdown menu
- Updated user preferences page to support saving Recent Version as the saved collection sorting key.

- Collection search panel
- User preferences

# Testing

### Reproduction steps

For testing, searching the keyword `CYGNSS` in UAT will yield some useful collections for testing this.

1. Navigate toe EDSC search page
2. Select the Recent Version sort key under collection sorting menu
3. Search for a collection that has multiple versions with the same name
4. Validate that the collection appearing at the top is a more recent version than the one(s) appearing below
5. Select a different sort key other than Recent Version
6. Validate that collections are sorted by Recent Version after being sorted by the selected sorting option.
7. Repeat steps 5-6 for all sort keys

# Checklist

### DIFF
--- a/static/src/js/util/collections.js
+++ b/static/src/js/util/collections.js
@@ -221,6 +221,7 @@ export const buildCollectionSearchParams = (params) => {
   // Only include has_granules_or_cwic sort key if the parameter is being used
   if (hasGranulesOrCwic) sortKey.unshift('has_granules_or_cwic')
 
+  // Add Recent Version as a secondary sort key for all collection searches
   if (selectedSortKey[0] !== collectionSortKeys.recentVersion) {
     sortKey.push(collectionSortKeys.recentVersion)
   }


static/src/js/util/collections.js
```diff
@@ -220,6 +221,10 @@ export const buildCollectionSearchParams = (params) => {
   // Only include has_granules_or_cwic sort key if the parameter is being used
   if (hasGranulesOrCwic) sortKey.unshift('has_granules_or_cwic')
 
+  if (selectedSortKey[0] !== collectionSortKeys.recentVersion) {
```
eudoroolivares2016, line 224: This custom logic does not seem right we don't do this for other sort keys

dpesall: This is because we are appending the sort key of `recentVersion` to the sort key array in all cases, unless we have already selected the `recentVersion` sort key as our primary method of sorting. We only do this for this key because we only want to add `recentVersion` as the secondary sort method for other sort keys.

macrouch (approved):

static/src/js/util/collections.js
eudoroolivares2016, line 224:
```suggestion
  // Add a secondary sort-key to CMR queries to ensure we are always sorting collections by their most recent version
  if (selectedSortKey[0] !== collectionSortKeys.recentVersion) {
```
Definitely does not have to be that exact phrasing above

eudoroolivares2016 (approved):

PR #1918, opened 2025-05-28
EDSC-4502: Fixes bug in EDD redirect

macrouch:
# Overview

### What is the feature?

Implementing Zustand has broken the EDD redirect after the user logs in. UrlQueryContainer is being re-rendered an extra time because Redux props are changing, but the Zustand store is also changing. A `setTimeout` in AuthCallbackContainer wrapping the redirect to `/earthdata-download-callback` solves this issue, but it should be removed after UrlQueryContainer is removed (after Zustand is handling the URL)

Earthdata Download redirects

# Testing

Delete your token from EDD, then start a new download. You will get redirected through EDSC to login with EDL, then ensure EDSC redirects you back to EDD

# Checklist

### DIFF
--- a/static/src/js/containers/AuthCallbackContainer/AuthCallbackContainer.jsx
+++ b/static/src/js/containers/AuthCallbackContainer/AuthCallbackContainer.jsx
@@ -61,7 +61,12 @@ export const AuthCallbackContainer = ({
         })
 
         // Redirect to the edd callback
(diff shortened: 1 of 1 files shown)


trevorlang (approved):
This feels like a possibly unpredictable solution but if we're sure it's okay and its not going to live long, I'm okay with it. We should do our best to try and break it when testing in SIT.

PR #1923, opened 2025-06-17
EDSC-4515: Removes facetsParams from Redux

macrouch:
# Overview

### What is the feature?

Removes facetsParams from Redux
Removes autocomplete selection display
When selecting an autocomplete suggestion, opens the facet group the selected autocomplete belongs to

Facets, autocomplete

# Testing

General regression test of Facets and autocomplete
Ensure landing page topics open as expected

# Checklist


static/src/js/zustand/slices/createFacetParamsSlice.ts
```diff
@@ -0,0 +1,132 @@
...
+    },
+    addCmrFacetFromAutocomplete: (facet) => {
+      const [facetType] = Object.keys(facet) as FacetKeys[]
+      const facetValue = facet[facetType]
+
+      set((state) => {
+        state.facetParams.cmrFacets = {
+          ...state.facetParams.cmrFacets,
+          [facetType]: [
+            // TODO why does this work with ScienceKeywordFacet[]?
```
macrouch: Remove

static/src/js/zustand/types.ts
```diff
@@ -25,9 +112,9 @@ export type HomeSlice = {
     /** Function to set the startDrawing value */
     setStartDrawing: (startDrawing: boolean | string) => void
     /** Flag if the Keyword facet group should be opened */
```
macrouch: Update comments

static/src/js/util/request/request.ts
```diff
@@ -1,13 +1,13 @@
 import axios, { CancelTokenSource } from 'axios'
 import { v4 as uuidv4 } from 'uuid'
 
-// @ts-expect-error Types are not defined for this module
-import configureStore from '../../store/configureStore'
-// @ts-expect-error Types are not defined for this module
-import { metricsTiming } from '../../middleware/metrics/actions'
+// // @ts-expect-error Types are not defined for this module
```
macrouch: Remove

dpesall (approved):

static/src/js/components/Facets/__tests__/FacetsGroup.test.jsx
```diff
@@ -0,0 +1,142 @@
...
+    })
+  })
+
+  describe('when the group is open', () => {
+    test('when rendering valid facets', async () => {
+      setup()
+
+      expect(FacetsList).toHaveBeenCalledTimes(0)
+
+      jest.clearAllMocks()
```
eudoroolivares2016, line 122: do we need this here?

static/src/js/components/Facets/__tests__/FacetsModal.test.jsx
```diff
@@ -1,230 +1,402 @@
...
-          isLoading: false,
-          selectedCategory: 'Test Category'
-        },
-        isOpen: true
+          isOpen: true
+        }
       })
 
-      enzymeWrapper.instance().onModalClose()
+      EDSCModalContainer.mock.calls[0][0].onClose()
```
eudoroolivares2016, line 301: theres no way to do this with just a user.click on the button?

eudoroolivares2016:
Testing locally looks good aside from an existing issue with view all on the data-format
timing event also looks correct I do see a change in how the `virtualPageView ` which isn't including the name in the API call 
image
(reproduce by clicking one of the topics on the home page)

static/src/js/components/Facets/__tests__/FacetsGroup.test.jsx
macrouch: While it isn't strictly necessary, this is a valid use for clearAllMocks. (beforeEach is where we don't need them)

It is used in a case like this to 
1. Make an assertion on a mock
2. Clear the mock counters
3. Perform an action that will call the mock
4. Make another assertion on the mock

It is useful to reset those counters just to make the code easier to understand. I check something got called once, and I want it to get called one time after I do something: you could have the second assertion check that it had been called twice, or you can reset the counter so it shows one call after the action.

It is a little weird having it here because it was called 0 times before, but I like this pattern so I want to leave the example in the repo

static/src/js/components/Facets/__tests__/FacetsModal.test.jsx
macrouch: I think it is because this is the contents of a modal, and it isn't loading the modal itself, or because we are mocking EDSCModalContainer, but there is no button to press.

static/src/js/components/Facets/__tests__/FacetsGroup.test.jsx
eudoroolivares2016: That does make sense thank you!

PR #1926, opened 2025-06-30
EDSC-4522: Fixes bug with cloc report

macrouch:
# Overview

### What is the feature?

The cloc npm library had an update that caused our job to fail to create a report. Instead of using that wrapper around clock, I've updated the metrics GitHub Actions workflow to download cloc directly.

Metrics workflow


KushaalP:
Everything looks good to me. I can see the cloc report.

KushaalP (approved):
Everything looks good to me. I can see the cloc report.

PR #1928, opened 2025-07-02
EDSC-4512: Move preferences from Redux to Zustand

dpesall:
# Overview

### What is the feature?

- Removing the Redux code handling preferences and implement Zustand to handle the preferences state.

### What is the Solution?

- Creating a new Zustand slice for preferences
- Removing Redux files for preferences and rewriting preferences state references and updates to use Zustand.

- User preferences
- URL parameters

# Testing

### Reproduction steps

1. Full regression testing on user preferences and URL parameters

# Checklist


static/src/js/actions/__tests__/focusedCollection.test.js
```diff
@@ -54,6 +54,7 @@ const createVariableResults = () => [{
 beforeEach(() => {
   jest.clearAllMocks()
   jest.restoreAllMocks()
+  useEdscStore.setState(useEdscStore.getInitialState())
```
dpesall: Line probably not needed

dpesall: Also clearAllMocks should be removable

static/src/js/actions/__tests__/urlQuery.test.js
```diff
@@ -13,6 +13,7 @@ const mockStore = configureMockStore([thunk])
 
 beforeEach(() => {
   jest.restoreAllMocks()
+  useEdscStore.setState(useEdscStore.getInitialState())
```
dpesall: Same as above comment

static/src/js/containers/PreferencesContainer/PreferencesContainer.jsx
```diff
@@ -1,32 +1,55 @@
...
+      preferences={preferences}
+      onUpdatePreferences={updatePreferences}
+    />
+  )
 }
 
-export default withRouter(
-  connect(mapStateToProps, mapDispatchToProps)(PreferencesContainer)
-)
+export default withRouter(PreferencesContainer)
```
dpesall: Look at where PreferencesContainer is being rendered and just go down to that level and retrieve Zustand store at that level.

static/src/js/containers/SearchPanelsContainer/SearchPanelsContainer.jsx
```diff
@@ -94,35 +94,45 @@ export const SearchPanelsContainer = ({
...
-)
+}) => {
+  const preferences = useEdscStore((state) => ({
+    panelState: state.preferences.panelState,
+    collectionListView: state.preferences.collectionListView,
+    granuleListView: state.preferences.granuleListView,
+    collectionSort: state.preferences.collectionSort,
+    granuleSort: state.preferences.granuleSort,
+    mapView: state.preferences.mapView
+  }))
```
dpesall: Use state.preferences instead

static/src/js/containers/SearchPanelsContainer/__tests__/SearchPanelsContainer.test.jsx
```diff
@@ -58,15 +59,14 @@ const setup = setupTest({
...
-      panelState: 'default',
-      collectionListView: 'default',
-      granuleListView: 'default'
     }
   }
 })
 
+beforeEach(() => {
+  useEdscStore.setState(useEdscStore.getInitialState())
+})
```
dpesall: Probably not needed

static/src/js/zustand/slices/createPreferencesSlice.ts
```diff
@@ -0,0 +1,268 @@
...
+        state.preferences.granuleSort = granuleSort
+      })
+    },
+
+    setMapView: (mapView) => {
+      set((state) => {
+        state.preferences.mapView = {
+          ...state.preferences.mapView,
+          ...mapView
+        }
```
dpesall: We can grab these from setPreferences, not needed here

static/src/js/zustand/slices/createPreferencesSlice.ts
```diff
@@ -0,0 +1,268 @@
...
+      })
+    },
+
+    setPreferencesFromJwt: (jwtToken) => {
+      if (!jwtToken) return
+
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
dpesall: Don't do this

static/src/js/zustand/types.ts
```diff
@@ -163,6 +163,55 @@ export type MapSlice = {
...
+    latitude: number
+    baseLayer: string
+    longitude: number
+    projection: string
+    overlayLayers: string[]
+    rotation: number
+  }
+  isSubmitting: boolean
+  isSubmitted: boolean
+}
```
dpesall: Add comments

static/src/js/zustand/types.ts
```diff
@@ -163,6 +163,55 @@ export type MapSlice = {
...
+    /** Function to set collection list view */
+    setCollectionListView: (collectionListView: string) => void
+    /** Function to set granule list view */
+    setGranuleListView: (granuleListView: string) => void
+    /** Function to set collection sort preference */
+    setCollectionSort: (collectionSort: string) => void
+    /** Function to set granule sort preference */
+    setGranuleSort: (granuleSort: string) => void
+    /** Function to set map view preferences */
+    setMapView: (mapView: Partial<PreferencesState['mapView']>) => void
```
dpesall: Not needed

static/src/js/zustand/types.ts
```diff
@@ -163,6 +163,55 @@ export type MapSlice = {
...
+    /** Function to set preferences from JWT token */
+    setPreferencesFromJwt: (jwtToken: string) => void
+    /** Function to update preferences via API */
+    updatePreferences: (data: { formData: PreferencesState }) => Promise<void>
```
dpesall: Try to make this more clear that this is the submission of the preferences

static/src/js/containers/SearchPanelsContainer/SearchPanelsContainer.jsx
dpesall: I can do `const preferences = useEdscStore((state) => state.preferences)` but apparently it is not recommended to pass the unneeded function references due to worse performance and (while not relevant in this case) if any of the functions got called on this component it would cause an unnecessary re-render.

macrouch: Sounds good, thanks for looking into that

static/src/js/routes/Preferences/Preferences.jsx
```diff
@@ -20,7 +59,10 @@ export const Preferences = () => {
       <div className="route-wrapper route-wrapper--light route-wrapper--content-page">
         <div className="route-wrapper__content">
           <div className="route-wrapper__content-inner">
-            <PreferencesContainer />
+            <PreferencesComponent
```
macrouch: This PreferencesComponent is like the old container, the only thing it is doing is lazy loading the form, and passing the props down. I think we can delete that component and load the PreferencesForm here directly, load the zustand values within the form component and avoid passing any props.

We don't need to lazy load the PreferencesForm component, this route component is being lazy loaded in App.jsx

static/src/js/types/sharedTypes.ts
```diff
@@ -297,7 +297,16 @@ export type TimelineRequestParams = {
 }
 
 /** The request parameters for our request classes */
```
macrouch: This comment should stand attached to RequestParams

static/src/js/types/sharedTypes.ts
```diff
@@ -297,7 +297,16 @@ export type TimelineRequestParams = {
...
-export type RequestParams = TimelineRequestParams | CollectionRequestParams | ShapefileRequestParams
+/** The request parameters for a preferences request */
+export type PreferencesRequestParams = {
+  /** The preferences data */
+  preferences: Record<string, unknown>
```
macrouch: This should define whatever parameters we pass in the request, it isn't `unknown`

static/src/js/zustand/slices/createPreferencesSlice.ts
```diff
@@ -0,0 +1,223 @@
...
+    submitAndUpdatePreferences: async (data) => {
+      const { formData: preferences } = data
+
+      get().preferences.setIsSubmitting(true)
+
+      try {
+        const { getState: reduxGetState, dispatch: reduxDispatch } = configureStore()
+        const reduxState = reduxGetState()
+
+        const earthdataEnvironment = getEarthdataEnvironment(reduxState)
```
macrouch: I think this selector is overkill, in the timeline slice I just did 
```
      const {
        authToken,
        earthdataEnvironment
      } = reduxState

```

static/src/js/zustand/types.ts
```diff
@@ -163,6 +163,54 @@ export type MapSlice = {
...
+  }
+  /** Flag indicating if preferences are currently being submitted */
+  isSubmitting: boolean
+  /** Flag indicating if preferences have been submitted */
+  isSubmitted: boolean
+}
+
+export type PreferencesSlice = {
+  /** The Preferences Slice of the store */
+  preferences: PreferencesState & {
```
macrouch: In redux we had this 
```
preferences: {
  preferences: {} // all the values
  isSubmitting
  isSubmitted
}
```

but this has moved all of the preferences values into that top level with the submitting flags, and all the functions. If we keep the old structure of nesting all of the preferences values I think that makes the selectors you are using in components much easier
```
const preferences = useEdscStore((state) => state.preferences.preferences)
```

static/src/js/zustand/slices/createPreferencesSlice.ts
```diff
@@ -0,0 +1,223 @@
...
+            zoom
+          })
+        }
+      }
+    },
+
+    submitAndUpdatePreferences: async (data) => {
+      const { formData: preferences } = data
+
+      get().preferences.setIsSubmitting(true)
```
macrouch: I'm wondering if we need the setIsSubmitting function, or if we just call set() here.

```
set((state) => state.preferences.isSubmitting = true)
```
I think this would do the same thing.

Should we reserve using `set*` functions for access outside of zustand, like if a component needs to set some data? Or if the setter is more complex?

static/src/js/containers/UrlQueryContainer/UrlQueryContainer.jsx
```diff
@@ -38,15 +37,14 @@ export const mapStateToProps = (state) => ({
...
-  mapPreferences: getMapPreferences(state),
   onlyEosdisCollections: state.query.collection.onlyEosdisCollections,
   overrideTemporalSearch: state.query.collection.overrideTemporal,
   pathname: state.router.location.pathname,
   pointSearch: state.query.collection.spatial.point,
   polygonSearch: state.query.collection.spatial.polygon,
   project: state.project,
   query: state.query,
-  paramCollectionSortKey: getCollectionSortKeyParameter(state),
+  paramCollectionSortKey: state.query.collection.paramCollectionSortKey,
```
macrouch: This is where the logic came from that you added into url.js. The logic needs to be removed from url.js, it is now handling that parameter differently than all the other parameters. 

But I don't know if I have an answer to how I think this should work yet (we should pair again tomorrow on the selector's stuff).

static/src/js/containers/SearchPanelsContainer/SearchPanelsContainer.jsx
macrouch: See my note about the structure of the preferences slice, I think this could be easier

static/src/js/containers/SearchPanelsContainer/__tests__/SearchPanelsContainer.test.jsx
macrouch, line 1: `expect(SearchPanels).toHaveBeenCalledWith` should have failed because you were removing props and you didn't add in a defaultZustandState to the setup function.

Within the test `passes its props and renders a single SearchPanels component`, remove the `expect.objectContaining` and we will get a more reliable test. Verifying every prop ensures we are alerting to any of the props changing

static/src/js/containers/AuthTokenContainer/__tests__/AuthTokenContainer.test.jsx
```diff
@@ -109,10 +135,18 @@ describe('AuthTokenContainer component', () => {
         disableDatabaseComponents: 'true'
       }))
 
+      const mockSetPreferencesFromJwt = jest.fn()
```
macrouch: If you are defining this function, add assertions for it being called the right number of times and with the right data if it is called

static/src/js/containers/AuthTokenContainer/__tests__/AuthTokenContainer.test.jsx
```diff
@@ -90,13 +85,44 @@ describe('AuthTokenContainer component', () => {
...
-    expect(props.onSetPreferencesFromJwt).toHaveBeenCalledTimes(1)
-    expect(props.onSetPreferencesFromJwt).toHaveBeenCalledWith('token')
+    expect(mockSetPreferencesFromJwt).toHaveBeenCalledTimes(1)
+    expect(mockSetPreferencesFromJwt).toHaveBeenCalledWith('token')
 
     expect(props.onSetUserFromJwt).toHaveBeenCalledTimes(1)
     expect(props.onSetUserFromJwt).toHaveBeenCalledWith('token')
   })
 
+  test('should use Zustand for preferences processing', () => {
```
macrouch: Is this test a duplicate of the previous test? I'm not sure what is different

static/src/js/actions/urlQuery.js
```diff
@@ -227,8 +227,8 @@ export const changePath = (path = '') => async (dispatch, getState) => {
     await dispatch(actions.getProjectGranules())
   }
 
-  const zustandState = useEdscStore.getState()
-  const { timeline } = zustandState
+  const zustand = useEdscStore.getState()
```
macrouch: Keep `zustandState` to be consistent across the app, or you could just grab `{ timeline }`

static/src/js/actions/urlQuery.js
```diff
@@ -49,7 +48,8 @@ export const updateStore = ({
   const { location } = router
   const { pathname } = location
 
-  const collectionSortPreference = getCollectionSortPreference(state)
+  const zustand = useEdscStore.getState()
+  const collectionSortPreference = zustand.preferences.collectionSort
```
macrouch: Keep `zustandState` to be consistent across the app, or you could just grab `{ preferences }`

Also destructure down to `collectionSort`

static/src/js/actions/focusedCollection.js
```diff
@@ -447,7 +446,8 @@ export const changeFocusedCollection = (collectionId) => (dispatch, getState) =>
     }))
   } else {
     // Initialize a nested query element in Redux for the new focused collection
-    const granuleSortPreference = getGranuleSortPreference(state)
+    const zustandState = useEdscStore.getState()
+    const granuleSortPreference = zustandState.preferences.granuleSort
```
macrouch: Destructure, but also might replace with something else, tbd

static/src/js/zustand/slices/createPreferencesSlice.ts
```diff
@@ -0,0 +1,223 @@
...
+          overlayLayers[referenceLabelsIndex] = mapLayers.placeLabels
+        }
+      }
+
+      set((state) => {
+        state.preferences = {
+          ...state.preferences,
+          ...preferences
+        }
+      })
```
macrouch: does this work? 
```
state.preferences.preferences = preferences
```

static/src/js/zustand/selectors/preferences.js
```diff
@@ -0,0 +1,36 @@
+import { translateDefaultCollectionSortKey } from '../../util/collections'
+import useEdscStore from '../useEdscStore'
+
+/**
+ * Get map preferences from Zustand store
+ */
+export const getMapPreferences = () => {
```
macrouch: `state` needs to be passed in as a parameter here, it is the Zustand state

static/src/js/util/url/url.js
```diff
@@ -333,6 +333,7 @@ export const encodeUrlQuery = (props) => {
   })
 
   const mapParams = encodeMap(props.mapView, props.mapPreferences)
+  const collectionSortParams = { csk: props.paramCollectionSortKey }
```
macrouch: I don't think you need this here at all, the loop above on line 328 is meant to handle this parameter

static/src/js/containers/UrlQueryContainer/UrlQueryContainer.jsx
```diff
@@ -38,14 +41,14 @@ export const mapStateToProps = (state) => ({
...
   location: state.router.location,
-  mapPreferences: getMapPreferences(state),
   onlyEosdisCollections: state.query.collection.onlyEosdisCollections,
   overrideTemporalSearch: state.query.collection.overrideTemporal,
   pathname: state.router.location.pathname,
   pointSearch: state.query.collection.spatial.point,
   polygonSearch: state.query.collection.spatial.polygon,
   project: state.project,
   query: state.query,
+  mapPreferences: getMapPreferences(),
```
macrouch: `mapStateToProps` is for pulling values out of Redux, so this should be pulled out of `useEdscStore` below

static/src/js/zustand/selectors/preferences.js
```diff
@@ -0,0 +1,36 @@
...
+
+  return preferences?.mapView || {}
+}
+
+/**
+ * Get collection sort key parameter with preference comparison
+ * Returns null if the sort key matches user preference (to hide from URL)
+ */
+export const getCollectionSortKeyParameter = (state) => {
+  const paramCollectionSortKey = state?.query?.collection?.paramCollectionSortKey
```
macrouch: `state` is supposed to be the Zustand state passed in, not Redux

```
const {
    getState: reduxGetState
  } = configureStore()
  const reduxState = reduxGetState() 
  const paramCollectionSortKey = getParamCollectionSortKey(reduxState) // this is the old redux selector
```

static/src/js/zustand/selectors/preferences.js
```diff
@@ -0,0 +1,36 @@
...
+
+  if (!paramCollectionSortKey) {
+    return null
+  }
+
+  // Get user preference from Zustand store
+  const preferences = useEdscStore.getState().preferences?.preferences
```
macrouch: `const { preferences } = state.preferences`, or use a `getPreferences` selector function

static/src/js/components/Preferences/__tests__/PreferencesForm.test.jsx
```diff
@@ -11,48 +11,66 @@ import uiSchema from '../../../../../../schemas/sitePreferencesUISchema.json'
 import mapLayers from '../../../constants/mapLayers'
 import projectionCodes from '../../../constants/projectionCodes'
 
+import mockUseEdscStore from '../../../zustand/useEdscStore'
+
 Enzyme.configure({ adapter: new Adapter() })
```
macrouch: Rewrite with Testing Library

static/src/js/actions/focusedCollection.js
```diff
@@ -447,7 +446,11 @@ export const changeFocusedCollection = (collectionId) => (dispatch, getState) =>
     }))
   } else {
     // Initialize a nested query element in Redux for the new focused collection
-    const granuleSortPreference = getGranuleSortPreference(state)
+    const {
```
macrouch: Don't use the nested destructuring like this, 

```
const { preferences, timeline } = useEdscStore.getState()
const { preferences: preferencesValues } = preferences
const { granuleSort: granuleSortPreference } = preferencesValues
```

static/src/js/actions/focusedCollection.js
```diff
@@ -460,9 +463,7 @@ export const changeFocusedCollection = (collectionId) => (dispatch, getState) =>
     dispatch(actions.getFocusedCollection())
 
     // Fetch timeline data for the focused collection
-    const zustandState = useEdscStore.getState()
-    const { timeline } = zustandState
-    const { getTimeline } = timeline
+    const { timeline: { getTimeline } } = useEdscStore.getState()
```
macrouch: Don't do nested destructuring, but also timeline can be pulled out of the state above

static/src/js/actions/urlQuery.js
```diff
@@ -49,7 +48,11 @@ export const updateStore = ({
   const { location } = router
   const { pathname } = location
 
-  const collectionSortPreference = getCollectionSortPreference(state)
+  const {
```
macrouch: Let's keep this as a selector, `const collectionSortPreference = getCollectionSortPreference(useEdscStore.getState())`

static/src/js/actions/urlQuery.js
```diff
@@ -76,6 +79,7 @@ export const updateStore = ({
       }
     }))
 
+    // eslint-disable-next-line no-shadow
```
macrouch: Don't disable this rule

static/src/js/actions/urlQuery.js
```diff
@@ -102,6 +106,7 @@ export const updateStore = ({
     }))
   } else {
     // We always need to load the portal config
+    // eslint-disable-next-line no-shadow
```
macrouch: Don't disable this rule

static/src/js/containers/AuthTokenContainer/__tests__/AuthTokenContainer.test.jsx
```diff
@@ -109,10 +104,16 @@ describe('AuthTokenContainer component', () => {
         disableDatabaseComponents: 'true'
       }))
 
+      useEdscStore.setState((state) => ({
```
macrouch: I don't think this setState is actually doing anything

static/src/js/containers/SearchPanelsContainer/__tests__/SearchPanelsContainer.test.jsx
macrouch: Line 241, remove `expect.objectContaining`

static/src/js/containers/UrlQueryContainer/UrlQueryContainer.jsx
```diff
@@ -38,14 +41,14 @@ export const mapStateToProps = (state) => ({
...
   paramCollectionSortKey: getCollectionSortKeyParameter(state),
```
macrouch: This should also be removed from mapStateToProps as the selector is now in the zustand directory

static/src/js/zustand/selectors/preferences.js
```diff
@@ -0,0 +1,36 @@
+import { translateDefaultCollectionSortKey } from '../../util/collections'
+import useEdscStore from '../useEdscStore'
+
+/**
+ * Get map preferences from Zustand store
+ */
+export const getMapPreferences = () => {
+  const preferences = useEdscStore.getState().preferences?.preferences
```
macrouch: `export const getPreferences = (state) => state.preferences.preferences` would be a good selector, that path through the state is used in a few places in your code

static/src/js/zustand/slices/createPreferencesSlice.ts
```diff
@@ -0,0 +1,237 @@
...
+import projectionCodes from '../../constants/projectionCodes'
+import mapLayers from '../../constants/mapLayers'
+import { projectionConfigs } from '../../util/map/crs'
+
+/** Type for JWT token payload containing preferences */
+type JwtTokenPayload = {
+  preferences?: Partial<PreferencesData>
+}
+
+export const initialPreferencesData = {
```
macrouch: This object is exported, but I'm not sure if it is actually used anywhere else. If not, don't export it

static/src/js/zustand/slices/createPreferencesSlice.ts
```diff
@@ -0,0 +1,237 @@
...
+    setPreferences: (preferences) => {
+      set((state) => {
+        state.preferences.preferences = {
+          ...state.preferences.preferences,
+          ...preferences
+        }
+      })
+    },
+
+    setIsSubmitting: (isSubmitting) => {
```
macrouch: I think this can be removed after changing `submitAndUpdatePreferences` to call `set` directly

static/src/js/actions/__tests__/focusedCollection.test.js
```diff
@@ -885,17 +884,15 @@ describe('changeFocusedCollection', () => {
...
-      })
-
-      const store = mockStore({
+        },
         preferences: {
-          preferences: {
-            granuleSort: 'default'
-          }
+          ...useEdscStore.getInitialState().preferences,
+          granuleSort: 'default'
```
macrouch: I think this now needs to be nested within another `preferences` object. I think this shows up more times in the PR, double check everywhere is using the right structure for the slice

static/src/js/zustand/types.ts
```diff
@@ -163,6 +163,59 @@ export type MapSlice = {
...
+  preferences: PreferencesData
+  /** Flag indicating if preferences are currently being submitted */
+  isSubmitting: boolean
+  /** Flag indicating if preferences have been submitted */
+  isSubmitted: boolean
+}
+
+export type PreferencesSlice = {
+  /** The Preferences Slice of the store */
+  preferences: PreferencesState & {
```
macrouch: I'm having a hard time following the structure of `PreferencesSlice`, I think if you don't use `PreferencesState` it will be easier to understand.

```
export type PreferencesSlice = {
  preferences: {
    preferences: PreferencesData
    isSubmitting: boolean
    isSubmitted: boolean
    setPreferences: (preferences: Partial<PreferencesData>) => void
    ... // other functions
  }
```

static/src/js/components/Preferences/__tests__/PreferencesForm.test.jsx
```diff
@@ -1,70 +1,79 @@
...
+const setup = setupTest({
+  Component: PreferencesForm,
+  defaultZustandState
+})
 
 describe('PreferencesForm component', () => {
-  test('renders a Form component', () => {
-    const { enzymeWrapper, props } = setup()
+  beforeEach(() => {
+    useEdscStore.setState(useEdscStore.getInitialState(), true)
```
macrouch: This is unnecessary, the zustand store gets reset between tests through `__mocks__/zustand.ts`

static/src/js/components/Preferences/__tests__/PreferencesForm.test.jsx
```diff
@@ -1,70 +1,79 @@
-import React from 'react'
-import Enzyme, { shallow, mount } from 'enzyme'
-import Adapter from '@wojtekmaj/enzyme-adapter-react-17'
-import Form from '@rjsf/core'
-import { act } from 'react-dom/test-utils'
+import { screen } from '@testing-library/react'
+import '@testing-library/jest-dom'
```
macrouch: This is imported globally in `jestConfigs/test-env.ts`, so we don't have to include it in every test

static/src/js/components/Preferences/__tests__/PreferencesForm.test.jsx
```diff
@@ -1,70 +1,79 @@
...
 
 describe('PreferencesForm component', () => {
-  test('renders a Form component', () => {
-    const { enzymeWrapper, props } = setup()
+  beforeEach(() => {
+    useEdscStore.setState(useEdscStore.getInitialState(), true)
+  })
 
-    const form = enzymeWrapper.find(Form)
+  test('renders a Form component', () => {
```
macrouch, line 38: This is testing that a few form fields are in the document, but it isn't testing values of the form fields like the enzyme test was

static/src/js/components/Preferences/__tests__/PreferencesForm.test.jsx
```diff
@@ -1,70 +1,79 @@
...
-    expect(form.props().onSubmit).toEqual(props.onUpdatePreferences)
+    expect(screen.getByRole('button', { name: /submit/i })).toBeInTheDocument()
+    expect(screen.getByText(/Panel State/i)).toBeInTheDocument()
+    expect(screen.getByText(/Collection.*Sort/i)).toBeInTheDocument()
+    expect(screen.getByText(/Granule.*Sort/i)).toBeInTheDocument()
   })
 
-  test('onChange sets the state', () => {
-    const { enzymeWrapper } = setupMount()
+  test('onChange sets the state', async () => {
```
macrouch: This isn't a good name for a test. Enzyme was calling the onChange function, but that isn't how we write tests with Testing Library.

This test also is not performing any action, and it isn't checking that anything has changed

static/src/js/components/Preferences/__tests__/PreferencesForm.test.jsx
```diff
@@ -1,70 +1,79 @@
...
-        formData: {
+    expect(submitButton).toBeInTheDocument()
+    expect(screen.getByText(/Panel State/i)).toBeInTheDocument()
+    expect(screen.getByText(/Collection.*Sort/i)).toBeInTheDocument()
+    expect(screen.getByText(/Granule.*Sort/i)).toBeInTheDocument()
+
+    expect(submitButton).not.toBeDisabled()
+  })
+
+  test('updating store data updates the state', () => {
```
macrouch: This test isn't testing that anything has been updated, it has the same issues as the first test. 

But I don't think this test is necessary. If the second test performs an action and verifies the form data is updated, then that covers all of the logic in the component.

static/src/js/containers/AuthTokenContainer/__tests__/AuthTokenContainer.test.jsx
macrouch: Remove this

static/src/js/containers/SearchPanelsContainer/SearchPanelsContainer.jsx
```diff
@@ -94,35 +95,38 @@ export const SearchPanelsContainer = ({
...
-    onToggleAboutCwicModal={onToggleAboutCwicModal}
-    onTogglePanels={onTogglePanels}
-    onExport={onExport}
-    panels={panels}
-    preferences={preferences}
-    match={match}
-  />
-)
+}) => {
+  const preferences = useEdscStore((state) => getPreferences(state))
```
macrouch:
```
const preferences = useEdscStore(getPreferences)
```

static/src/js/zustand/selectors/preferences.js
```diff
@@ -0,0 +1,47 @@
...
+  const preferences = getPreferences(state)
+
+  return preferences?.collectionSort
+}
+
+/**
+ * Get collection sort key parameter with preference comparison
+ * Returns null if the sort key matches user preference (to hide from URL)
+ */
+export const getCollectionSortKeyParameter = (paramCollectionSortKey, state) => {
```
macrouch: selectors shouldn't have more than one parameter, they only receive the zustand state.

Think of these selectors as the exact same function you'd pass into `useEdscStore()`. The easy example is `getPreferences`:
```
useEdscStore((state) => state.preferences.preferences)
// is exactly the same as 
useEdscStore(getPreferences)
```

So you can't pass in paramCollectionSortKey, you have to go get it within the function
```
  const {
    getState: reduxGetState
  } = configureStore()
  const reduxState = reduxGetState() 
  const paramCollectionSortKey = getParamCollectionSortKey(reduxState) // this is the old redux selector
```

Where this is being used in `UrlQueryContainer` needs to be moved out of `mapStateToProps` and into the `useEdscStore` call

static/src/js/actions/__tests__/focusedCollection.test.js
```diff
@@ -885,17 +884,18 @@ describe('changeFocusedCollection', () => {
       useEdscStore.setState({
         timeline: {
           getTimeline: getTimelineMock
-        }
-      })
-
-      const store = mockStore({
+        },
         preferences: {
+          ...useEdscStore.getInitialState().preferences,
```
macrouch: I saw this syntax in another file and we should do this instead of the calls to `getInitialState()`

```
    useEdscStore.setState((state) => ({
      ...state,
      preferences: {
        ...state.preferences,
        preferences: {
          ...state.preferences.preferences, 
          granuleSort: 'default'
        }
      }
    }))
```

static/src/js/zustand/selectors/preferences.js
macrouch, line 1: This file needs unit tests

static/src/js/components/Preferences/__tests__/PreferencesForm.test.jsx
```diff
@@ -1,135 +1,130 @@
...
+    const worldImageryLayer = screen.getByLabelText('World Imagery', { selector: '[name="baseLayer"]' })
+    expect(worldImageryLayer).toBeChecked()
+
+    expect(screen.getByText(/Overlay Layers/i)).toBeInTheDocument()
   })
 
-  test('updating props updates the state', () => {
-    const { enzymeWrapper } = setup()
+  test('allows user to change form values', async () => {
+    const user = userEvent.setup()
```
eudoroolivares2016: you can just return this from setup if the prop gets passed to that setupTest wrapper

static/src/js/zustand/selectors/__tests__/preferences.test.js
```diff
@@ -0,0 +1,287 @@
...
+        preferences: {
+          collectionSort: 'default' // This translates to '-usage_score'
+        }
+      }
+    }
+
+    expect(getCollectionSortKeyParameter('-score', state)).toEqual('-score')
+  })
+
+  test('returns null when paramCollectionSortKey matches translated user preference', () => {
```
eudoroolivares2016: Super nitpick I would just use -usage_score in this test or have two tests here one with 'default' as well. Its just a bit more confusing what this is trying to test theres a little bit of code downstream to process the special 'default' string

static/src/js/zustand/selectors/preferences.js
eudoroolivares2016: We do still need to try to make this function without arguments looks like that conforms to standard zustand which would be good to try to stay on as long as we can definitely cleans up the implementation in the urlQueryContainer without it.

eudoroolivares2016:
I did do functional testing for all the preferences on this feature branch rebased on top of main
- [x]  base layer
- [x] Projection
- [x] zoom
- [x] overlay layers
- [x] Ensure map preferences are working lat, long including url changes
- [x] Panel state
- [x] Ensure granule sort preferences are working including url changes
- [x] Granule list view
- [x] Collection list view
- [x] Ensure collection sort preferences are working including url changes
- [x] Rotation also continued to work
- [x] Only esdis query continues to work

static/src/js/containers/UrlQueryContainer/UrlQueryContainer.jsx
```diff
@@ -66,26 +68,35 @@ export const UrlQueryContainer = (props) => {
...
-    shapefileId: state.shapefile.shapefileId,
-    timelineQuery: state.timeline.query,
-    twoDCoordinateSystemNameFacets: state.facetParams.cmrFacets.two_d_coordinate_system_name
-  }))
+  const zustandValues = useEdscStore((state) => {
+    // Get paramCollectionSortKey from Redux state
+    const paramCollectionSortKeyFromRedux = getParamCollectionSortKey(props)
+
+    // Get the selector function and apply it to the parameter
+    const collectionSortKeyParameterSelector = getCollectionSortKeyParameterSelector(state)
```
macrouch: Remove this, it should happen within `getCollectionSortKeyParameter`

static/src/js/containers/UrlQueryContainer/UrlQueryContainer.jsx
```diff
@@ -66,26 +68,35 @@ export const UrlQueryContainer = (props) => {
...
+      featureFacets: state.facetParams.featureFacets,
+      granuleDataFormatFacets: state.facetParams.cmrFacets.granule_data_format_h,
+      horizontalDataResolutionRangeFacets:
+        state.facetParams.cmrFacets.horizontal_data_resolution_range,
+      instrumentFacets: state.facetParams.cmrFacets.instrument_h,
+      latency: state.facetParams.cmrFacets.latency,
+      mapPreferences: getMapPreferences(state),
+      mapView: state.map.mapView,
+      organizationFacets: state.facetParams.cmrFacets.data_center_h,
+      paramCollectionSortKey: collectionSortKeyParameterSelector(paramCollectionSortKeyFromRedux),
```
macrouch: `paramCollectionSortKey: getCollectionSortKeyParameter(state)`

static/src/js/zustand/selectors/preferences.js
```diff
@@ -0,0 +1,48 @@
...
+  const preferences = getPreferences(state)
+
+  return preferences?.collectionSort
+}
+
+/**
+ * Creates a selector function for collection sort key parameter with preference comparison
+ * Returns a function that takes paramCollectionSortKey and returns null if it matches user preference
+ */
+export const getCollectionSortKeyParameterSelector = (state) => {
```
macrouch: should just be named `getCollectionSortKeyParameter `

static/src/js/zustand/slices/createPreferencesSlice.ts
```diff
@@ -0,0 +1,231 @@
...
+  preferences: initialPreferencesData,
+  isSubmitting: false,
+  isSubmitted: false
+}
+
+const createPreferencesSlice: ImmerStateCreator<PreferencesSlice> = (set, get) => ({
+  preferences: {
+    ...initialState,
+
+    setPreferences: (preferences) => {
```
macrouch, line 62: This function is not tested

static/src/js/zustand/slices/createPreferencesSlice.ts
```diff
@@ -0,0 +1,231 @@
...
+            longitude,
+            overlays,
+            projection: projection as ProjectionCode,
+            zoom
+          })
+        }
+      }
+    },
+
+    submitAndUpdatePreferences: async (data) => {
```
macrouch, line 168: This function is not tested

static/src/js/zustand/selectors/preferences.js
```diff
@@ -0,0 +1,48 @@
...
+  const preferences = getPreferences(state)
+  const userPrefSortKey = preferences?.collectionSort
+  const translatedUserPrefSortKey = translateDefaultCollectionSortKey(userPrefSortKey)
+
+  // Return a function that can be used to compare against paramCollectionSortKey
```
macrouch: Selectors can't return a function, they return a value. This selector should take in the zustand state, fetch the `paramCollectionSortKey` from redux, and then do the same logic the previous redux selector did. You can fetch the value from redux using this code

```
const {
    getState: reduxGetState
  } = configureStore()
  const reduxState = reduxGetState() 
  const paramCollectionSortKey = getParamCollectionSortKey(reduxState) // this is the old redux selector
```

static/src/js/zustand/selectors/__tests__/preferences.test.js
```diff
@@ -0,0 +1,308 @@
...
+        longitude: 0,
+        projection: 'epsg4326',
+        baseLayer: 'worldImagery',
+        overlayLayers: ['bordersRoads', 'placeLabels'],
+        rotation: 0
+      }
+    })
+  })
+
+  test('returns preferences when preferences structure exists', () => {
```
macrouch: How is this different than the previous test?

static/src/js/zustand/selectors/__tests__/preferences.test.js
```diff
@@ -0,0 +1,308 @@
...
+          panelState: 'default',
+          collectionListView: 'default',
+          granuleListView: 'default',
+          mapView: {
+            zoom: 3,
+            latitude: 0,
+            longitude: 0,
+            projection: 'epsg4326',
+            baseLayer: 'worldImagery',
+            overlayLayers: ['bordersRoads', 'placeLabels'],
```
macrouch: We have constants defined for these values so we don't have to use strings

static/src/js/zustand/selectors/__tests__/preferences.test.js
```diff
@@ -0,0 +1,308 @@
...
+      latitude: 39.5,
+      longitude: -98.35,
+      projection: 'epsg4326',
+      baseLayer: 'worldImagery',
+      overlayLayers: ['bordersRoads'],
+      rotation: 0
+    })
+  })
+
+  test('returns undefined when mapView is not set', () => {
```
macrouch: types.ts shows mapView is required.

static/src/js/zustand/selectors/__tests__/preferences.test.js
```diff
@@ -0,0 +1,308 @@
...
+          collectionSort: 'default',
+          granuleSort: 'default'
+        }
+      }
+    }
+
+    expect(getMapPreferences(state)).toBeUndefined()
+  })
+
+  test('returns undefined when preferences are undefined', () => {
```
macrouch: types.ts shows preferences is required.

static/src/js/zustand/selectors/preferences.js
macrouch, line 1: Switching this file and the test to .ts is probably a good idea

static/src/js/zustand/selectors/__tests__/preferences.test.js
```diff
@@ -0,0 +1,308 @@
...
+          collectionSort: 'default',
+          granuleSort: 'default'
+        }
+      }
+    }
+
+    expect(getCollectionSortPreference(state)).toEqual('default')
+  })
+
+  test('returns undefined when collectionSort is not set', () => {
```
macrouch: collectionSort is required

static/src/js/zustand/selectors/__tests__/preferences.test.js
```diff
@@ -0,0 +1,308 @@
...
+        preferences: {
+          granuleSort: 'default'
+        }
+      }
+    }
+
+    expect(getCollectionSortPreference(state)).toBeUndefined()
+  })
+
+  test('returns undefined when preferences are undefined', () => {
```
macrouch: preferences are required

eudoroolivares2016 (approved):

static/src/js/containers/UrlQueryContainer/__tests__/UrlQueryContainer.test.jsx
```diff
@@ -36,11 +44,19 @@ const setup = setupTest({
     temporalSearch: {},
     onChangePath: jest.fn(),
     onChangeUrl: jest.fn()
+  },
+  defaultZustandState: {
+    preferences: {
+      collectionSort: 'default'
```
macrouch: preferences.preferences.collectionSort

static/src/js/selectors/query.js
```diff
@@ -32,6 +32,16 @@ export const getCollectionsQuery = (state) => {
   return collection
 }
 
+/**
+ * Retrieve the paramCollectionSortKey from Redux
+ * @param {Object} state Current state of Redux
+ */
+export const getParamCollectionSortKey = (state) => {
```
macrouch, line 39: Missing a test

static/src/js/zustand/slices/__tests__/createPreferencesSlice.test.ts
```diff
@@ -0,0 +1,544 @@
...
+      }
+
+      mockHandleError = jest.fn()
+
+      mockConfigureStore.mockReturnValue({
+        getState: mockGetState,
+        dispatch: mockDispatch
+      })
+
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
macrouch: Don't disable eslint rules

static/src/js/zustand/slices/__tests__/createPreferencesSlice.test.ts
```diff
@@ -0,0 +1,544 @@
...
+      mockPreferencesRequest.mockImplementation(() => mockPreferencesRequestInstance as any)
+      mockActions.handleError = mockHandleError
+      mockAddToast.mockClear()
+    })
+
+    afterEach(() => {
+      jest.clearAllMocks()
```
macrouch: This happens globally

static/src/js/zustand/selectors/__tests__/preferences.test.ts
```diff
@@ -0,0 +1,229 @@
...
+    }))
+
+    const mockGetState = jest.fn()
+    mockConfigureStore.mockReturnValue({
+      getState: mockGetState
+    })
+  })
+
+  afterEach(() => {
+    jest.restoreAllMocks()
```
macrouch, line 127: This isn't necessary

static/src/js/zustand/selectors/__tests__/preferences.test.ts
```diff
@@ -0,0 +1,229 @@
...
+describe('getCollectionSortPreference', () => {
+  test('returns collection sort preference from the zustand state', () => {
+    const state = createMockState({
+      collectionSort: '-usage_score'
+    })
+
+    expect(getCollectionSortPreference(state)).toEqual('-usage_score')
+  })
+
+  test('returns default when collectionSort is default', () => {
```
macrouch: This test isn't providing any value, there is no difference in the logic between the previous test and this one.

static/src/js/zustand/selectors/__tests__/preferences.test.ts
```diff
@@ -0,0 +1,229 @@
...
+      collectionSort: '-score'
+    })
+
+    mockGetParamCollectionSortKey.mockReturnValue('-score')
+    expect(getCollectionSortKeyParameter(state)).toBeNull()
+  })
+
+  test('returns the paramCollectionSortKey when user has no preferences set', () => {
+    const state = createMockState({
+      collectionSort: undefined
```
macrouch: collectionSort is required in types.ts

static/src/js/zustand/selectors/__tests__/preferences.test.ts
```diff
@@ -0,0 +1,229 @@
...
+    expect(getCollectionSortKeyParameter(state)).toEqual('-usage_score')
+  })
+
+  test('handles undefined collectionSort preference by defaulting to usage_score', () => {
+    jest.spyOn(getApplicationConfig, 'getApplicationConfig').mockImplementation(() => ({
+      collectionSearchResultsSortKey: '-usage_score'
+    }))
+
+    const state = createMockState({
+      collectionSort: undefined
```
macrouch: collectionSort is required in types.ts

static/src/js/zustand/selectors/__tests__/preferences.test.ts
```diff
@@ -0,0 +1,229 @@
...
+  test('returns the paramCollectionSortKey when user has no preferences set', () => {
+    const state = createMockState({
+      collectionSort: undefined
+    })
+
+    mockGetParamCollectionSortKey.mockReturnValue('-score')
+    expect(getCollectionSortKeyParameter(state)).toEqual('-score')
+  })
+
+  test('handles different default sort key configuration', () => {
```
macrouch: The default sort key being different causes no change in the logic in the code, this test isn't useful

static/src/js/zustand/selectors/__tests__/preferences.test.ts
```diff
@@ -0,0 +1,229 @@
...
+  test('returns null when paramCollectionSortKey matches translated default user preference', () => {
+    const state = createMockState({
+      collectionSort: 'default' // This translates to '-usage_score'
+    })
+
+    mockGetParamCollectionSortKey.mockReturnValue('-usage_score')
+    expect(getCollectionSortKeyParameter(state)).toBeNull()
+  })
+
+  test('returns null when paramCollectionSortKey matches non-default user preference', () => {
```
macrouch: The value of the user preference is irrelevant, only if it matches the `paramCollectionSortKey`. This test isn't useful

static/src/js/zustand/selectors/__tests__/preferences.test.ts
```diff
@@ -0,0 +1,229 @@
...
+  test('returns null when paramCollectionSortKey matches direct user preference', () => {
+    const state = createMockState({
+      collectionSort: '-usage_score'
+    })
+
+    mockGetParamCollectionSortKey.mockReturnValue('-usage_score')
+    expect(getCollectionSortKeyParameter(state)).toBeNull()
+  })
+
+  test('returns null when paramCollectionSortKey matches translated default user preference', () => {
```
macrouch: The behavior of `translateDefaultCollectionSortKey` is tested in its own file, this test isn't providing any additional value

static/src/js/actions/__tests__/urlQuery.test.js
```diff
@@ -50,12 +50,18 @@ describe('updateStore', () => {
       }
     }
 
-    const store = mockStore({
+    useEdscStore.setState((state) => ({
```
macrouch: This is setting the collectionSort to 'default', but that is already the default value in the slice. This call to setState can be removed

static/src/js/actions/__tests__/urlQuery.test.js
```diff
@@ -181,16 +187,28 @@ describe('updateStore', () => {
       const storeActions = store.getActions()
       expect(storeActions[0]).toEqual({
         payload: {
-          ...params,
+          advancedSearch: undefined,
```
macrouch: These changes are unnecessary, the test still passes with these lines the same as main. It needs to stay consistent with the other assertions in similar tests, like on line 75

static/src/js/actions/__tests__/urlQuery.test.js
```diff
@@ -296,17 +314,23 @@ describe('updateStore', () => {
...
```
macrouch: These changes are unnecessary, the test still passes with these lines the same as main. It needs to stay consistent with the other assertions in similar tests, like on line 75

PR #1930, opened 2025-07-10
EDSC-4525: Update regionHost to new domain

frankinspace:
# Overview

### What is the feature?

Change to new domain for Feature Translation Service. This is a drop-in replacement, request/response structure has not changed.

### What is the Solution?

Replaced URL

Advanced search

# Testing

### Reproduction steps

1. Ensure that advanced Search continues to work as expected

### Attachments

N/A

# Checklist


eudoroolivares2016:
Ran locally included change in my overridestatic all the advanced search options did work when given the correct input the only thing that was different was when given an incorrect input the response must have changed tried it against the old domain and it had a different response. I think we just need a minor change in the processing for the outgoing lambda `serverless/src/regionSearch/handler.js`

HUCID: [031700](https://search.usa.gov/search?affiliate=usgs&utf8=%E2%9C%93&query=Pascagoula.%20Mississippi%20&charset=iso-8859-1)
River ID: 82242100021
Huc region: Colorado Mine

image

frankinspace:
@eudoroolivares2016 I was just trying some invalid input as well and was wondering if this matches your observation.

**New Domain**
Request:
`soto.podaac.earthdatacloud.nasa.gov/fts/v1/huc/two`
Response:
```
200 OK
Content-Type: text/html
Content: 
<!doctype html>
<html>
<head>
    <title>Page not found - NASA Worldview</title>
...
```

**Old Domain**
Request:
`fts.podaac.earthdata.nasa.gov/v1/huc/two`
Response:
```
404 Not Found
Content-Type: 
application/json
Content: 
{
    "error": "404: Results with the specified HUC two were not found."
}
```

I believe we actually want the response to mirror the behaviour of the old domain (returning a 404). I've logged a bug ticket to fix this https://github.com/podaac/feature-translation-service/issues/23

But if we want to update EDSC to be able to handle this situation as well it would make it more robust overall.

serverless/src/regionSearch/handler.js
```diff
@@ -35,6 +35,16 @@ const regionSearch = async (event) => {
...
     const { elapsedTime } = config
 
+    const contentType = regionResponse.headers?.['content-type'] || ''
+    const responseBody = typeof regionResponse.data === 'string' ? regionResponse.data : ''
+    if (
+      regionResponse.status === 200
+      && contentType.includes('text/html')
+      && /page not found/i.test(responseBody)
+    ) {
+      throw new Error('Not Found')
```
macrouch: We should return the same result as we previously did. With this change the user will see `Error: Not Found`, but in prod today they will see `404: Results with the specified HUC 2134123412341243 were not found`. 

I think we can achieve something closer with this:
```
      return {
        isBase64Encoded: false,
        headers: defaultResponseHeaders,
        statusCode: 404,
        body: JSON.stringify({
          errors: [`404: Results with the query ${query} were not found.`]
        })
      }
```

It won't show HUC or reach like the API did, but providing the value they submitted is nice

frankinspace: Fair enough, once https://github.com/podaac/feature-translation-service/issues/23 is addressed it will revert to previous behaviour anyway

macrouch (approved):

eudoroolivares2016 (approved):
Error response looks good now thanks
image

PR #1941, opened 2025-08-06
EDSC-4521: Adding MultiPolygon, MultiLine, and MultiPoint support

dpesall:
# Overview

### What is the feature?

Adding support for shapefiles of type `MutliPolygon`, `MutliLine`, and `MutliPoint`

### What is the Solution?

- Added handling for the shapefile structure of these new types to be gracefully handled and sent to CMR as an array of Polygons, Lines, or Points.

- getQueryFromShapefile
- onClickShapefile

# Testing

### Reproduction steps

1. Navigate to EDSC
2. Upload a MultiPolygon shapefile
3. Verify that collections are filtered properly
4. Verify the URL reflects both polygons

### Attachments

Screenshot 2025-08-06 at 10 16 59 AM

# Checklist

### DIFF
--- a/static/src/js/actions/__tests__/viewAllFacets.test.js
+++ b/static/src/js/actions/__tests__/viewAllFacets.test.js
@@ -18,6 +18,8 @@ import {
   TOGGLE_VIEW_ALL_FACETS_MODAL
 } from '../../constants/actionTypes'
 
+import * as collectionUtils from '../../util/collections'
+
 const mockStore = configureMockStore([thunk])
 
 describe('updateViewAllFacets', () => {
@@ -142,24 +144,28 @@ describe('getViewAllFacets', () => {
       .post(/collections/)
       .reply(200, stubResponse, { 'cmr-hits': 1 })
 
+    const prepareCollectionParamsSpy = jest.spyOn(collectionUtils, 'prepareCollectionParams')
+
     // MockStore with initialState
     const store = mockStore({
       authToken: '',
       searchResults: {
-        collections: {},
         facets: {},
-        granules: {},
         viewAllFacets: {}
-      },
-      query: {
-        collection: {
-          keyword: 'search keyword'
-        }
       }
     })
 
     // Call the dispatch
     await store.dispatch(getViewAllFacets('Instruments')).then(() => {
+      expect(prepareCollectionParamsSpy).toHaveBeenCalledTimes(1)
+      expect(prepareCollectionParamsSpy).toHaveBeenCalledWith({
+        authToken: '',
+        searchResults: {
+          facets: {},
+          viewAllFacets: { selectedCategory: 'Instruments' }
+        }
+      })
+
       const storeActions = store.getActions()
       expect(storeActions[0]).toEqual({
         type: LOADING_VIEW_ALL_FACETS,
@@ -195,24 +201,28 @@ describe('getViewAllFacets', () => {
         'jwt-token': 'token'
       })
 
+    const prepareCollectionParamsSpy = jest.spyOn(collectionUtils, 'prepareCollectionParams')
+
     // MockStore with initialState
     const store = mockStore({
       authToken: 'token',
       searchResults: {
-        collections: {},
         facets: {},
-        granules: {},
         viewAllFacets: {}
-      },
-      query: {
-        collection: {
-          keyword: 'search keyword'
-        }
       }
     })
 
     // Call the dispatch
     await store.dispatch(getViewAllFacets('Instruments')).then(() => {
+      expect(prepareCollectionParamsSpy).toHaveBeenCalledTimes(1)
+      expect(prepareCollectionParamsSpy).toHaveBeenCalledWith({
+        authToken: 'token',
+        searchResults: {
+          facets: {},
+          viewAllFacets: { selectedCategory: 'Instruments' }
+        }
+      })
+
       const storeActions = store.getActions()
       expect(storeActions[0]).toEqual({
         type: LOADING_VIEW_ALL_FACETS,
@@ -251,24 +261,28 @@ describe('getViewAllFacets', () => {
       .post(/error_logger/)
       .reply(200)
 
+    const prepareCollectionParamsSpy = jest.spyOn(collectionUtils, 'prepareCollectionParams')
+
     // MockStore with initialState
     const store = mockStore({
       authToken: '',
       searchResults: {
-        collections: {},
         facets: {},
-        granules: {},
         viewAllFacets: {}
-      },
-      query: {
-        collection: {
-          keyword: 'search keyword'
-        }
       }
     })
 
     // Call the dispatch
     await store.dispatch(getViewAllFacets('Instruments')).then(() => {
+      expect(prepareCollectionParamsSpy).toHaveBeenCalledTimes(1)
+      expect(prepareCollectionParamsSpy).toHaveBeenCalledWith({
+        authToken: '',
+        searchResults: {
+          facets: {},
+          viewAllFacets: { selectedCategory: 'Instruments' }
+        }
+      })
+
       const storeActions = store.getActions()
       expect(storeActions[0]).toEqual({
         type: LOADING_VIEW_ALL_FACETS,
--- a/static/src/js/actions/viewAllFacets.js
+++ b/static/src/js/actions/viewAllFacets.js
@@ -55,7 +55,18 @@ export const getViewAllFacets = (category = '') => (dispatch, getState) => {
 
   // `onViewAllFacetsLoading` changes the state, use getState() again here to ensure the
   // collection request has the updated state
-  const collectionParams = prepareCollectionParams(state)
+  // EDSC-4567: Temporary fix to get the View All facet modal working correctly. Once
+  // this moves to Zustand in EDSC-4561, this won't be necessary.
+  const collectionParams = prepareCollectionParams({
+    ...state,
+    searchResults: {
+      ...state.searchResults,
+      viewAllFacets: {
+        ...state.searchResults.viewAllFacets,
+        selectedCategory: category
+      }
+    }
+  })
 
   const requestObject = new CollectionRequest(authToken, earthdataEnvironment)
 
--- a/static/src/js/components/OrderStatus/OrderStatus.jsx
+++ b/static/src/js/components/OrderStatus/OrderStatus.jsx
@@ -48,6 +48,7 @@ export const OrderStatus = ({
   retrieval = {}
 }) => {
   const params = useParams()
+  const { id: paramsId } = params
   const earthdataEnvironment = useEdscStore(getEarthdataEnvironment)
 
   useEffect(() => {
@@ -60,7 +61,7 @@ export const OrderStatus = ({
 
   const {
     collections,
-    id,
+    id: loadedId,
     isLoaded,
     isLoading,
     jsondata = {},
@@ -140,11 +141,13 @@ export const OrderStatus = ({
 
   const eeLink = earthdataEnvironment === deployedEnvironment() ? '' : `?ee=${earthdataEnvironment}`
 
+  const shouldShowLoading = paramsId !== loadedId || (isLoading && !isLoaded)
+
   const introduction = (
     <p>
       {'This page will automatically update as your orders are processed. The Download Status page can be accessed later by visiting '}
-      <a href={`${edscHost}/downloads/${id}${eeLink}`}>
-        {`${edscHost}/downloads/${id}${eeLink}`}
+      <a href={`${edscHost}/downloads/${paramsId}${eeLink}`}>
+        {`${edscHost}/downloads/${paramsId}${eeLink}`}
       </a>
       {' or the '}
       <PortalLinkContainer
@@ -184,7 +187,7 @@ export const OrderStatus = ({
             <Well.Heading>Download Status</Well.Heading>
             <Well.Introduction>{introduction}</Well.Introduction>
             {
-              (isLoading && !isLoaded) && (
+              shouldShowLoading && (
                 <Skeleton
                   className="order-status__item-skeleton"
                   containerStyle={
@@ -218,7 +221,7 @@ export const OrderStatus = ({
             <Well.Heading>Additional Resources and Documentation</Well.Heading>
             <Well.Section>
               {
-                isLoading && (
+                shouldShowLoading && (
                   <Skeleton
                     className="order-status__item-skeleton"
                     containerStyle={
--- a/static/src/js/components/OrderStatus/__tests__/OrderStatus.test.jsx
+++ b/static/src/js/components/OrderStatus/__tests__/OrderStatus.test.jsx
@@ -3,12 +3,14 @@ import {
   waitFor,
   within
 } from '@testing-library/react'
+import React from 'react'
 import { Helmet } from 'react-helmet'
 
 import setupTest from '../../../../../../jestConfigs/setupTest'
 
-import { retrievalStatusProps } from './mocks'
+import { initalizedRetrievalStatusProps, retrievalStatusProps } from './mocks'
 import { OrderStatus } from '../OrderStatus'
+import Skeleton from '../../Skeleton/Skeleton'
 import * as config from '../../../../../../sharedUtils/config'
 
 jest.mock('react-router-dom', () => ({
@@ -17,6 +19,8 @@ jest.mock('react-router-dom', () => ({
     id: '7'
   })
 }))
+
+jest.mock('../../Skeleton/Skeleton', () => jest.fn(() => <div />))
 
 const setup = setupTest({
   Component: OrderStatus,
@@ -30,16 +34,89 @@ beforeEach(() => {
 })
 
 describe('OrderStatus component', () => {
-  test('renders itself correctly', () => {
+  test('renders itself correclty upon page load', () => {
+    setup({
+      overrideProps: initalizedRetrievalStatusProps
+    })
+
+    expect(screen.getByText('https://search.earthdata.nasa.gov/downloads/7')).toBeInTheDocument()
+    expect(Skeleton).toHaveBeenCalledTimes(2)
+    expect(Skeleton).toHaveBeenNthCalledWith(1, {
+      className: 'order-status__item-skeleton',
+      containerStyle: {
+        display: 'inline-block',
+        height: '175px',
+        width: '100%'
+      },
+      shapes: [{
+        height: 18,
+        left: 0,
+        radius: 2,
+        shape: 'rectangle',
+        top: 2,
+        width: 200
+      }, {
+        height: 14,
+        left: 0,
+        radius: 2,
+        shape: 'rectangle',
+        top: 31,
+        width: '80%'
+      }, {
+        height: 96,
+        left: 0,
+        radius: 2,
+        shape: 'rectangle',
+        top: 60,
+        width: '100%'
+      }]
+    }, {})
+
+    expect(Skeleton).toHaveBeenNthCalledWith(2, {
+      className: 'order-status__item-skeleton',
+      containerStyle: {
+        display: 'inline-block',
+        height: '175px',
+        width: '100%'
+      },
+      shapes: [{
+        height: 16,
+        left: 0,
+        radius: 2,
+        shape: 'rectangle',
+        top: 2,
+        width: '60%'
+      }, {
+        height: 14,
+        left: 20,
+        radius: 2,
+        shape: 'rectangle',
+        top: 31,
+        width: '50%'
+      }, {
+        height: 14,
+        left: 20,
+        radius: 2,
+        shape: 'rectangle',
+        top: 55,
+        width: '57%'
+      }]
+    }, {})
+  })
+
+  test('renders itself correctly after initial page load', () => {
     setup()
 
     expect(screen.getByText('Download Status')).toBeInTheDocument()
+    expect(Skeleton).toHaveBeenCalledTimes(0)
   })
 
   test('renders the correct Helmet meta information', async () => {
     setup()
 
     await waitFor(() => expect(document.title).toEqual('Download Status'))
+
+    expect(Skeleton).toHaveBeenCalledTimes(0)
 
     const helmet = Helmet.peek()
 
@@ -88,6 +165,7 @@ describe('OrderStatus component', () => {
 
       const link = screen.getByRole('link', { name: 'http://localhost/downloads/7' })
       expect(link).toBeInTheDocument()
+      expect(Skeleton).toHaveBeenCalledTimes(0)
     })
 
     test('download status link has correct href when earthdataEnvironment is different than the deployed environment', () => {
@@ -100,6 +178,7 @@ describe('OrderStatus component', () => {
 
       const link = screen.getByRole('link', { name: 'http://localhost/downloads/7?ee=prod' })
       expect(link).toBeInTheDocument()
+      expect(Skeleton).toHaveBeenCalledTimes(0)
     })
 
     test('status link has correct href', () => {
@@ -107,6 +186,7 @@ describe('OrderStatus component', () => {
 
       const link = screen.getByRole('link', { name: 'Download Status and History' })
       expect(link.href).toEqual('http://localhost/downloads')
+      expect(Skeleton).toHaveBeenCalledTimes(0)
     })
 
     test('status link has correct href when earthdataEnvironment is different than the deployed environment', () => {
@@ -119,6 +199,7 @@ describe('OrderStatus component', () => {
 
       const link = screen.getByRole('link', { name: 'Download Status and History' })
       expect(link.href).toEqual('http://localhost/downloads?ee=prod')
+      expect(Skeleton).toHaveBeenCalledTimes(0)
     })
   })
 
@@ -127,6 +208,7 @@ describe('OrderStatus component', () => {
       setup()
 
       expect(screen.getByRole('link', { name: 'http://linkurl.com/test' })).toBeInTheDocument()
+      expect(Skeleton).toHaveBeenCalledTimes(0)
     })
   })
 
@@ -145,6 +227,7 @@ describe('OrderStatus component', () => {
       expect(within(relatedCollections[0]).getByRole('link', { key: 'related-collection-TEST_COLLECTION_3_111' })).toBeInTheDocument()
       expect(within(relatedCollections[1]).getByRole('link', { key: 'related-collection-TEST_COLLECTION_2_111' })).toBeInTheDocument()
       expect(within(relatedCollections[2]).getByRole('link', { key: 'related-collection-TEST_COLLECTION_1_111' })).toBeInTheDocument()
+      expect(Skeleton).toHaveBeenCalledTimes(0)
     })
   })
 
@@ -160,6 +243,7 @@ describe('OrderStatus component', () => {
       })
 
       expect(props.onChangePath).toHaveBeenCalledWith('/search?test=source_link')
+      expect(Skeleton).toHaveBeenCalledTimes(0)
     })
   })
 })
--- a/static/src/js/components/OrderStatus/__tests__/mocks.js
+++ b/static/src/js/components/OrderStatus/__tests__/mocks.js
@@ -20,7 +20,7 @@ export const retrievalStatusProps = {
     portalId: 'edsc'
   },
   retrieval: {
-    id: 7,
+    id: '7',
     isLoading: false,
     isLoaded: true,
     collections: {
@@ -128,10 +128,27 @@ export const retrievalStatusProps = {
   }
 }
 
+export const initalizedRetrievalStatusProps = {
+  authToken: 'testToken',
+  granuleDownload: {},
+  onFetchRetrieval: jest.fn(),
+  onFetchRetrievalCollection: jest.fn(),
+  onFetchRetrievalCollectionGranuleLinks: jest.fn(),
+  onFetchRetrievalCollectionGranuleBrowseLinks: jest.fn(),
+  onMetricsRelatedCollection: jest.fn(),
+  onToggleAboutCSDAModal: jest.fn(),
+  retrieval: {
+    collections: {},
+    id: null,
+    isLoaded: false,
+    isLoading: false
+  }
+}
+
 export const retrievalStatusPropsTwo = {
   authToken: 'testToken2',
   retrieval: {
-    id: 7,
+    id: '7',
     collections: {
       byId: {
         1: {
@@ -165,7 +182,7 @@ export const retrievalStatusPropsTwo = {
 export const retrievalStatusPropsEsi = {
   authToken: 'testToken2',
   retrieval: {
-    id: 7,
+    id: '7',
     collections: {
       byId: {
         1: {
@@ -247,7 +264,7 @@ export const retrievalStatusPropsEsi = {
 export const retrievalStatusPropsEchoOrder = {
   authToken: 'testToken2',
   retrieval: {
-    id: 7,
+    id: '7',
     collections: {
       byId: {
         1: {


eudoroolivares2016:
Ran all of these locally and they were actually subsetting CMR
I noticed that multi points aren't getting drawn correctly, technically that is outside this scope

macrouch (commented):
Codecov is also showing your new additions to onClickShapefile are not being tested

static/src/js/util/map/interactions/__tests__/onClickShapefile.test.ts
macrouch, line 1: I don't think this is how we should test map functionality. So many things are mocked in this test in order to get at testing the logic, and if any of those mocks need to change it could be very difficult to determine what code needs to change in order to keep the test valid. 

The lines that needed to be tested were what happens when a shape is deselected, you can use the existing playwright tests we have in map_shapefile.spec.js and deselect the shape by clicking on it. Those tests verify everything we need already (the CMR query, URL values, and a screenshot of the map).

tests/e2e/map/map_shapefile.spec.js
```diff
@@ -587,5 +587,182 @@ test.describe('Map: Shapefile interactions', () => {
...
+
+        const initialMapPromise = page.waitForResponse(/World_Imagery\/MapServer\/tile\/2/)
+        await page.goto('/search')
+
+        // Wait for the map to load
+        await initialMapPromise
+
+        // Upload the shapefile
+        await uploadShapefile(page, 'multipoint.geojson')
+        await page.waitForTimeout(1000)
```
macrouch: Why is a waitForTimeout here? None of the existing tests needed this

tests/e2e/map/map_shapefile.spec.js
```diff
@@ -587,5 +587,182 @@ test.describe('Map: Shapefile interactions', () => {
...
+        // Checking that the right number of results are loaded ensures that the route
+        // was fulfilled correctly with the successful paramCheck
+        await expect(page.getByText('Showing 2 of 2 matching collections')).toBeVisible()
+
+        // Waiting for the URL to include the correct zoom level ensures the map is finished drawing - flexible zoom
+        await page.waitForURL(/zoom=11/, { timeout: 3000 })
+
+        // Draws the spatial on the map
+        await expect(page).toHaveScreenshot('multipoint.png', {
+          clip: screenshotClip
```
macrouch: Adding `maxDiffPixelRatio: 0.005` will have this test failing because the screenshots haven't been updated to have the points drawn in them. When we find tests like this that don't have very many pixels different with and without what we are looking for, we should lower that ration to make sure they fail when expected.

tests/e2e/map/map_shapefile.spec.js
```diff
@@ -359,8 +359,8 @@ test.describe('Map: Shapefile interactions', () => {
         // was fulfilled correctly with the succesfull paramCheck
         await expect(page.getByText('Showing 2 of 2 matching collections')).toBeVisible()
 
-        // Updates the URL
-        await expect(page).toHaveURL(/search\?polygon\[0\]=42.1875%2C-2.40647%2C42.1875%2C-16.46517%2C56.25%2C-16.46517%2C42.1875%2C-2.40647&polygon\[1\]=44.1875%2C0.40647%2C58.25%2C-14.46517%2C58.25%2C0.40647%2C44.1875%2C0.40647&sf=1&sfs\[0\]=0&sfs\[1\]=1&lat=-8\.\d+&long=46\.\d+&zoom=3\.\d+/)
+        // Updates the URL - accept multiple zoom levels
+        await expect(page).toHaveURL(/search\?polygon\[0\]=42.1875%2C-2.40647%2C42.1875%2C-16.46517%2C56.25%2C-16.46517%2C42.1875%2C-2.40647&polygon\[1\]=44.1875%2C0.40647%2C58.25%2C-14.46517%2C58.25%2C0.40647%2C44.1875%2C0.40647&sf=1&sfs\[0\]=0&sfs\[1\]=1&lat=-8\.\d+&long=46\.\d+&zoom=[3-5]\.\d+/)
```
macrouch: Why was this changed? Zoom level 3.0 to zoom 5.99 on my window size is the difference of seeing all of North America vs seeing only the width of Kansas. These assertions are designed to give some leniency, but accepting 3 or 4 different zoom levels is not ensuring the app is behaving the way we want it to. All of the assertions in this file that accept multiple zoom levels need to change

tests/e2e/map/map_shapefile.spec.js
```diff
@@ -587,5 +587,182 @@ test.describe('Map: Shapefile interactions', () => {
...
+        await expect(page.getByText('Showing 2 of 2 matching collections')).toBeVisible()
+
+        // Waiting for the URL to include the correct zoom level ensures the map is finished drawing - flexible zoom
+        await page.waitForURL(/zoom=11/, { timeout: 3000 })
+
+        // Draws the spatial on the map
+        await expect(page).toHaveScreenshot('multipoint.png', {
+          clip: screenshotClip
+        })
+
```
macrouch: All of these tests are missing an important assertion 

```
        // Checking that the right number of results are loaded ensures that the route
        // was fulfilled correctly with the succesfull paramCheck
        await expect(page.getByText('Showing 2 of 2 matching collections')).toBeVisible()
```

tests/e2e/map/map_shapefile.spec.js
macrouch: All of the other tests use the map tiles to ensure the test is ready to move on 

```
        const shapefilePromise = page.waitForResponse(/World_Imagery\/MapServer\/tile\/20/)
        await uploadShapefile(page, 'point.geojson')
        await shapefilePromise
```

tests/e2e/map/map_shapefile.spec.js
dpesall: I've now gone back to the single accepted zoom level, I had increased the bounds when I was having some issues and debugging, and I never reverted the wider threshold.

static/src/js/util/map/interactions/findShapefileFeature.ts
```diff
@@ -34,7 +34,7 @@ const findShapefileFeature = ({
 
   // If the feature is a point or line, we need to check if the coordinate is close
   // to the feature, not within it
-  const closeEnoughTypes = [spatialTypes.POINT, spatialTypes.LINE_STRING]
+  const closeEnoughTypes = [spatialTypes.POINT, spatialTypes.MULTI_POINT, spatialTypes.LINE_STRING]
```
macrouch:
```
  const closeEnoughTypes = [
    spatialTypes.POINT,
    spatialTypes.MULTI_POINT,
    spatialTypes.LINE_STRING,
    spatialTypes.MULTI_LINE_STRING
  ]
```

tests/e2e/map/map_shapefile.spec.js
```diff
@@ -361,6 +361,20 @@ test.describe('Map: Shapefile interactions', () => {
 
         // Updates the URL
         await expect(page).toHaveURL(/search\?polygon\[0\]=42.1875%2C-2.40647%2C42.1875%2C-16.46517%2C56.25%2C-16.46517%2C42.1875%2C-2.40647&polygon\[1\]=44.1875%2C0.40647%2C58.25%2C-14.46517%2C58.25%2C0.40647%2C44.1875%2C0.40647&sf=1&sfs\[0\]=0&sfs\[1\]=1&lat=-8\.\d+&long=46\.\d+&zoom=3\.\d+/)
+
+        // Test deselection by clicking on the lower polygon again
```
macrouch: Don't just add this to an existing test, isolate this functionality to its own test. Pull out all of the shared logic in this test into a `test.beforeEach` to share that logic with this test and the new test.

This test isn't verifying the correct parameters are sent to CMR, and it isn't verifying the shapes on the map are drawn correctly.

We also don't have a test that is verifying unselected multipoint shapes are drawn correctly

tests/e2e/map/map_shapefile.spec.js
```diff
@@ -361,6 +361,20 @@ test.describe('Map: Shapefile interactions', () => {
...
+            x: 1200,
+            y: 448
+          }
+        })
+
+        // Should show only 1 shape selected now
+        await expect(page.getByTestId('filter-stack-item__hint')).toHaveText('1 shape selected')
+
+        // URL should only contain the upper polygon now
+        await expect(page).toHaveURL(/search\?polygon\[0\]=44.1875%2C0.40647%2C58.25%2C-14.46517%2C58.25%2C0.40647%2C44.1875%2C0.40647&sf=1&sfs\[0\]=1&lat=-[8-9]\.\d+&long=46\.\d+&zoom=[3-4]\.\d+/)
```
macrouch: Don't do `zoom=[3-4]`

tests/e2e/map/map_shapefile.spec.js
```diff
@@ -587,5 +601,186 @@ test.describe('Map: Shapefile interactions', () => {
...
+        await page.waitForURL(/zoom=11/, { timeout: 3000 })
+
+        // Draws the spatial on the map
+        await expect(page).toHaveScreenshot('multipoint.png', {
+          clip: screenshotClip,
+          maxDiffPixelRatio: 0.005
+        })
+
+        // Updates the URL and expect all points in the URL
+        await expect(page).toHaveURL(/search\?sp\[0\]=-109.6%2C38.81&sp\[1\]=-109.55%2C38.75&sp\[2\]=-109.5%2C38.7&sf=1&sfs\[0\]=0&lat=38\.\d+&long=-109\.\d+&zoom=1[0-2]\.\d+/)
```
macrouch: Don't do `zoom=1[0-2]`

tests/e2e/map/map_shapefile.spec.js
```diff
@@ -587,5 +601,186 @@ test.describe('Map: Shapefile interactions', () => {
...
+        // Waiting for the URL to include the correct zoom level ensures the map is finished drawing
+        await page.waitForURL(/zoom=11/, { timeout: 3000 })
+
+        // Draws the spatial on the map
+        await expect(page).toHaveScreenshot('multilinestring.png', {
+          clip: screenshotClip
+        })
+
+        // Updates the URL and expect both lines in the URL
+        await expect(page).toHaveURL(/search\?line\[0\]=-109.6%2C38.81%2C-109.62%2C38.83%2C-109.64%2C38.85&line\[1\]=-109.55%2C38.75%2C-109.57%2C38.77%2C-109.59%2C38.79&sf=1&sfs\[0\]=0&lat=38\.\d+&long=-109\.\d+&zoom=1[0-2]\.\d+/)
```
macrouch: Don't do `zoom=1[0-2]`

static/src/js/util/map/getQueryFromShapefileFeature.ts
```diff
@@ -4,8 +4,36 @@ import { Point } from 'ol/geom'
...
+    return (coordinates as number[][]).map((coord: number[]) => coord.slice(0, 2))
+  }
+
+  // Handle Point
+  return (coordinates as number[]).slice(0, 2)
+}
+
 // Get a CMR spatial query from the given feature
-const getQueryFromShapefileFeature = (feature: Feature) => {
+const getQueryFromShapefileFeature = (feature: Feature): SpatialQuery => {
```
macrouch: The `Spatial` type in sharedTypes.ts works here

static/src/js/util/map/__tests__/getQueryFromShapefileFeature.test.ts
```diff
@@ -0,0 +1,73 @@
...
+        ],
+        circleGeometry: null
+      })
+    } as unknown as Feature
+
+    const result = getQueryFromShapefileFeature(mockFeature)
+
+    expect(result).toEqual({
+      polygon: [
+        '-77.0163,38.883,100,-77.0164,38.884,101,-77.0165,38.885,102,-77.0163,38.883,100',
```
macrouch: Isn't this supposed to be removing the altitude?

static/src/js/util/map/interactions/findShapefileFeature.ts
```diff
@@ -34,7 +34,11 @@ const findShapefileFeature = ({
 
   // If the feature is a point or line, we need to check if the coordinate is close
   // to the feature, not within it
-  const closeEnoughTypes = [spatialTypes.POINT, spatialTypes.LINE_STRING]
+  const closeEnoughTypes = [
+    spatialTypes.POINT,
+    spatialTypes.MULTI_POINT,
+    spatialTypes.LINE_STRING
```
macrouch:
```
  const closeEnoughTypes = [
    spatialTypes.POINT,
    spatialTypes.MULTI_POINT,
    spatialTypes.LINE_STRING,
    spatialTypes.MULTI_LINE_STRING
  ]
```

tests/e2e/map/map_shapefile.spec.js
macrouch: You still have duplicate logic in both of these tests, everything up to clicking on the second shape has to happen in both tests, so pull all of it into the beforeEach

tests/e2e/map/map_shapefile.spec.js
```diff
@@ -587,5 +657,219 @@ test.describe('Map: Shapefile interactions', () => {
...
+        await expect(page).toHaveScreenshot('multipoint.png', {
+          clip: screenshotClip,
+          maxDiffPixelRatio: 0.005
+        })
+
+        // Updates the URL and expect all points in the URL
+        await expect(page).toHaveURL(/search\?sp\[0\]=-109.6%2C38.81&sp\[1\]=-109.55%2C38.75&sp\[2\]=-109.5%2C38.7&sf=1&sfs\[0\]=0&lat=38\.\d+&long=-109\.\d+/)
+      })
+
+      test('renders selected multipoint correctly @screenshot', async ({ page }) => {
```
macrouch: This seems to be the same test as above. Did you mean to test the deselected multipoint? If so you should do the same beforeEach as above and pull out all the shared logic so you don't have to repeat code

tests/e2e/map/map_shapefile.spec.js-snapshots/chromium/multiple-shapes-deselected.png
macrouch, line 1: This screenshot has been zoomed in, you don't want that to happen when clicking on the shapes in the test

static/src/js/util/map/getQueryFromShapefileFeature.ts
```diff
@@ -2,10 +2,33 @@ import { Feature } from 'ol'
 import { Point } from 'ol/geom'
 
 import spatialTypes from '../../constants/spatialTypes'
-import { SpatialQueryType } from '../../types/sharedTypes'
+import { Spatial, SpatialQueryType } from '../../types/sharedTypes'
+
+// Remove altitude from coordinate arrays for different types
+const removeAltitudeFromCoordinates = (
+  coordinates: number[] | number[][] | number[][][]
```
trevorlang: If I am reading this right, each of these types represent a Point or MultiPoint, Polygon or LineString, and a MultiPolygon. You could maybe make types for each of those sets, then create a Coordinate type that `or`ed all of them together. That Coordinate type could then be used as the input and output from this function to make things a little easier to grok. You might even be able to use the types that already exist in sharedTypes.ts.

static/src/js/util/map/getQueryFromShapefileFeature.ts
```diff
@@ -2,10 +2,33 @@ import { Feature } from 'ol'
...
+): number[] | number[][] | number[][][] => {
+  if (!coordinates) return coordinates
```
trevorlang: Rather than baking this logic into this function, I think you could make the coordinates param required to ensure that its only called on valid coordinates. I think your `hasAltitude` check in the `if` on line 65 is already acting as a guard that will prevent `coordinates` from being undefined, since it's already checking that the altitude exists.

static/src/js/util/map/getQueryFromShapefileFeature.ts
```diff
@@ -2,10 +2,33 @@ import { Feature } from 'ol'
...
+
+  // Handle different coordinate structures
+  if (Array.isArray(coordinates[0])) {
```
trevorlang: If the altitude is defined, do the coordinates always have [x, y, altitude]? If thats the case, I think we might be able to find a more concise way to recursively trim off the 3rd item in the array.

static/src/js/util/map/getQueryFromShapefileFeature.ts
```diff
@@ -16,19 +39,60 @@ const getQueryFromShapefileFeature = (feature: Feature) => {
...
-      (coords: number[][]) => coords.map(
-        (coord: number[]) => coord.slice(0, 2)
-      )
-    )
+    geographicCoordinatesWithoutAltitude = removeAltitudeFromCoordinates(geographicCoordinates)
+  }
+
+  // Handle multi-geometry types
+  if (geometryType === spatialTypes.MULTI_POLYGON) {
+    const polygons = geographicCoordinatesWithoutAltitude.map((polygon: number[][][]) => polygon[0].flat().join(','))
```
trevorlang: The types defined for the different spatial formats could be used as arguments to these `map` functions

static/src/js/util/map/interactions/onClickShapefile.ts
```diff
@@ -112,14 +112,18 @@ const onClickShapefile = ({
     // Add the feature to the selectedFeatures
     selectedFeatures = updatedSelectedFeatures.concat(edscId)
   } else {
-    // Remove the feature from the existing spatial query
-    const queryByType = [...currentQuery[queryType] || []]
-    const queryIndex = queryByType.indexOf(queryValue)
-    queryByType.splice(queryIndex!, 1)
+    // Remove all queryValues from the existing spatial query (for multi-geometries)
+    const currentQueryValue = currentQuery[queryType] ? [...currentQuery[queryType]] : []
```
trevorlang: Does the spread `[...currentQuery[queryType]]` need to happen? There might be a better way to default the currentQuery values to empty arrays so you could avoid this spread. It just makes this ternary a little hard to understand.

dpesall: We need the spread to make a copy of the array because we call splice() later to remove items and we don't want to accidentally modify the original `spatialSearch` state.

But you make a good point on the ternary here being a bit confusing. I've updated the `currentQuery` so we can just do `const currentQueryValue = [...currentQuery[queryType]]` here.

tests/e2e/map/map_shapefile.spec.js
```diff
@@ -92,7 +92,7 @@ test.describe('Map: Shapefile interactions', () => {
         })
 
         // Updates the URL
-        await expect(page).toHaveURL(/search\?polygon\[0\]=42.1875%2C-2.40647%2C42.1875%2C-16.46517%2C56.25%2C-16.46517%2C42.1875%2C-2.40647&sf=1&sfs\[0\]=0&lat=-9\.\d+&long=49\.\d+&zoom=4\.\d+/)
+        await expect(page).toHaveURL(/search\?polygon\[0\]=42.1875%2C-2.40647%2C42.1875%2C-16.46517%2C56.25%2C-16.46517%2C42.1875%2C-2.40647&sf=1&sfs\[0\]=0&lat=-9\.\d+&long=49\.\d+/)
```
macrouch: Missing the zoom parameter

tests/e2e/map/map_shapefile.spec.js
```diff
@@ -587,5 +636,202 @@ test.describe('Map: Shapefile interactions', () => {
...
+        await expect(page.getByTestId('filter-stack-item__hint')).toHaveText('1 shape selected')
+
+        // Checking that the right number of results are loaded ensures that the route
+        // was fulfilled correctly with the successful paramCheck
+        await expect(page.getByText('Showing 2 of 2 matching collections')).toBeVisible()
+      })
+
+      test('renders correctly and filters collections @screenshot', async ({ page }) => {
+        // Wait for URL to contain spatial parameters
+        await expect(page).toHaveURL(/search\?sp\[0\]=-109.6%2C38.81&sp\[1\]=-109.55%2C38.75&sp\[2\]=-109.5%2C38.7&sf=1&sfs\[0\]=0&lat=38\.\d+&long=-109\.\d+/)
```
macrouch: Missing the zoom parameter

tests/e2e/map/map_shapefile.spec.js
```diff
@@ -587,5 +636,202 @@ test.describe('Map: Shapefile interactions', () => {
...
+        // Draws the spatial on the map
+        await expect(page).toHaveScreenshot('multipoint.png', {
+          clip: screenshotClip,
+          maxDiffPixelRatio: 0.005
+        })
+      })
+
+      test('renders selected multipoint correctly @screenshot', async ({ page }) => {
+        // Wait for URL to contain spatial parameters
+        await expect(page).toHaveURL(/search\?sp\[0\]=-109.6%2C38.81&sp\[1\]=-109.55%2C38.75&sp\[2\]=-109.5%2C38.7&sf=1&sfs\[0\]=0&lat=38\.\d+&long=-109\.\d+/)
```
macrouch: Missing the zoom parameter

tests/e2e/map/map_shapefile.spec.js
```diff
@@ -587,5 +636,202 @@ test.describe('Map: Shapefile interactions', () => {
...
+        // Checking that the right number of results are loaded ensures that the route
+        // was fulfilled correctly with the successful paramCheck
+        await expect(page.getByText('Showing 2 of 2 matching collections')).toBeVisible()
+      })
+
+      test('renders correctly and filters collections @screenshot', async ({ page }) => {
+        // Wait for URL to contain spatial parameters
+        await expect(page).toHaveURL(/search\?sp\[0\]=-109.6%2C38.81&sp\[1\]=-109.55%2C38.75&sp\[2\]=-109.5%2C38.7&sf=1&sfs\[0\]=0&lat=38\.\d+&long=-109\.\d+/)
+
+        await page.waitForTimeout(200)
```
macrouch: waitForTimeout shouldn't be used

tests/e2e/map/map_shapefile.spec.js
```diff
@@ -587,5 +636,202 @@ test.describe('Map: Shapefile interactions', () => {
...
+      })
+
+      test('renders selected multipoint correctly @screenshot', async ({ page }) => {
+        // Wait for URL to contain spatial parameters
+        await expect(page).toHaveURL(/search\?sp\[0\]=-109.6%2C38.81&sp\[1\]=-109.55%2C38.75&sp\[2\]=-109.5%2C38.7&sf=1&sfs\[0\]=0&lat=38\.\d+&long=-109\.\d+/)
+
+        // Verify collections are loaded
+        await expect(page.getByText('Showing 2 of 2 matching collections')).toBeVisible()
+
+        await page.waitForTimeout(200)
```
macrouch: waitForTimeout shouldn't be used

tests/e2e/map/map_shapefile.spec.js
```diff
@@ -362,6 +364,53 @@ test.describe('Map: Shapefile interactions', () => {
...
+
+          // Select both polygons
+          await page.locator('.map').click({
+            position: {
+              x: 1200,
+              y: 448
+            }
+          })
+
+          await page.waitForTimeout(300)
```
macrouch: This is the one place waitForTimeout is ok to use. In the test this should have been copy/pasted from there is a comment `// Wait to avoid a double click`, keep that comment as it explains why we are doing something that should raise a red flag.

Actually scratch that, this is duplicated logic between two tests, that should be pulled out into the beforeEach to avoid the duplicated code

tests/e2e/map/map_shapefile.spec.js
```diff
@@ -587,5 +636,202 @@ test.describe('Map: Shapefile interactions', () => {
...
+      test('renders selected multipoint correctly @screenshot', async ({ page }) => {
+        // Wait for URL to contain spatial parameters
+        await expect(page).toHaveURL(/search\?sp\[0\]=-109.6%2C38.81&sp\[1\]=-109.55%2C38.75&sp\[2\]=-109.5%2C38.7&sf=1&sfs\[0\]=0&lat=38\.\d+&long=-109\.\d+/)
+
+        // Verify collections are loaded
+        await expect(page.getByText('Showing 2 of 2 matching collections')).toBeVisible()
+
+        await page.waitForTimeout(200)
+
+        await expect(page).toHaveScreenshot('multipoint-auto-selected.png', {
```
macrouch: This is the exact same screenshot as `multipoint.png`, what is the point of this test?

tests/e2e/map/map_shapefile.spec.js-snapshots/chromium/multiple-shapes-deselected.png
macrouch, line 1: This screenshot is still zoomed in, deselecting the shape on the map should not result in any zoom action being take. Delete this screenshot and when retaking it be sure to compare it to `multiple-shapes.png`

dpesall: I don't understand why this is happening in the playwright test. If I recreate the exact steps of the test, it does not result in any zoom action when deselecting, yet for some reason it does in the playback.

tests/e2e/map/map_shapefile.spec.js
dpesall: The intent was to make sure all the multipoints were selected and not just one of them when the shapefile was uploaded, but I realize now it's redundant.

static/src/js/util/map/getQueryFromShapefileFeature.ts
```diff
@@ -2,10 +2,27 @@ import { Feature } from 'ol'
 import { Point } from 'ol/geom'
 
 import spatialTypes from '../../constants/spatialTypes'
-import { SpatialQueryType } from '../../types/sharedTypes'
+import { Spatial, SpatialQueryType } from '../../types/sharedTypes'
+
+// Coordinate types for spatial types
+type Coordinate = [number, number] | [number, number, number]
```
trevorlang: Maybe move these to sharedTypes.ts? They should have ts style (`/** **/`) comments as well

static/src/js/util/map/getQueryFromShapefileFeature.ts
```diff
@@ -2,10 +2,27 @@ import { Feature } from 'ol'
...
+type LineString = Coordinate[]
+type Polygon = Coordinate[][]
+type MultiPolygon = Coordinate[][][]
+type CoordinateInput = Coordinate | LineString | Polygon | MultiPolygon
+
+// Remove altitude from coordinate arrays recursively
+const removeAltitudeFromCoordinates = (coordinates: CoordinateInput): CoordinateInput => {
```
trevorlang: It's weird to use `CoordinateInput` as the name for the input and output of this function. Maybe `CoordinateGroup` or `Coordinates` might work better?

static/src/js/util/map/getQueryFromShapefileFeature.ts
```diff
@@ -16,19 +33,65 @@ const getQueryFromShapefileFeature = (feature: Feature) => {
...
-    geographicCoordinatesWithoutAltitude = geographicCoordinates.map(
-      (coords: number[][]) => coords.map(
-        (coord: number[]) => coord.slice(0, 2)
-      )
-    )
+    geographicCoordinatesWithoutAltitude = removeAltitudeFromCoordinates(geographicCoordinates)
+  }
+
+  // Transformation functions for different geometry types
+  const transformPolygonToString = (polygon: number[][][]): string => polygon[0].flat().join(',')
```
trevorlang: A few things:

1. You can use the new types for the inputs from these functions
2: Why are we pulling only the first polygon on line 65? Is it possible that there are multiple polygons?
3: I'd like to see these functions commented with JSDoc style comments. I also feel like they might make sense as standalone utils with tests rather than existing in this file.
4: Lastly, I'm tempted not to stress it right now but I wonder if this could be cleaned up with a switch statement or something. Not a huge fan of the repetitive `if` statements when theres seems to be a fairly concrete pattern in each.

static/src/js/util/map/interactions/onClickShapefile.ts
```diff
@@ -85,24 +85,24 @@ const onClickShapefile = ({
...
+    circle: spatialSearch.circleSearch || [],
+    line: spatialSearch.lineSearch || [],
+    point: spatialSearch.pointSearch || [],
+    polygon: spatialSearch.polygonSearch || []
   }
 
   const newQuery = getQueryFromShapefileFeature(feature)
   const [queryType] = Object.keys(newQuery) as SpatialQueryType[]
-  const [queryValue] = newQuery[queryType]
+  const queryValues = newQuery[queryType] || []
```
trevorlang, line 97: Is there any way `newQuery[queryType]` can be undefined now that you have the defaults above?

static/src/js/util/map/interactions/onClickShapefile.ts
```diff
@@ -112,14 +112,18 @@ const onClickShapefile = ({
     // Add the feature to the selectedFeatures
     selectedFeatures = updatedSelectedFeatures.concat(edscId)
   } else {
-    // Remove the feature from the existing spatial query
-    const queryByType = [...currentQuery[queryType] || []]
-    const queryIndex = queryByType.indexOf(queryValue)
-    queryByType.splice(queryIndex!, 1)
+    // Remove all queryValues from the existing spatial query (for multi-geometries)
+    const currentQueryValue = [...currentQuery[queryType]]
+    queryValues.forEach((queryValue) => {
```
trevorlang, line 119: We should always have blank lines between variable declarations and blocks. Same issue exists with lines 118 and 119

trevorlang (changes requested):

macrouch (dismissed):
Approving to remove the requested changes, but we'll need two approvals besides me

trevorlang (approved):

static/src/js/util/map/__tests__/removeAltitudeFromSpatial.test.ts
```diff
@@ -0,0 +1,158 @@
...
+            [-109.57, 38.75],
+            [-109.57, 38.77],
+            [-109.55, 38.77],
+            [-109.55, 38.75]
+          ]
+        ]
+      ])
+    })
+  })
+})
```
mandyparson, line 158: No circle?

static/src/js/util/map/transformSpatialToQuery.ts
```diff
@@ -0,0 +1,73 @@
...
+ * @returns The spatial query object
+ */
+const transformSpatialToQuery = (
+  geometryType: string,
+  spatial: SpatialCoordinates,
+  circleGeometry: Circle | [] = []
+) => {
+  switch (geometryType) {
+    case spatialTypes.CIRCLE: {
+      const [, radius] = circleGeometry
```
mandyparson, line 37: Need a default for radius if circleGeometry = []

static/src/js/util/map/transformSpatialToQuery.ts
```diff
@@ -0,0 +1,73 @@
...
+    case spatialTypes.POLYGON:
+      return {
+        polygon: [transformPolygonToString(spatial as Polygon)]
+      }
+    case spatialTypes.MULTI_POLYGON:
+      return {
+        polygon: (spatial as MultiPolygon).map(transformPolygonToString)
+      }
+    default:
+      return {}
```
mandyparson, line 69: What other scenarios could be here? Wondering if it should default to an error since there are no other spatialTypes

static/src/js/util/map/__tests__/removeAltitudeFromSpatial.test.ts
macrouch: Circle is just a point that has extra properties (a radius), This function is only looking at the actual coordinates, so a circle is the same as a point

static/src/js/util/map/transformSpatialToQuery.ts
macrouch: If this case of the switch is triggered circleGeometry will always have a point and radius. I'm not sure if I can describe that in the TS types

static/src/js/util/map/transformSpatialToQuery.ts
macrouch: I can't think of when this would ever get triggered, but switch states need a default

PR #1944, opened 2025-08-11
EDSC-4514: Fixes issues with spatial queries

macrouch:
# Overview

### What is the feature?

Bug found: If you add a spatial value, then add a different type spatial value, both values remain in the store.

### What is the Solution?

Check the incoming query within `changeQuery` in `createQuerySlice.ts`. If spatial is included, use it, but default to an empty spatial object.
Refactored the `getCollectionsQuerySpatial` to default values to empty arrays. This avoids the need to save empty arrays in the store.

Also found a bug with deselecting a shapefile feature.

Spatial queries

# Testing

Spatial bug
* Add a spatial search to the map. 
* Add a different spatial search to the map
* Ensure only the second query exists

Shapefile bug
* Upload a shapefile with one shape
* Click on the shape to deselect it
* Ensure the spatial query is removed from the URL and CMR query

# Checklist

### DIFF
--- a/static/src/js/components/AccessMethod/__tests__/AccessMethod.test.jsx
+++ b/static/src/js/components/AccessMethod/__tests__/AccessMethod.test.jsx
@@ -1331,7 +1331,7 @@ describe('AccessMethod component', () => {
                 conceptId: 'collectionId'
               },
               selectedAccessMethod: 'harmony0',
-              spatial: emptySpatial
+              spatial: {}
             }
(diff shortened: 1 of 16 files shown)


trevorlang (approved):

eudoroolivares2016 (approved):
- [x] Searching one shape after the other is working as expected
- [x] Selecting features from multi-polygon

PR #1945, opened 2025-08-13
EDSC-4487: Removes focusedGranule from Redux

macrouch:
# Overview

### What is the feature?

Removes focusedGranule from Redux

### What is the Solution?

Adds it to Zustand

Focused granules

# Testing

Regression test of anything that points to a focused granule. Clicking on a granule in the results list or map, viewing granule details.

# Checklist


static/src/js/zustand/slices/createFocusedGranuleSlice.ts
```diff
@@ -0,0 +1,162 @@
...
+      const {
+        hasAllMetadata = false,
+        isOpenSearch = false
+      } = focusedGranuleMetadata
+
+      // If this is an opensearch granule, we've already retrieved everything we can from the search
+      if (isOpenSearch) return
+
+      // If we already have the metadata for the focusedGranule, don't fetch it again
+      if (hasAllMetadata) return
```
eudoroolivares2016, line 66: I see the old actions had it so I mean I think for risk reduction we can keep it but, I don't see how this is true at this part in the code locally ever?

macrouch: Just had a design review with Trevor for my next ticket, this is going to be changing a bit in that effort, so I'm not worried about it here for now

static/src/js/zustand/types.ts
```diff
@@ -145,6 +145,18 @@ export type FocusedCollectionSlice = {
...
 
+export type FocusedGranuleSlice = {
+  /** The Focused Granule Slice of the store */
+  focusedGranule: {
+    /** The currently focused granule */
+    focusedGranule: string | null
+    /** Function to set or remove the focused granule */
+    changeFocusedGranule: (granuleId: string | null) => void
+    /** Function to get the focused granule metadata */
+    getFocusedGranule: () => void
```
trevorlang, line 156: If this is getting the granule, seems like `void` might not be the right return type?

tests/support/isGetFocusedGranuleQuery.js
```diff
@@ -0,0 +1,14 @@
+const variablesMatch = (variables, granuleId) => variables.params.conceptId === granuleId
+
+export const isGetFocusedGranuleQuery = (route, granuleId) => {
```
trevorlang, line 9: Might we worth while to do some jsdoc for these helper functions

static/src/js/zustand/types.ts
macrouch: Nothing is returned from the function, another part of the store receives the data, but no actual return

tests/support/isGetFocusedGranuleQuery.js
macrouch: :doh: Definitely want that

static/src/js/zustand/types.ts
trevorlang: :ill_see_myself_out:

PR #1949, opened 2025-08-28
EDSC-4516: Fix issue where first page of layer was not rendering

eudoroolivares2016:
…ixes the text on the error page

# Overview

### What is the feature?

Minor fix to `MapContainer` the issue is that the first page of a collection would not load the 

### What is the Solution?

Summarize what you changed.

List impacted areas.

# Testing

### Reproduction steps

1. Step 1
2. Step 2...

### Attachments

# Checklist


static/src/js/components/Errors/ErrorBoundary.jsx
```diff
@@ -55,7 +55,7 @@ class ErrorBoundary extends Component {
     if (hasError) {
       return (
         <div className="wrap">
-          <h2 className="h1">
```
trevorlang: Why was this changed?

static/src/css/modules/_error-page.scss
```diff
@@ -18,9 +18,10 @@
...
   }
 
-  h1 {
-    font-weight: 900;
+  h2 {
+    font-weight: 750;
```
trevorlang: Where did you get `750` from? I think we should probably just remove this and use the default font-weight.

static/src/css/modules/_error-page.scss
```diff
@@ -18,9 +18,10 @@
...
     font-size: 2.8em;
+    color: variables.$color__spacesuit-white;
```
trevorlang: We can probably just use a `text-white` class from bootstrap.

static/src/css/modules/_error-page.scss
```diff
@@ -18,9 +18,10 @@
...
   }
 
-  h1 {
-    font-weight: 900;
+  h2 {
```
trevorlang: Does this page load our bootstrap classes? Looking at how this gets imported, I think it does. If so, I think we should just remove this and use the bootstrap `h1` or `h2` and  `text-white` classes.

static/src/js/containers/MapContainer/MapContainer.tsx
```diff
@@ -428,8 +428,10 @@ export const MapContainer: React.FC<MapContainerProps> = (props) => {
   }
 
   // Generate a key based on the nonExcludedGranules and the addedGranuleIds and removedGranuleIds
```
trevorlang: We should probably just change this comment to:
```js
// Generate a key based on the nonExcludedGranules, addedGranuleIds and removedGranuleIds, 
// and gibs tags. `granulesKey` is used to prevent unnecessary rerenders in the Map component.
```

There are comments in Map.tsx around line 956 and below that should be updated as well.

This granulesKey thing seems a little funky. I feel like its probably not idea to generate this buffer on every rerender. Feels like we should be relying more on react state and then some memoization. @macrouch when youre back, I'd like to pick your brain on this a little.

trevorlang (changes requested):

static/src/css/modules/_error-page.scss
eudoroolivares2016: Will move this fix to my layer picker PR

trevorlang (approved):

PR #1954, opened 2025-09-12
EDSC-4432: Order Status Page loading state briefly shows no collections in the order and a broken project link is displayed

mandyparson:
# Overview

### What is the feature?

A skeleton is displayed for the retrieval intro text so that a user is not shown a broken link
A skeleton is displayed for a order while the order statuses are loading

### What is the Solution?

Pulled id# for link from params
Based loading logic off of id from retrieval as opposed to isLoading from retrieval. isLoading is initiated as false which causes a brief lapse of flawed logic.

OrderStatus

# Testing

### Reproduction steps

- **Environment for testing: local
- **Collection to test with: Any

1. Click on a collection and download it's data.
2. Pause the application at the point of retrieving the orderstatus. I've included attachments below of what it looked like before and after. 

### Attachments
Before
Screenshot 2025-09-12 at 1 07 47 PM

After

Screenshot 2025-09-12 at 10 48 29 AM

# Checklist

### DIFF
--- a/static/src/js/components/OrderStatus/OrderStatus.jsx
+++ b/static/src/js/components/OrderStatus/OrderStatus.jsx
@@ -61,7 +61,7 @@ export const OrderStatus = ({
 
   const {
     collections,
-    id,
+    id: loadedId,
     isLoaded,
     isLoading,
     jsondata = {},
@@ -141,6 +141,8 @@ export const OrderStatus = ({
 
   const eeLink = earthdataEnvironment === deployedEnvironment() ? '' : `?ee=${earthdataEnvironment}`
 
+  const shouldShowLoading = paramsId !== loadedId || (isLoading && !isLoaded)
+
   const introduction = (
     <p>
       {'This page will automatically update as your orders are processed. The Download Status page can be accessed later by visiting '}
@@ -185,7 +187,7 @@ export const OrderStatus = ({
             <Well.Heading>Download Status</Well.Heading>
             <Well.Introduction>{introduction}</Well.Introduction>
             {
-              (!id && !isLoaded) && (
+              shouldShowLoading && (
                 <Skeleton
                   className="order-status__item-skeleton"
                   containerStyle={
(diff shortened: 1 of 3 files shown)


static/src/js/components/OrderStatus/OrderStatus.jsx
```diff
@@ -184,7 +185,7 @@ export const OrderStatus = ({
             <Well.Heading>Download Status</Well.Heading>
             <Well.Introduction>{introduction}</Well.Introduction>
             {
-              (isLoading && !isLoaded) && (
+              (!id && !isLoaded) && (
```
macrouch: I think a better way to ensure this skeleton shows up would be defaulting `isLoading` to `true` on line 66.

What you have works, but we have a variable specifically for this in `isLoading`, using `!id` isn't as clear. 

There is also another Skeleton that should be showing up on line 223.

static/src/js/components/OrderStatus/__tests__/OrderStatus.test.jsx
```diff
@@ -30,7 +34,46 @@ beforeEach(() => {
...
 
 describe('OrderStatus component', () => {
-  test('renders itself correctly', () => {
+  test('renders itself correclty upon page load', () => {
+    setup({
+      overrideProps: initalizedRetrievalStatusProps
+    })
+
+    expect(screen.getByText('https://search.earthdata.nasa.gov/downloads/7')).toBeInTheDocument()
+    expect(Skeleton).toHaveBeenCalledTimes(1)
```
macrouch: On any tests where the data has been loaded, add in an assertion to ensure the Skeleton is not being rendered
```
expect(Skeleton).toHaveBeenCalledTimes(0)
```

static/src/js/components/OrderStatus/OrderStatus.jsx
mandyparson: I considered this but was concerned that it was set to false for a specific reason. Will try it out and see if it triggers anything.

macrouch: I think it was undefined because `retrieval` is defaulted to `{}` until the request finishes.

macrouch: After a huddle we found a better way to do this, so I'll resolve this comment

eudoroolivares2016 (approved):

PR #1960, opened 2025-09-18
EDSC-4568 - Updating page titles for SEO/Accessibility

dpesall:
# Overview

### What is the feature?

Updated page titles using `::` or `|` to instead us `-` to better support SEO and Accessibility.

### What is the Solution?

Updated page title field values where we were previously using `::` or `|`

- Page titles

# Testing

### Reproduction steps

- Navigate to EDSC home page
- Navigate to Preferences
- Verify the page title in the tab says "Preferences - Earthdata Search" rather than "Preferences | Earthdata Search"
- Repeat steps for all other sub routes

### Attachments

image

# Checklist

### DIFF
--- a/static/src/js/App.jsx
+++ b/static/src/js/App.jsx
@@ -279,7 +279,7 @@ const App = () => {
         <ToastProvider ref={window.reactToastProvider}>
           <Helmet
             defaultTitle="Earthdata Search"
-            titleTemplate={`${titleEnv} %s | Earthdata Search`}
+            titleTemplate={`${titleEnv} %s - Earthdata Search`}
           >
             <meta name="description" content={description} />
             <meta property="og:type" content="website" />
--- a/static/src/js/containers/PortalContainer/__tests__/PortalContainer.test.jsx
+++ b/static/src/js/containers/PortalContainer/__tests__/PortalContainer.test.jsx
@@ -89,7 +89,7 @@ describe('PortalContainer component', () => {
       }
     })
 
-    await waitFor(() => expect(document.title).toEqual('Earthdata Search :: example Portal'))
+    await waitFor(() => expect(document.title).toEqual('Earthdata Search - example Portal'))
   })
 
   test('updates the url if the url is using a portal path', async () => {
--- a/static/src/js/containers/PortalContainer/PortalContainer.jsx
+++ b/static/src/js/containers/PortalContainer/PortalContainer.jsx
@@ -76,7 +76,7 @@ export const PortalContainer = ({
   const { primary: primaryTitle } = title
 
   let portalTitle = ''
-  if (!isDefaultPortal(portalId)) portalTitle = ` :: ${primaryTitle || startCase(portalId)} Portal`
+  if (!isDefaultPortal(portalId)) portalTitle = ` - ${primaryTitle || startCase(portalId)} Portal`
 
   const { [defaultPortalId]: defaultPortal } = availablePortals
 


tests/e2e/paths/home/home.spec.js
```diff
@@ -17,6 +17,12 @@ import whatIsThisImageGranules from './__mocks__/what-is-this-image-granules.bod
 import whatIsThisImageGranulesHeaders from './__mocks__/what-is-this-image-granules.headers.json'
 import whatIsThisImageGraphQlBody from './__mocks__/what-is-this-image-collections.graphql.body.json'
 
+const expectTitle = async (page, title) => {
```
eudoroolivares2016: Maybe move this to a utility file in tests/support feels like it does not really need it own file but, at least then we aren't writing it twice

tests/e2e/authentication/authentication.spec.js
```diff
@@ -74,5 +82,70 @@ test.describe('Authentication', () => {
     await expect(
       (await page.getByTestId('collection-results-item').all()).length
     ).toEqual(expectedCollectionCount)
+
+    await expectTitle(page, 'Earthdata Search - Earthdata Search')
```
eudoroolivares2016: this one is just a copy since we have the assertion in the search.spec.js

static/src/js/App.jsx
```diff
@@ -77,6 +77,10 @@ const browserRouter = createBrowserRouter([
         path: '/portal/:portalId/*',
         element: <PortalContainer />
       },
+      {
+        path: '/portal/:portalId/*',
```
eudoroolivares2016: wait that does not make sense to me maybe I'm missing something but, this is already in here on line 80 and having this is why we need the test change in app.test.js

I took out both changes on this branch locally and the portal title was still working so I think you can remove both.
image

Only other things is I'd rather a single playwright test instead of adding to a bunch but, I don't feel strongly about it either way.

dpesall: Oops, I didn't hit undo enough times when I was debugging this with Matthew.

dpesall: My only concern with having a standalone playwright test, is we would need to navigate to each endpoint that has a title we want to check which is much slower compared to just checking the page title on tests which are already on that page, which is fairly instant and doesn't slow down the existing tests.

eudoroolivares2016: You can also just remove the `static/src/js/__tests__/App.test.jsx` change via `git checkout origin/main -- static/src/js/__tests__/App.test.jsx` its still passing for me locally now that you removed that change to app.jsx

static/src/js/containers/PortalContainer/__tests__/PortalContainer.test.jsx
macrouch, line 1: Add a new test that mocks a different environment

tests/e2e/authentication/authentication.spec.js
macrouch, line 1: Move these tests to a new file and do an exact string match for the page title.

Leave a big comment block at the top of that new file explaining the tests will only pass when running the production build

tests/e2e/page_titles.spec.js
macrouch, line 1: Put this into a directory

tests/e2e/page_titles.spec.js
```diff
@@ -0,0 +1,149 @@
+import { test, expect } from 'playwright-test-coverage'
+
+import { setupTests } from '../support/setupTests'
+import { login } from '../support/login'
+
+import commonBody from './paths/search/__mocks__/common.body.json'
+import commonHeaders from './paths/search/__mocks__/common.headers.json'
+import graphQlHeaders from './authentication/__mocks__/graphql.headers.json'
+import getSubscriptionsGraphQlBody from './authentication/__mocks__/getSubscriptions.graphql.body.json'
```
macrouch: Copy whatever mocks you need into this new directory. Reusing mocks from unrelated tests can cause maintenance headaches

tests/e2e/page_titles.spec.js
```diff
@@ -0,0 +1,149 @@
...
+
+const expectTitle = async (page, title) => {
```
macrouch: This is unnecessary, just use `await expect(page).toHaveTitle('...')` in each test

tests/e2e/page_titles.spec.js
```diff
@@ -0,0 +1,149 @@
...
+
+  test('displays the default title on the home page', async ({ page }) => {
+    await respondWithEmptyTimeline(page)
+
+    await page.goto('/')
+
+    await expectTitle(page, 'Earthdata Search - Earthdata Search')
+  })
+
+  test.describe('search', () => {
```
macrouch: Tell a story with each of the describe blocks.

'when viewing the search page' 
'when no portal is active'
'when viewing the download history'

macrouch: Missing the titles for a named and unnamed project, download status, and earthdata download redirect

static/src/js/__tests__/App.test.jsx
```diff
@@ -29,6 +30,12 @@ const setup = setupTest({
 })
 
 describe('App component', () => {
+  // AfterEach(() => {
```
macrouch: Remove

macrouch:
To fix the playwright env issue, add the following line to bin/playwright-prepare-ci, on line 9

```
config="`jq '.application.env = $newValue' --arg newValue prod <<< $config`"
```

tests/e2e/download/download.spec.js
```diff
@@ -123,6 +123,8 @@ test.describe('Download spec', () => {
     // Click the Download Data button
     await page.getByTestId('project-download-data').click()
 
+    await page.waitForURL(/\/downloads\/\d+(\?.*)?$/)
```
eudoroolivares2016: Is this just to make it less flaky? I don't recall seeing this one fail much

tests/e2e/page_titles/page_titles.spec.js
```diff
@@ -0,0 +1,290 @@
...
+
+        await page.goto('/subscriptions')
+
+        await expect(page).toHaveTitle('Subscriptions - Earthdata Search')
+      })
+    })
+
+    test.describe('when opening the Earthdata Download redirect', () => {
+      test('shows the Earthdata Download redirect title', async ({ page }) => {
+        await page.addInitScript(() => {
```
eudoroolivares2016, line 298: Comment here would be nice to explain why its being done

tests/e2e/page_titles/page_titles.spec.js
```diff
@@ -0,0 +1,290 @@
...
+    if (request.method() === 'POST') {
+      route.fulfill({
+        body: JSON.stringify(collectionsResponse),
+        headers: collectionsHeaders
+      })
+
+      return
+    }
+
+    route.continue()
```
eudoroolivares2016: do we need this route.continue?

dpesall: Seems to work without it.

tests/e2e/download/download.spec.js
dpesall: Looks like I just missed removing a line when I had a page title test in this file.

tests/e2e/page_titles/page_titles.spec.js
```diff
@@ -0,0 +1,286 @@
...
+
+const respondWithEmptyTimeline = async (page) => {
+  await page.route('**/search/granules/timeline', (route) => {
+    route.fulfill({ body: JSON.stringify([]) })
+  })
+}
+
+const mockSearchCollections = async (page) => {
+  await page.route('**/search/collections.json', (route, request) => {
+    if (request.method() === 'POST') {
```
macrouch: You don't need this check, it will always be a POST

tests/e2e/page_titles/page_titles.spec.js
```diff
@@ -0,0 +1,286 @@
...
+      }
+
+      route.fulfill({ json: response })
+
+      return
+    }
+
+    route.fulfill({ json: {} })
+  })
+}
```
macrouch: Its annoying that we need to handle the projects document, but we don't need all the other checks

```
const mockProjectsIndex = async (page) => {
  await page.route('**/projects', (route, request) => {
    if (request.resourceType() === 'document') {
      route.continue()

      return
    }

    route.fulfill({ json: [] })
  })
}
```

tests/e2e/page_titles/page_titles.spec.js
```diff
@@ -0,0 +1,286 @@
...
+
+        await page.goto('/contact-info')
+
+        await expect(page).toHaveTitle('Contact Information - Earthdata Search')
+      })
+    })
+
+    test.describe('when managing subscriptions', () => {
+      test('shows the subscriptions title', async ({ page }) => {
+        await page.route(/graphql.*\/api/, (route) => {
```
macrouch, line 281:
```suggestion
        await page.route(/graphql/, (route) => {
```

tests/e2e/page_titles/page_titles.spec.js
macrouch, line 1: You are doing the mock routes inconsistently in this file. You start out using helper functions you defined at the top of the file, but then at the bottom you are just doing the mocks directly in the test blocks. To be more consistent with the rest of the test suite just do the mocks in the test blocks

PR #1972, opened 2025-10-03
EDSC-4573 - Fixing swodlr order

dpesall:
# Overview

### What is the feature?

- Fixing the `creation failed` error on the SWODLR download status page.
- This was because we were not sending all the `granule_params` with our request to cmr.

### What is the Solution?

- Send all the `granule_params` for Swodlr orders, not just granule `concept_id`

List impacted areas.

# Testing

### Reproduction steps

- **Collection to test with:** SWOT Level 2 Water Mask Raster Image Data Product, Version C (C2799438271-POCLOUD)

1. Select a couple granules that support SWODLR orders (use `*_011_424_027*` for granule id filtering)
2. Click Download
3. On the Access Method panel, select SWODLR
4. Click Download Data
5. Ensure that the order succeeds

# Checklist

### DIFF
--- a/serverless/src/submitSwodlrOrder/__tests__/retrieveCMRGranules.test.js
+++ b/serverless/src/submitSwodlrOrder/__tests__/retrieveCMRGranules.test.js
@@ -15,7 +15,12 @@ describe('retrieveCMRGranules', () => {
   test('retrieve cmr granules', async () => {
     nock(/cmr/)
       .matchHeader('Authorization', 'Bearer access-token')
-      .post(/search\/granules.json/)
+      .post('/search/granules.json', (body) => {
+        const params = parseQueryString(body)
+        expect(params.collection_concept_id).toBe('C2799438271-POCLOUD')
+
+        return true
+      })
       .reply(200, {
         feed: {
           entry: [{
@@ -62,7 +67,16 @@ describe('retrieveCMRGranules', () => {
   test('retrieves the granules with the added granule params', async () => {
     nock(/cmr/)
       .matchHeader('Authorization', 'Bearer access-token')
-      .post(/search\/granules.json/)
+      .post('/search/granules.json', (body) => {
+        const params = parseQueryString(body)
+        expect(params.concept_id).toEqual(['G2938390910-POCLOUD', 'G2938390924-POCLOUD'])
+        expect(params.echo_collection_id).toBe('C2799438271-POCLOUD')
+        expect(params.page_num).toBe('1')
+        expect(params.sort_key).toBe('-start_date')
+        expect(params.page_size).toBe('2000')
+
+        return true
+      })
       .reply(200, {
         feed: {
           entry: [{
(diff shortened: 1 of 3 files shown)


serverless/src/submitSwodlrOrder/retrieveCMRGranules.js
```diff
@@ -12,11 +12,9 @@ export const retrieveCMRGranules = async ({
   accessToken,
   granuleParams
 }) => {
-  const { concept_id: granuleIds } = granuleParams
-
   const granuleResponse = await axios({
     url: `${getEarthdataConfig(earthdataEnvironment).cmrHost}/search/granules.json`,
-    params: { concept_id: granuleIds },
```
eudoroolivares2016, line 19: I'm a bit confused here because it works when we manually include those granules in the projects but, in the state where we do a "Download All" it does not but, why would this need to change for that first case?

eudoroolivares2016: this does work in SIT. I think the reason this is working is because we are sending CMR a request with the readable granule name and just making it respond back with the granule metadata where as before we just send the concept-ids directly. http://localhost:8080/search/granules?p=C2799438271-POCLOUD&pg[0][v]=f&pg[0][qt]=2025-05-01T23%3A08%3A54.000Z%2C2025-05-02T23%3A09%3A03.000Z&pg[0][gsk]=-start_date&q=C2799438271-POCLOUD&circle[0]=-174.71213%2C65.69531%2C23871&lat=65.24909013341784&long=-172.77596780592683&zoom=6.601967366536458
I tried with non-readable granule-ids being passed and instead used other filters and I didn't get the CMR error either. Most swodlr use-case is using the readable granule-id at some point the API is meant to be updated and folded behind Harmony as a service.

CMR should almost certainly resolve the same granules if we pass these queries that should work
image

serverless/src/submitSwodlrOrder/retrieveCMRGranules.js
```diff
@@ -12,11 +12,9 @@ export const retrieveCMRGranules = async ({
...
+    params: granuleParams,
```
macrouch: A change like this should have broken a test. Because it didn't, that means we aren't testing the parameters we are sending CMR. Whatever tests we might have need to be improved to verify we are sending the correct data to CMR

serverless/src/submitSwodlrOrder/__tests__/retrieveCMRGranules.test.js
```diff
@@ -131,4 +132,39 @@ describe('retrieveCMRGranules', () => {
...
+      .post('/search/granules.json', (body) => {
+        const params = parseQueryString(body)
+
+        // Verify params are sent
+        expect(params.echo_collection_id).toBe('C2799438271-POCLOUD')
+        expect(params.readable_granule_name).toEqual(['*_011_424_027*'])
+        expect(params.page_num).toBe('1')
+
+        return true
+      })
```
macrouch: I don't think we need a new test in this file, but this block needs to be added to the existing tests, because right now they aren't really testing anything (you can change the input parameters and they still pass).

PR #1973, opened 2025-10-05
EDSC-4580: Fix granule counts when undoing and exclude granules ensure granule list does not contain duplicates

eudoroolivares2016:
# Overview

### What is the feature?

Fixes and issue where the granule exclusion count was correct but, when undoing that excluded granule the count was not getting incremented again. This also fixes an issue where excluding granules can cause duplicates to appear on the list since it refetches the subsequent page. Edge cases where the granule sort key can be null (displays as "unknown") should also be closed out here since the sort key is being passed everywhere consistently

### What is the Solution?
We need to refetch the granules again if we run the undoExclude granule function to ensure that the counts update again. We need to null out the collecion-id to skip over the logic in `static/src/js/zustand/slices/createGranulesSlice.ts`

We refresh when a user exlcludes the first granule page (ignoring the excluded granules of course) and redo that search similarly on the undoExclude but, with that excluded granule back in

Granule exclusion on the `/granules` page of a collection

# Testing

### Reproduction steps

- **Environment for testing:*Any*
- **Collection to test with:*Any*

1. Ensure that selecting the granule card and excluding the granule-id will result in the total count decrementing by 1
2. Ensure that when we `undo` the exclusion the count increments back

### Attachments

image

# Checklist

### DIFF
--- a/tests/e2e/paths/search/granules/granules.spec.js
+++ b/tests/e2e/paths/search/granules/granules.spec.js
@@ -195,7 +195,7 @@ test.describe('Path /search/granules', () => {
         const request = route.request()
         const body = request.postData()
 
-        expect(body).toBe('echo_collection_id=C1214470488-ASF&options[readable_granule_name][pattern]=true&page_num=1&page_size=20&readable_granule_name[]=S1A_S3_SLC__1SDH_20140615T034444_20140615T034512_001055_00107C_16F1')
+        expect(body).toBe('echo_collection_id=C1214470488-ASF&options[readable_granule_name][pattern]=true&page_num=1&page_size=20&readable_granule_name[]=S1A_S3_SLC__1SDH_20140615T034444_20140615T034512_001055_00107C_16F1&sort_key=-start_date')
 
         route.fulfill({
           body: JSON.stringify(readableGranuleNameGranulesBody),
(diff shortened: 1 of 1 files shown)


trevorlang (approved):

static/src/js/zustand/selectors/__tests__/query.test.ts
```diff
@@ -152,7 +155,7 @@ describe('query selectors', () => {
             byId: {
               collectionId: {
                 granules: {
-                  ...initialGranuleState,
+                  ...initialGranuleQuery,
                   browseOnly: true,
                   pageNum: 2,
                   sortKey: '-start_date'
```
mandyparson, line 161: can remove

static/src/js/zustand/selectors/__tests__/query.test.ts
```diff
@@ -104,7 +107,7 @@ describe('query selectors', () => {
...
```
mandyparson, line 113: can remove

mandyparson (approved):

static/src/js/zustand/selectors/__tests__/query.test.ts
eudoroolivares2016: I'll take a double look in my next ticket. I think that this way its more clear as to what the outgoing query is since its including the value in intialGranuleQuery but, you are right it does pass the unit test without it

PR #1975, opened 2025-10-08
EDSC-4354: Initial changes to optimize thumbnail scaler lambda and show download button on single image modal

william-valencia:
# Overview

### What is the feature?

Optimize thumbnail scaler lambda by passing in the browse image url for collections so the lambda doesnt need to make an extra CMR call.

### What is the Solution?

Extracted the browse image URL and passed it to the call for the browse scaler for both collections and granules.

Earthdata Search
Less calls to CMR

# Testing

### Reproduction steps

Navigate to Earthdata Search. View a collections that have a browse image. Make sure the image shows up properly.
Next view a collection's granules that have browse images. Make sure those browse images show up properly.
View the network tab when inspecting using your browser. Make sure the calls to browse scaler are called with the path scale and the imageSrc passed in ie:

http://localhost:3001/scale?h=85&w=85&imageSrc=https://archive.podaac.earthdata.nasa.gov/podaac-ops-cumulus-public/VIIRS_NPP-STAR-L2P-v2.80/20251015213000-STAR-L2P_GHRSST-SSTsubskin-VIIRS_NPP-ACSPO_V2.80-v02.0-fv01.0.sea_surface_temperature.png

### Attachments

Screenshot 2025-10-15 at 8 51 05 PM

# Checklist

### DIFF
--- a/serverless/src/scaleImage/utils/__tests__/buildResponse.test.js
+++ b/serverless/src/scaleImage/utils/__tests__/buildResponse.test.js
@@ -11,7 +11,8 @@ describe('buildResponse', () => {
       statusCode: 200,
       headers: {
         'Content-Type': 'image/png',
-        'Access-Control-Allow-Origin': '*'
+        'Access-Control-Allow-Origin': '*',
+        'Cache-Control': 'max-age=604800, public'
       },
       body: 'dGVzdC1pbWFnZS1jb250ZW50cw=='
     })
@@ -27,7 +28,8 @@ describe('buildResponse', () => {
       statusCode: 404,
       headers: {
         'Content-Type': 'image/png',
-        'Access-Control-Allow-Origin': '*'
+        'Access-Control-Allow-Origin': '*',
+        'Cache-Control': 'max-age=604800, public'
       },
       body: ''
     })
--- a/serverless/src/scaleImage/utils/buildResponse.js
+++ b/serverless/src/scaleImage/utils/buildResponse.js
@@ -12,7 +12,8 @@ export const buildResponse = (imageBuffer, statusCode = 200) => {
     statusCode,
     headers: {
       'Content-Type': 'image/png',
-      'Access-Control-Allow-Origin': '*'
+      'Access-Control-Allow-Origin': '*',
+      'Cache-Control': 'max-age=604800, public' // Cache for 1 week
     },
     body: base64Image
   }


static/src/js/util/request/granuleRequest.js
```diff
@@ -12,6 +12,9 @@ import { granuleRequestPermittedCmrKeys } from '../../../../../sharedConstants/p
 import { getTemporal } from '../../../../../sharedUtils/edscDate'
 
 import normalizeSpatial from '../map/normalizeSpatial'
+import {
+  getBrowseImageUrlFromConcept
+} from '../../../../../serverless/src/scaleImage/utils/cmr/getBrowseImageUrlFromConcept'
```
macrouch: We don't import `serverless` code from the `static` directory. If this util is still being used within the serverless code it needs to be moved into `sharedUtils`. If it isn't used in serverless anymore it needs to be moved into `static/src/js/util`

serverless/src/scaleImage/handler.js
```diff
@@ -51,13 +40,27 @@ const scaleImage = async (event) => {
       height,
       width
     }
+
+    if (!imageSrc) {
```
macrouch, line 43: Will this ever happen? We shouldn't be making this lambda call at all now that we are using the links from the frontend

william-valencia: throw your own error here

static/src/js/util/request/granuleRequest.js
william-valencia: Moved to static/src/js/util

static/src/js/util/request/granuleRequest.js
```diff
@@ -75,29 +73,12 @@ export default class GranuleRequest extends CmrRequest {
 
       if (id) {
         // Retrieve collection thumbnail if it exists
```
eudoroolivares2016: Can we drop this comment it does not seem like its valid anymore (or really ever was?)

macrouch (approved):
We deployed this to SIT and when playing around I noticed we are sending duplicate requests to the lambda every time you view and collection and then back out to the collection results. To fix this we can add a cache-control header to the response in the lambda. This tells the browser to cache the image and not send a network request 

In `buildResponse.js`, add this header: 
```
      'Cache-Control': 'max-age=604800, public' // Cache for 1 week
```

eudoroolivares2016 (approved):

PR #1979, opened 2025-10-21
EDSC-4575: Improve granule gibs imagery drawing

mandyparson:
# Overview

### What is the feature?

Simplify the gibsData object by passing the time (the only unique value in each granule per gibs layer) and just building them in the static/src/js/util/map/drawGranuleBackgroundsAndImagery.ts file

### What is the Solution?

Removed gibsData from individual granules and send it down as an object keyed by the collectionId. This way, we are building the gibsdata once per collection instead of once per granule. Please note that 'time' is excluded from this and is handled directly in the drawGranuleBackgroundsAndImagery.ts file.

MapContainer
Map
drawGranulesBackgroundsAndImagery
sharedTypes

# Testing

### Reproduction steps

- **Environment for testing: prod
- **Collection to test with: C194001235-LPDAAC_ECS

1. Select the granule in the collection above. The gibsData should render the appropriate data the same as before.
2. Select a granule on any other collection that does not have gibsData. It should be drawn correctly and without issue. 

### Attachments

N/A

# Checklist

### DIFF
--- a/static/src/js/util/map/drawGranuleBackgroundsAndImagery.ts
+++ b/static/src/js/util/map/drawGranuleBackgroundsAndImagery.ts
@@ -146,7 +146,7 @@ const drawGranuleBackgroundsAndImagery = ({
       const gibsLayers = gibsLayersByCollection[collectionId] || []
 
       // If the collection has no GIBS layers, return without drawing imagery
-      if (!gibsLayers || gibsLayers.length === 0) return
+      if (gibsLayers.length === 0) return
 
       // There are a few things we need to keep in mind when drawing the granule imagery:
       // 1. The granule imagery is served by GIBS in tiles, but we only show the imagery that is contained


static/src/js/containers/MapContainer/MapContainer.tsx
```diff
@@ -28,7 +28,8 @@ import { projectionConfigs } from '../../util/map/crs'
 // @ts-expect-error The file does not have types
 import murmurhash3 from '../../util/murmurhash3'
 import hasGibsLayerForProjection from '../../util/hasGibsLayerForProjection'
-import buildGibsData from '../../util/map/buildGibsData'
+
+import projectionCodes from '../../constants/projectionCodes'
```
macrouch: Keep this import where it was, grouped with the other constants

static/src/js/util/map/drawGranuleBackgroundsAndImagery.ts
```diff
@@ -139,7 +143,10 @@ const drawGranuleBackgroundsAndImagery = ({
       // Add the feature to the vector source
       vectorSource.addFeature(backgroundFeature)
 
-      // If the granule has no GIBS data, return without drawing imagery
+      const gibsDataObject = gibsDataByCollection[collectionId] || {}
+      const { layers: gibsData = [] } = gibsDataObject
```
macrouch: `gibsLayers` is a more clear variable name here

static/src/js/util/map/drawGranuleBackgroundsAndImagery.ts
```diff
@@ -296,7 +318,7 @@ const drawGranuleBackgroundsAndImagery = ({
                   }
                 }
               },
-              url: gibsDataItem.url,
+              url: `https://gibs-{a-c}.earthdata.nasa.gov/wmts/${gibsDataObject.projection}/best/wmts.cgi?TIME=${gibsTime}`,
```
macrouch:
```suggestion
              url: `https://gibs-{a-c}.earthdata.nasa.gov/wmts/${projectionCode}/best/wmts.cgi?TIME=${gibsTime}`,
```

We already know what projection the map is in with `projectionCode`, you don't need to pass that value again within your gibs object

static/src/js/util/map/drawGranuleBackgroundsAndImagery.ts
```diff
@@ -219,10 +226,25 @@ const drawGranuleBackgroundsAndImagery = ({
...
+          ? time
+          : time.substring(0, 10)
+
         // Create a cache key for the layer.
-        const cacheKey = `${granuleId}-${gibsDataItem.product}-${gibsDataItem.time}-${projectionCode}`
+        const cacheKey = `${granuleId}-${gibsLayer.product}-${gibsTime}-${projectionCode}`
+
+        let resolution: string
+        if (gibsDataObject.projection === projectionCodes.antarctic) {
+          resolution = `${gibsLayer.antarctic_resolution}`
```
macrouch: This is a little confusing to me, because `gibsLayer.antarctic_resolution` is already a string, so why use expression interpolation?. 

If it is because TS tells you "Type 'string | undefined' is not assignable to type 'string'", then you can use the `!` to tell it you are sure there will be a value. I think that is how we've handled similar situations before

```suggestion
          resolution = gibsLayer.antarctic_resolution!
```

static/src/js/containers/MapContainer/MapContainer.tsx
```diff
@@ -403,6 +404,22 @@ export const MapContainer: React.FC<MapContainerProps> = (props) => {
...
+  const gibsDataByCollection = useMemo(() => {
+    const result: GibsDataByCollection = {}
+
+    // Check if focusedCollectionId exists and layersForProjection has elements
+    if (focusedCollectionId && layersForProjection.length > 0) {
+      result[focusedCollectionId] = {
+        layers: layersForProjection,
+        projection
+      }
+    }
```
macrouch: After seeing where projection was being used, we don't need to pass it like this. So I think this would make more sense as `gibsLayersByCollection`, instead of "data"

```suggestion
  const gibsLayersByCollection = useMemo(() => {
    const result: GibsLayersByCollection = {}

    // Check if focusedCollectionId exists and layersForProjection has elements
    if (focusedCollectionId && layersForProjection.length > 0) {
      result[focusedCollectionId] = layersForProjection
    }
```

static/src/js/util/map/drawGranuleBackgroundsAndImagery.ts
mandyparson: good to know! That was exactly why I had to do that.

tests/e2e/map/map_spatial.spec.js
```diff
@@ -622,10 +622,10 @@ test.describe('Map: Spatial interactions', () => {
           await spatialDropdownMenu.getByRole('button', { name: 'Rectangle' }).click()
 
           // Enter the bounding box values
-          await page.getByTestId('spatial-display_southwest-point').focus()
-          await page.keyboard.type('-9.53964,42.1875')
-          await page.getByTestId('spatial-display_northeast-point').focus()
-          await page.keyboard.type('4.5297,56.25')
+          await page.getByTestId('spatial-display_southwest-point').click()
+          await page.getByTestId('spatial-display_southwest-point').fill('-9.53964,42.1875')
```
eudoroolivares2016: The only issue with the .fill instead of type is that it implies for the test users are going to paste in that value from clipboard instead of type it out kind of think we should stick with the existing. I assume this was to try to see if we could cut out an intermittently failing test

static/src/js/util/map/drawGranuleBackgroundsAndImagery.ts
```diff
@@ -139,8 +143,10 @@ const drawGranuleBackgroundsAndImagery = ({
       // Add the feature to the vector source
       vectorSource.addFeature(backgroundFeature)
 
-      // If the granule has no GIBS data, return without drawing imagery
-      if (!gibsData || gibsData.length === 0) return
+      const gibsLayers = gibsLayersByCollection[collectionId] || []
+
+      // If the collection has no GIBS layers, return without drawing imagery
+      if (!gibsLayers || gibsLayers.length === 0) return
```
dpesall: You can just do:
`if (gibsLayers.length === 0) return`

Since you have `gibsLayers` value fallback to `[]`

dpesall (approved):

PR #1980, opened 2025-10-22
EDSC-4272: Removes site preferences from JWT, adds graphql calls for site preferences

macrouch:
# Overview

### What is the feature?

Removes site preferences from JWT
Adds graphql query `user` and mutation `updatePreferences`
Adds UserContainer to make useQuery call and populate redux/zustand with user data
Updates PreferencesForm to use new mutation
Updates Preferences Slice to User Slice

Loading user preferences
Updating user preferences

# Testing

As a logged in user, update your preferences and ensure they are applied when you use EDSC

# Checklist

### DIFF
--- a/tests/support/setupTests.js
+++ b/tests/support/setupTests.js
@@ -1,8 +1,6 @@
 import fs from 'fs'
 import path from 'path'
 
-import { localStorageKeys } from '../../static/src/js/constants/localStorageKeys'
-
 const getImageFileName = (url) => {
   let filename
   // For the arcgis images, don't return the real image, just mock a single image that will repeat
@@ -91,12 +89,12 @@ export const setupTests = async ({
 }) => {
   // Set the 'dontShowTour' flag in localStorage
   await context.addInitScript((value) => {
-    const previousValue = window.localStorage.getItem(localStorageKeys.dontShowTour)
+    const previousValue = window.localStorage.getItem('dontShowTour')
 
     // If we already provided a value, we don't want to overwrite any changes that have been made
     if (previousValue) return
 
-    window.localStorage.setItem(localStorageKeys.dontShowTour, value)
+    window.localStorage.setItem('dontShowTour', value)
   }, dontShowTour.toString())
 
   // Prevent loading of images and map tiles to speed up tests


static/src/js/zustand/selectors/__tests__/query.test.ts
```diff
@@ -82,7 +82,7 @@ describe('query selectors', () => {
     })
 
     test('returns an empty object when there is no focusedCollection', () => {
-      const granuleQuery = getFocusedCollectionGranuleQuery(useEdscStore.getState())
+      const granuleQuery = getFocusedCollectionGranuleQuery(useEdscStore.getInitialState())
```
macrouch: Do I need this change?

static/src/js/components/Preferences/PreferencesForm.jsx
```diff
@@ -1,38 +1,71 @@
...
+
+  const [formData, setFormData] = useState(sitePreferences)
 
   useEffect(() => {
-    setFormData(preferencesData)
-  }, [preferencesData])
+    setFormData(sitePreferences)
+  }, [sitePreferences])
+
+  const [updatePreferences, { loading }] = useMutation(gql(UPDATE_PREFERENCES))
```
macrouch: `updatePreferencesMutation`

serverless/src/graphQl/permissions/__tests__/index.test.js
```diff
@@ -56,6 +57,7 @@ describe('permissions', () => {
             isValidUser,
             userOwnsProject
           ),
+          updatePreferences: isValidUser,
```
trevorlang, line 60: Does this need a check to prevent users from updating another users preferences?

static/src/js/actions/authToken.js
```diff
@@ -41,6 +41,9 @@ export const logout = () => (dispatch, getState) => {
       // Remove the auth cookie
       remove('authToken')
 
+      // Clear the user information from local storage
+      localStorage.removeItem('edscUser')
```
trevorlang: Minor thing but I think we should maybe stray away from "edsc". I know its common internal language but it feels unnecessarily specific. Any reason we wouldn't just do "user" here.

static/src/js/containers/UserContainer/UserContainer.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const earthdataEnvironment = useEdscStore(getEarthdataEnvironment)
+
+  const navigate = useNavigate()
+
+  const [preferencesLoaded, setPreferencesLoaded] = useState(false)
+
+  // When the page loads, check local storage for user information
+  useEffect(() => {
+    if (authToken) {
+      const localUser = localStorage.getItem('edscUser')
```
trevorlang: Same comment re: "edsc" in previous file

serverless/src/graphQl/permissions/__tests__/index.test.js
macrouch: The user that gets updated in the resolver is determined by the token, not a parameter, so a user can only ever update their own data

static/src/js/actions/authToken.js
macrouch: I think I was thinking of naming a cookie, where you would need to be more specific, but with local storage we don't have to worry about another site using that name. I'll update it.

I'll actually make a new constant for localStorage keys so they aren't just strings in the code

dpesall (approved):

PR #1984, opened 2025-11-12
EDSC-4583: Adding the download button without showing the navigation buttons

william-valencia:
# Overview

### What is the feature?

Adding a download button to the image modal when there is only one image present.

### What is the Solution?

Added the download button to the single expanded image modal and made sure the navigation buttons are not there.

The expanded image modal for a granule with a single browse image.

# Testing

### Reproduction steps

In earthdata search, go to a collection that has single images for its granules such as C2343109950-LPCLOUD.
Focus on one of the granules to have its image pop up. Expand that image.
Ensure the download button is present on the expanded image modal.
It should appear on the lower right of the image.

### Attachments

Screenshot 2025-11-11 at 10 27 29 PM

# Checklist

### DIFF
--- a/static/src/js/components/GranuleResults/__tests__/GranuleResultsFocusedMeta.test.jsx
+++ b/static/src/js/components/GranuleResults/__tests__/GranuleResultsFocusedMeta.test.jsx
@@ -327,7 +327,7 @@ describe('GranuleResultsFocusedMeta component', () => {
           expect(images[1]).toHaveClass('granule-results-focused-meta__thumb--is-active')
           expect(pagination).toBeInTheDocument()
 
-          expect(props.onMetricsBrowseGranuleImage).toBeCalledTimes(1)
+          expect(props.onMetricsBrowseGranuleImage).toHaveBeenCalledTimes(1)
           expect(props.onMetricsBrowseGranuleImage).toHaveBeenCalledWith({
             modalOpen: false,
             granuleId: 'G-1234-TEST',
@@ -377,8 +377,14 @@ describe('GranuleResultsFocusedMeta component', () => {
           expect(images[0]).toHaveClass('granule-results-focused-meta__thumb--is-active')
           expect(pagination).toBeInTheDocument()
 
-          expect(props.onMetricsBrowseGranuleImage).toBeCalledTimes(2)
-          expect(props.onMetricsBrowseGranuleImage).toHaveBeenCalledWith({
+          expect(props.onMetricsBrowseGranuleImage).toHaveBeenCalledTimes(2)
+          expect(props.onMetricsBrowseGranuleImage).toHaveBeenNthCalledWith(1, {
+            modalOpen: false,
+            granuleId: 'G-1234-TEST',
+            value: 'Next'
+          })
+
+          expect(props.onMetricsBrowseGranuleImage).toHaveBeenNthCalledWith(2, {
             modalOpen: false,
             granuleId: 'G-1234-TEST',
             value: 'Previous'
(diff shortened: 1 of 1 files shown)


static/src/js/components/GranuleResults/GranuleResultsFocusedMeta.jsx
```diff
@@ -408,46 +408,52 @@ const GranuleResultsFocusedMeta = ({
                 }
               </div>
               {
-                (browseThumbnails.length && browseThumbnails.length > 1) && (
+                (browseThumbnails.length) && (
```
mandyparson, line 411: We can refactor this logic for browseThumbnails a bit. Could do something like:     
```
if (!focusedGranuleId || !browseThumbnails.length) return null

const hasBrowseThumbnails = browseThumbnails.length > 0
const hasMultipleThumbnails = browseThumbnails.length > 1
```

static/src/js/components/GranuleResults/GranuleResultsFocusedMeta.jsx
mandyparson: fix this to target="_blank" while here.

static/src/js/components/GranuleResults/__tests__/GranuleResultsFocusedMeta.test.jsx
mandyparson, line 963: Add one more click of the download button in this test to check that the download button downloads the correct href when flipping through the images.

static/src/js/components/GranuleResults/__tests__/GranuleResultsFocusedMeta.test.jsx
```diff
@@ -837,7 +837,45 @@ describe('GranuleResultsFocusedMeta component', () => {
...
+            const expandButton = screen.getByLabelText('Expand browse image')
+            await user.click(expandButton)
+
+            const downloadButton = screen.getByRole('button', { name: 'Download browse image' })
+
+            expect(downloadButton).toHaveAttribute('href', focusedGranuleLink)
+
+            await user.click(downloadButton)
+
+            expect(props.onMetricsBrowseGranuleImage).toBeCalledTimes(2)
```
macrouch:
```suggestion
            expect(props.onMetricsBrowseGranuleImage).toHaveBeenCalledTimes(2)
```

But why is this being called 2 times?

static/src/js/components/GranuleResults/GranuleResultsFocusedMeta.jsx
william-valencia: Checking the length is how it is done in the rest of the code/files. We don't make seperate "has" variables. If we want this, then it should be a separate ticket for all the code in EDSC.

static/src/js/components/GranuleResults/__tests__/GranuleResultsFocusedMeta.test.jsx
william-valencia: fixed

static/src/js/components/GranuleResults/__tests__/GranuleResultsFocusedMeta.test.jsx
william-valencia: onMetricsBrowseGranuleImage gets called per click. So it gets called during expand and during download.

mandyparson (approved):

static/src/js/components/GranuleResults/__tests__/GranuleResultsFocusedMeta.test.jsx
macrouch: `toBeCalledTimes` is deprecated, it needs to be switched to `toHaveBeenCalledTimes`

static/src/js/components/GranuleResults/__tests__/GranuleResultsFocusedMeta.test.jsx
```diff
@@ -837,7 +837,45 @@ describe('GranuleResultsFocusedMeta component', () => {
...
+            expect(props.onMetricsBrowseGranuleImage).toHaveBeenCalledWith({
```
macrouch: If we know this is called two times we need to check that each time it is called with the expected values. You can do this with `.toHaveBeenNthCalledWith(1, { ... })` and `.toHaveBeenNthCalledWith(2, { ... })`

william-valencia: The "expand" call was tested in another test earlier so maybe thats why they didn't add the expand call to this test(duplicated from another test). I can add the extra expand checks.

macrouch: Thanks for adding those in, it is now much more clear what is happening

PR #1994, opened 2025-11-26
EDSC-4603: Moves exportSearch to frontend

macrouch:
# Overview

### What is the feature?

- Moves exportSearch to frontend
- Deletes lambda
- Deletes Redux code

Export collection search results

# Testing

Load the /search page, export results in CSV and JSON.
Try logged out and logged in and ensure results are correct

# Checklist

### DIFF
--- a/static/src/js/hooks/useExportCollections.ts
+++ b/static/src/js/hooks/useExportCollections.ts
@@ -29,7 +29,7 @@ export const useExportCollections = () => {
   // The collection parameters used in the export query
   const [exportParams, setExportParams] = useState({})
 
-  const [loading, setLoading] = useState<false | 'csv' | 'json'>(false)
+  const [loading, setLoading] = useState<boolean>(false)
 
   const [collections, setCollections] = useState<CollectionMetadata[]>([])
 
@@ -63,14 +63,14 @@ export const useExportCollections = () => {
     // If there is no `data`, return early. (`data` is reset when the next request starts)
     if (isEmpty(data)) return
 
-    const { collections: collectionsData = {} } = data || {}
+    const { collections: collectionsData = {} } = data
     const {
       cursor,
       items = []
-    } = collectionsData
+    } = collectionsData || {}
 
     // If there are items in the current page, add them to the collections state and fetch the next page
-    if (items && items.length > 0) {
+    if (items.length > 0) {
       setCollections((previousCollections) => [
         ...previousCollections,
         ...items
@@ -109,7 +109,7 @@ export const useExportCollections = () => {
     if (loading !== false) return
 
     // Set the loading to the current export type
-    setLoading(type)
+    setLoading(true)
 
     // Set the export format so we remember which format is being exported
     setExportFormat(type)
@@ -151,6 +151,7 @@ export const useExportCollections = () => {
 
   return {
     exportCollections,
+    exportFormat,
     exportLoading: loading
   }
 }
--- a/static/src/js/components/Panels/PanelGroup.jsx
+++ b/static/src/js/components/Panels/PanelGroup.jsx
@@ -47,7 +47,6 @@ export const PanelGroup = ({
   headerMetaPrimaryLoading = false,
   headerMetaPrimaryText = null,
   isActive = false,
-  isExportRunning = false,
   isOpen = false,
   moreActionsDropdownItems = [],
   onChangePanel = null,
@@ -99,7 +98,6 @@ export const PanelGroup = ({
         headerMessage={headerMessage}
         headerMetaPrimaryLoading={headerMetaPrimaryLoading}
         headerMetaPrimaryText={headerMetaPrimaryText}
-        isExportRunning={isExportRunning}
         moreActionsDropdownItems={moreActionsDropdownItems}
         panelGroupId={panelGroupId}
         panelGroupIsActive={isActive}
@@ -131,7 +129,6 @@ PanelGroup.propTypes = {
   headerMetaPrimaryLoading: PropTypes.bool,
   headerMetaPrimaryText: PropTypes.string,
   isActive: PropTypes.bool,
-  isExportRunning: PropTypes.bool,
   isOpen: PropTypes.bool,
   onChangePanel: PropTypes.func,
   moreActionsDropdownItems: PropTypes.arrayOf(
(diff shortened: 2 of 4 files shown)


static/src/js/components/SearchPanels/__tests__/SearchPanels.test.jsx
```diff
@@ -61,16 +68,11 @@ const setup = setupTest({
   defaultPropsByRoute: {
     [PAGE_ROUTE]: {
       collectionSubscriptions: [],
-      isExportRunning: {
```
mandyparson, line 64: Not really part of this, but I did notice this is present in PanelGroup but is not being used in the child component it sends it as a prop to

mandyparson, line 64: Not technically part of this, but this is present in PanelGroup and isn't being used

static/src/js/hooks/useExportCollections.ts
```diff
@@ -0,0 +1,156 @@
...
+      })
+    }
+  }, [error])
+
+  // This useEffect is triggered when the `data` from the query is returned
+  useEffect(() => {
+    // If there is no `data`, return early. (`data` is reset when the next request starts)
+    if (isEmpty(data)) return
+
+    const { collections: collectionsData = {} } = data || {}
```
mandyparson: don't need || {} with the if statement above

static/src/js/hooks/useExportCollections.ts
```diff
@@ -0,0 +1,156 @@
...
+    const {
+      cursor,
+      items = []
+    } = collectionsData
+
+    // If there are items in the current page, add them to the collections state and fetch the next page
+    if (items && items.length > 0) {
```
mandyparson: `items` will always exist

static/src/js/hooks/useExportCollections.ts
```diff
@@ -0,0 +1,156 @@
...
+      setCollections((previousCollections) => [
+        ...previousCollections,
+        ...items
+      ])
+
+      // Fetch the next page of collections
+      doExportCollections({
```
mandyparson, line 80: write a test for this functionality?

static/src/js/hooks/useExportCollections.ts
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * Custom hook to export the collection search results.
+ */
+export const useExportCollections = () => {
+  const [exportFormat, setExportFormat] = useState<'csv' | 'json'>('csv')
+
+  // The collection parameters used in the export query
+  const [exportParams, setExportParams] = useState({})
+
+  const [loading, setLoading] = useState<false | 'csv' | 'json'>(false)
```
mandyparson: I'm confused why this isn't just true or false. We already have the format above. Down below, on line 154, wouldn't it make more sense to return loading? What's the benefit of using 'csv' or 'json' in SearchPanels.jsx lines 311 and 317 over true or false?

static/src/js/operations/queries/exportCollections.ts
```diff
@@ -0,0 +1,26 @@
+import { gql } from '@apollo/client'
+
+const EXPORT_COLLECTIONS = gql`
```
mandyparson, line 3: GET_EXPORT_COLLECTIONS, also should have this in the other ticket but it should get GET_REGIONS as well.

static/src/js/hooks/useExportCollections.ts
macrouch: This is covered by the existing tests. The first request returns results so a second request is triggered here which the tests mock as the empty response that ends the loop

static/src/js/hooks/useExportCollections.ts
macrouch: In SearchPanels we only show the spinner for the export that is running, CSV or JSON. So we need to remember which one they clicked. So we have to remember which one they clicked somewhere. Would it make more sense to return the exportFormat so SearchPanels could understand which spinner to show?

Or maybe just show a spinner on both formats, because you can't run an export on both formats at the same time?

static/src/js/operations/queries/exportCollections.ts
macrouch: I don't like that we have "get" prefixed on every query we do. Most of the time it feels like we're sending extra characters for no reason. `REGIONS` matches the operation name `Regions`, which matches the name of the query `regions`. 

We have a couple of operations that use the `collections` query, but for different purposes, (project collections, and export collections). "Collections" is very generic so I like having "ProjectCollections" and "ExportCollections" to make it clear what the purpose for each is, but adding "Get" to either one doesn't add any additional value.

Most of the operations we have are the only ones that hit the query they use (like regions), so I think we should move to match our operation names with the query names, unless we do need some kind of clarification.

static/src/js/hooks/useExportCollections.ts
mandyparson: Interesting. I see why this solution was needed now. I think this is fine but one other thought I'm having is that we `return { exportCollections,
    exportLoading: loading,
    exportFormat }`
    
   Then you could do `nProgress: exportLoading && exportFormat == 'csv'`
The only real benefit I can see for doing that would be to preserve a true boolean type for loading. Not sure if it's worth it.

static/src/js/operations/queries/exportCollections.ts
mandyparson: currently, every query has either a 'get' or an 'admin' in front of it. ie. getProjectCollections. This would be a departure from what we currently have. If we want to move away from that, that's fine. Should we make an improvement ticket for it though?

macrouch: I feel like that is more of a 'change it when you are in the neighborhood' task, not something to write a ticket for. But @eudoroolivares2016 should weigh in on that

eudoroolivares2016: Yea I agree, those kind of tickets are really not going to be prioritized and subject to change as the app evolves so we're best off updating it as we go. Kind of same reason we didn't write ticket to update all the tests to use the setupTest util etc

eudoroolivares2016 (approved):
Testing both; logged in and out; with portals; even worked when including colls without granules ~50k colls

Issue #1995, opened 2025-11-30
Bug report: Unexpected Application Error when browse SWOT product.


Marine-rubbish:
Before contributing to this project, please review [Contribution.md](https://github.com/nasa/earthdata-search/blob/master/CONTRIBUTING.md).

### Description

**Reproduction Steps**
1. Go to https://search.earthdata.nasa.gov/search
2. Search 'SWOT'  (Fig.1)
3. Select the product in red circle (Fig.1)
4. See error (Fig.2)
**Expected behavior**
I want to preview the SWOT product.

**Screenshots**
<img width="1157" height="1195" alt="Image" src="https://github.com/user-attachments/assets/d7f0ba54-f90f-4798-a2aa-e6cc45a69d06" />
Fig.1 

<img width="1244" height="463" alt="Image" src="https://github.com/user-attachments/assets/e850c76b-8dd7-4361-89a8-cf27d9f8d359" />
Fig. 2

**System Information (please complete the following information):**
 - OS: win11
 - Browser edge
 - Version 142.0.3595.94

**error informaton**
```
Unexpected Application Error!
Cannot read properties of null (reading 'items')
TypeError: Cannot read properties of null (reading 'items')
    at lme (https://search.earthdata.nasa.gov/assets/index-CAas_8ke.js:31:79023)
    at p (https://search.earthdata.nasa.gov/assets/index-CAas_8ke.js:47:566)
    at https://search.earthdata.nasa.gov/assets/index-CAas_8ke.js:47:673
    at Object.N [as useSyncExternalStore] (https://search.earthdata.nasa.gov/assets/edscUtils-DlTmo-bC.js:30:19382)
    at Wd.He.useSyncExternalStore (https://search.earthdata.nasa.gov/assets/edscUtils-DlTmo-bC.js:9:6662)
    at D7.Wd.useSyncExternalStoreWithSelector (https://search.earthdata.nasa.gov/assets/index-CAas_8ke.js:47:742)
    at q7 (https://search.earthdata.nasa.gov/assets/index-CAas_8ke.js:47:1323)
    at r (https://search.earthdata.nasa.gov/assets/index-CAas_8ke.js:47:1671)
    at x (https://search.earthdata.nasa.gov/assets/EditSubscriptionModalContainer-Cou_w97M.js:1:446)
    at c (https://search.earthdata.nasa.gov/assets/edscUtils-DlTmo-bC.js:30:17262)
```

eudoroolivares2016:
[...]

PR #1996, opened 2025-12-01
EDSC-4607: Handle subscriptions null value

nirmalkar:
# Overview

### Issue Link
https://github.com/nasa/earthdata-search/issues/1995

### What is the feature?
This update fixes a bug where clicking on any matching collection card redirected the user to the search/granules page, but the page would break.

The issue occurred because `collectionMetadata.subscriptions` could sometimes be **null**, not just undefined.  
Destructuring directly from a null value causes a runtime error and breaks the page.

To resolve this, the fix ensures that:

- `collectionMetadata.subscriptions` is safely handled even when the value is `null`
- A fallback object (`|| {}` or `?? {}`) is used next to subscriptions, (subscriptions || {}).

With this change, the search/granules page loads correctly regardless of whether the `subscriptions` field is null.

Example fix:
Earlier:
```ts
 export const getFocusedCollectionSubscriptions = (
  state: EdscStore
) => {
  const collectionMetadata = getFocusedCollectionMetadata(state)
  const { subscriptions = {} } = collectionMetadata
  const { items = [] } = subscriptions

  return items
}

```

Now:
```ts
 export const getFocusedCollectionSubscriptions = (
  state: EdscStore
) => {
  const collectionMetadata = getFocusedCollectionMetadata(state)
  const { subscriptions = {} } = collectionMetadata
  // `subscriptions` can be `null` in some cases. Use an empty object as a fallback.
  const { items = [] } = subscriptions || {}

  return items
}
```

List impacted areas.
Page: search/granules?p=C2885053880

# Testing

### Reproduction steps
Go to https://search.earthdata.nasa.gov/search
- Click on any listed collections card
- It will take you to `search/granules?p={id}` page
- The page shouldn't break after the code is merged.

#### When the page was breaking

Screenshot 2025-12-01 at 1 55 55 PM

#### After it was fixed

Screenshot 2025-12-01 at 2 14 59 PM

# Checklist

### DIFF
--- a/static/src/js/zustand/selectors/collection.ts
+++ b/static/src/js/zustand/selectors/collection.ts
@@ -33,9 +33,9 @@ export const getFocusedCollectionSubscriptions = (
   state: EdscStore
 ) => {
   const collectionMetadata = getFocusedCollectionMetadata(state)
-  // `collectionMetadata.subscriptions` can be `null` (not just `undefined`). Using a fallback with `|| {}`
-  const subscriptions = collectionMetadata.subscriptions || {}
-  const { items = [] } = subscriptions
+  const { subscriptions = {} } = collectionMetadata
+  // `subscriptions` can be `null` in some cases. Use an empty object as a fallback.
+  const { items = [] } = subscriptions || {}
 
   return items
 }


static/src/js/zustand/selectors/collection.ts
```diff
@@ -33,7 +33,8 @@ export const getFocusedCollectionSubscriptions = (
   state: EdscStore
 ) => {
   const collectionMetadata = getFocusedCollectionMetadata(state)
-  const { subscriptions = {} } = collectionMetadata
+  // `collectionMetadata.subscriptions` can be `null` (not just `undefined`). Using a fallback with `|| {}`
+  const subscriptions = collectionMetadata.subscriptions || {}
   const { items = [] } = subscriptions
```
macrouch:
```suggestion
  const { subscriptions = {} } = collectionMetadata
  const { items = [] } = subscriptions || {}
```

`|| {}` is definitely the way we want to fix this, but to match our style we would still destructure `subscriptions` and use `|| {}` on the next line.

nirmalkar: Hi @macrouch, thanks for the review. I've made the requested changes. Please review when you get a chance.

PR #2002, opened 2025-12-08
EDSC-4404: The temporal selection filter clear button will clear both the startDate and the endDate if selected

mandyparson:
# Overview

### What is the feature?

When using the date-picker the clear button on the calendar will clear both the endDate and startDate instead of just the field that it is associated to i.e. if I am opening the calendar to select a startDate and I hit clear it should only get clear the start-date and not both the startDate and the endDate

### What is the Solution?

Two things:
1) temporal was being set to an empty object with search parameters were not present. Fixed. 
2) We were spreading the temporal variable instead of using the prevTemporal which can result in us using stale data. Fixed.

Few other things that were fixed here as well:
1) datesSelected weren't working as expected so I fixed that.
2) the isRecurring toggle had a minor issue where the end date was default to the current time as opposed to the end of day time. Fixed. 

Temporal Dropdown

# Testing

### Reproduction steps

- **Environment for testing: local
- **Collection to test with: N/A

1. In the temporal dropdown, mess around with the datepicker. Click 'clear', 'today', or select a specific date. The dropdown menu should work as expected.
2. Try again with temporal search parameters in the URL.

### Attachments

N/A

# Checklist

### DIFF
--- a/static/src/js/components/TemporalDisplay/__tests__/TemporalDisplay.test.jsx
+++ b/static/src/js/components/TemporalDisplay/__tests__/TemporalDisplay.test.jsx
@@ -110,7 +110,11 @@ describe('TemporalDisplay component', () => {
     expect(zustandState.query.changeQuery).toHaveBeenCalledTimes(1)
     expect(zustandState.query.changeQuery).toHaveBeenCalledWith({
       collection: {
-        temporal: {}
+        temporal: {
+          startDate: '',
+          endDate: '',
+          isRecurring: false
+        }
       }
     })
   })


tests/e2e/temporal_selection/temporal_selection.spec.js
```diff
@@ -0,0 +1,186 @@
...
+      // Check if the Start Date input is blank
+      const startDateInput = page.getByRole('textbox', { name: 'Start Date' })
+      await expect(startDateInput).toHaveValue('2021-03-04 00:00:00')
+
+      // Check if the End Date input is blank
+      const endDateInput = page.getByRole('textbox', { name: 'End Date' })
+      await expect(endDateInput).toHaveValue('2022-04-02 23:59:59')
+    })
+
+    test('clicking Clear from DatePicker modal clears the date', async ({ page }) => {
```
macrouch: Because there are two functions that handle the clear button within the datepicker (setStartDate and setEndDate) make two different tests that start with both fields populated and then only press one clear button. Then show only the correct field is empty and the other is still populated

macrouch (approved):

static/src/js/components/TemporalDisplay/__tests__/TemporalSelectionDropdown.test.jsx
```diff
@@ -1,883 +0,0 @@
-import React from 'react'
```
eudoroolivares2016, line 1: I assume we're just getting rid of this since the playwright test does this better

mandyparson: Yes. We shouldn't have two test files testing the exact same thing

PR #2010, opened 2025-12-24
EDSC-4605: Removes handleAlert from Redux

macrouch:
# Overview

### What is the feature?

Removes handleAlert from Redux

This alert is only triggered when the user adds more granules than the max cmr page size (2000). This is only when the user individually adds the granules, not when the collection contains more than 2000.

# Testing

Add individual granules, more than 2000 to your project, ensure the alert lambda is called.
For local testing you can override this value in overrideStatic.config.json (application.maxCmrPageSize), set it to 1, then add two granules and you will see the network request

# Checklist

### DIFF
--- a/static/src/js/util/handleAlert.ts
+++ b/static/src/js/util/handleAlert.ts
@@ -1,15 +1,23 @@
 import { v4 as uuidv4 } from 'uuid'
 
-import LoggerRequest from '../util/request/loggerRequest'
-import routerHelper from '../router/router'
+// @ts-expect-error This file does not have types
+import LoggerRequest from './request/loggerRequest'
+import routerHelper, { type Router } from '../router/router'
 
 export const handleAlert = ({
   message,
   action,
   resource,
   requestObject
-}) => () => {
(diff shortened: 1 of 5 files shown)


static/src/js/util/handleAlert.ts
```diff
@@ -1,15 +1,23 @@
...
-  const { location } = routerHelper.router
+}: {
+  message: string,
+  action: string,
+  resource?: string,
+  requestObject?: {
+    requestId: string
+  }
+}) => {
+  const { location } = routerHelper.router?.state || {} as Router['state']
```
mandyparson, line 20: Why router.state and not router.state.search? 

Not sure if it's possible to mock this in a test but it's unclear to me, as is, what the expected location should look like. Tried printing it out on main and it came in as 'undefined'

macrouch: Look at `static/src/js/router/router.ts`, `state` holds `location`, which holds the `pathname` and `search`

PR #2013, opened 2025-12-29
EDSC-4600: Removes urlQuery from Redux

macrouch:
# Overview

### What is the feature?

Removes urlQuery from Redux, adds the contained functions as standalone util functions

Anything that effects the URL, loading params, or updating params

# Testing

Full app regression test, focusing on the URL

# Checklist

### DIFF
--- a/static/src/js/util/url/__tests__/changePath.test.ts
+++ b/static/src/js/util/url/__tests__/changePath.test.ts
@@ -196,18 +196,26 @@ describe('changePath', () => {
           /* eslint-disable no-param-reassign */
           state.collection.getCollectionMetadata = jest.fn()
           state.collections.getCollections = jest.fn()
+          state.granule.getGranuleMetadata = jest.fn()
           /* eslint-enable no-param-reassign */
         })
 
         await changePath(newPath)
 
-        const { collection, collections } = useEdscStore.getState()
+        const {
+          collection,
+          collections,
+          granule
+        } = useEdscStore.getState()
 
         expect(collections.getCollections).toHaveBeenCalledTimes(1)
         expect(collections.getCollections).toHaveBeenCalledWith()
 
         expect(collection.getCollectionMetadata).toHaveBeenCalledTimes(1)
         expect(collection.getCollectionMetadata).toHaveBeenCalledWith()
+
+        expect(granule.getGranuleMetadata).toHaveBeenCalledTimes(1)
+        expect(granule.getGranuleMetadata).toHaveBeenCalledWith()
       })
     })
 
@@ -219,18 +227,26 @@ describe('changePath', () => {
           /* eslint-disable no-param-reassign */
           state.collection.getCollectionMetadata = jest.fn()
           state.collections.getCollections = jest.fn()
+          state.granule.getGranuleMetadata = jest.fn()
           /* eslint-enable no-param-reassign */
         })
 
         await changePath(newPath)
 
-        const { collection, collections } = useEdscStore.getState()
+        const {
+          collection,
+          collections,
+          granule
+        } = useEdscStore.getState()
 
         expect(collections.getCollections).toHaveBeenCalledTimes(1)
         expect(collections.getCollections).toHaveBeenCalledWith()
 
         expect(collection.getCollectionMetadata).toHaveBeenCalledTimes(1)
         expect(collection.getCollectionMetadata).toHaveBeenCalledWith()
+
+        expect(granule.getGranuleMetadata).toHaveBeenCalledTimes(1)
+        expect(granule.getGranuleMetadata).toHaveBeenCalledWith()
       })
     })
 
@@ -242,18 +258,25 @@ describe('changePath', () => {
           /* eslint-disable no-param-reassign */
           state.collection.getCollectionMetadata = jest.fn()
           state.collections.getCollections = jest.fn()
+          state.granule.getGranuleMetadata = jest.fn()
           /* eslint-enable no-param-reassign */
         })
 
         await changePath(newPath)
 
-        const { collection, collections } = useEdscStore.getState()
+        const {
+          collection,
+          collections,
+          granule
+        } = useEdscStore.getState()
 
         expect(collections.getCollections).toHaveBeenCalledTimes(1)
         expect(collections.getCollections).toHaveBeenCalledWith()
 
         expect(collection.getCollectionMetadata).toHaveBeenCalledTimes(1)
         expect(collection.getCollectionMetadata).toHaveBeenCalledWith()
+
+        expect(granule.getGranuleMetadata).toHaveBeenCalledTimes(0)
       })
     })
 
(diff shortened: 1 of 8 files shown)


static/src/js/util/url/changePath.ts
```diff
@@ -0,0 +1,131 @@
...
+export const changePath = async (path = '') => {
+  const zustandState = useEdscStore.getState()
+  const earthdataEnvironment = getEarthdataEnvironment(zustandState)
+
+  const [pathname, queryString] = path.split('?')
+
+  let decodedParams
+
+  // If query string is a projectId, call getProject
+  if (queryString && queryString.indexOf('projectId=') === 0) {
```
mandyparson, line 20: Also need to check for user. If I share a projectId with another user who is not logged in/doesn't have permission, I should receive an error. See: https://search.earthdata.nasa.gov/search?projectId=4309606521 when not logged in

static/src/js/util/url/updateStore.ts
```diff
@@ -0,0 +1,154 @@
...
+  QuerySlice,
+  ShapefileSlice,
+  TimelineSlice
+} from '../../zustand/types'
+
+export type UpdateStoreParams = {
+  /** Facet parameters for CMR Facets */
+  cmrFacets: FacetParamsSlice['facetParams']['cmrFacets'],
+  /** List of deprecated URL parameters found in the URL */
+  deprecatedUrlParams: string[],
```
mandyparson, line 36: How does our app know if a param is deprecated vs. not a valid param?

static/src/js/util/url/__tests__/changePath.test.ts
```diff
@@ -0,0 +1,393 @@
...
+  test('updates the store if there is not a projectId', async () => {
+    useEdscStore.setState((state) => {
+      /* eslint-disable no-param-reassign */
+      state.collections.getCollections = jest.fn()
+      state.savedProject.setProject = jest.fn()
+      state.timeline.getTimeline = jest.fn()
+      /* eslint-enable no-param-reassign */
+    })
+
+    const newPath = '/search?p=C00001-EDSC'
```
mandyparson, line 143: I'm a little hazy about the different between q and p.

When I do http://localhost:8080/search?p=C3299611519-NSIDC_ECS, I get the response you have outlined below. 

But when I do http://localhost:8080/search?q=C3299611519-NSIDC_ECS, it puts the collection id in query.collection.keyword.

I get q automatically when searching. Also, updateStore is only triggered when I refresh the page with q. If I type in 'C3299611519-NSIDC_ECS' into the search box and click 'enter', updateStore is not called.

static/src/js/util/url/__tests__/changePath.test.ts
```diff
@@ -0,0 +1,393 @@
...
+
+        await changePath(newPath)
+
+        const { collection, collections } = useEdscStore.getState()
+
+        expect(collections.getCollections).toHaveBeenCalledTimes(1)
+        expect(collections.getCollections).toHaveBeenCalledWith()
+
+        expect(collection.getCollectionMetadata).toHaveBeenCalledTimes(1)
+        expect(collection.getCollectionMetadata).toHaveBeenCalledWith()
```
mandyparson, line 215: also getGranuleMetadata(), same for portal tests below

static/src/js/util/url/__tests__/changePath.test.ts
```diff
@@ -0,0 +1,393 @@
...
```
mandyparson, line 246: also getGranuleMetadata(), same for portal tests below

static/src/js/util/url/__tests__/changePath.test.ts
```diff
@@ -0,0 +1,393 @@
...
+      })
+    })
+
+    describe('when the path matches subscription list', () => {
```
mandyparson, line 253: When the path matches granule* subscription list. Do we not need to do anything for routes.COLLECTION_SUBSCRIPTIONS?

static/src/js/util/url/__tests__/updateStore.test.ts
```diff
@@ -0,0 +1,504 @@
...
+        interval: 'month'
+      },
+      setQuery: expect.any(Function),
+      getTimeline: expect.any(Function)
+    })
+  })
+
+  describe('on the projects page', () => {
+    test('sets the zustand state', async () => {
+      const params = {
```
mandyparson, line 126: Why have focused Collections/granules on /projects?

static/src/js/components/GranuleResults/GranuleResultsActions.jsx
```diff
@@ -40,7 +40,6 @@ import './GranuleResultsActions.scss'
  * @param {Boolean} props.initialLoading - Flag designating the initial loading state.
  * @param {Boolean} props.isCollectionInProject - Flag designating if the collection is in the project.
  * @param {Function} props.onMetricsAddCollectionProject - Metrics callback for adding a collection to project event.
- * @param {Function} props.onChangePath - Callback to change the path.
  */
```
mandyparson, line 43: There are still a few instances of onChangePath throughout the repo.

static/src/js/util/url/changePath.ts
macrouch: There can't be a check for user here, you are allowed to retrieve any project. That link throwing an error is a bug I will fix right away.

The use case is if I create a useful project I want to share with a colleague, I should be able to send them the URL. We will fetch the project contents, but the API will create a new project for the new user so they don't overwrite the original project.

static/src/js/util/url/updateStore.ts
macrouch: static/src/js/constants/deprecatedURLParameters.js

static/src/js/util/url/__tests__/changePath.test.ts
macrouch: https://wiki.earthdata.nasa.gov/spaces/EDSC/pages/236028372/Earthdata+Search+URL+Parameters

`q` is the free text query the user can type in.
`p` is used for the focused collection and project collection ids

static/src/js/util/url/__tests__/updateStore.test.ts
macrouch: It is very common to have a focused collection when navigating to the project page. If the users clicks the download button while viewing project granules, then that will be a focused collection.

Focused granules might not be as common, but still very valid. If a user has a focused granule when clicking that download button it will be focused in the store/url

static/src/js/util/url/__tests__/changePath.test.ts
macrouch: No, `pathname.includes(routes.SEARCH)` will trigger the collections search, we don't need a single collection metadata or granules on that route because we don't have a focused collection. Apollo Client is fetching the subscriptions on that route

static/src/js/util/url/__tests__/changePath.test.ts
mandyparson: Thanks for the link! Super helpful. Are there any others you have bookmarked that you think I should have?

macrouch: I'll send you some on Slack

static/src/js/util/url/changePath.ts
```diff
@@ -0,0 +1,131 @@
...
+    await zustandProject.getProjectCollections()
+
+    zustandProject.getProjectGranules()
+  }
+
+  const { getTimeline } = timeline
+  getTimeline()
+
+  return null
+}
```
eudoroolivares2016, line 131: Should these be behind the conditionals for the pathname e.g. fetch projects and timeline only on the /granules and/or the projects page?

macrouch: No, if there are project collections we always want to fetch their metadata. And getTimeline has the necessary logic within that function to bail out of the request if all the parameters aren't available.

PR #2015, opened 2026-01-02
EDSC-4011: Reduce the size of the shapefiles table by aging out old records

eudoroolivares2016:
# Overview

### What is the feature?

We have many old shapefiles in the RDS instance which take up several gigs of storage (See Jira ticket for the breakdown of the tables and charts on when they were created). With the NLP feature that was recently introduced, we expect significantly more shapefiles to get created we want to have a way to remove these shapefiles so they don't continue stacking in size on the database

### What is the Solution?

Added a new lamdba function on a cron job that runs once a month and removes old shapefiles (Old being > 1 year). Logic is in place to detect if it is a parent shapefile (A shapefile with multiple shapes that a user clicked on the map which creates a child shapefile ) if it is, we should only delete it if all the children are also older than one year. We need to do this also to avoid getting an error from the foreign key-constraint in the shapefile table

Shapefiles storage. RDS contents

# Testing

**The way I tested this was I pulled down the shapefiles database table put it into my local database and ran the lambda manually**

### Reproduction steps

- **Environment for testing:*local*
- **Collection to test with:*C3292851686-LARC_CLOUD*

1. Create some shapefiles including some parents (To create a child shapefile) get a shapefile with multiple shapes such as `tests/fixtures/shapefiles/testing_parent_multiple_shapes.geojson` and use a Harmony or Echo-order method and then submit the order via using the start-optionals method

http://localhost:8080/project?p=C3292851686-LARC_CLOUD!C3292851686-LARC_CLOUD&pg[1][a]=3406547847!3406547707!LARC_CLOUD&pg[1][v]=t&pg[1][gsk]=-start_date&pg[1][m]=harmony0&pg[1][cd]=f&pg[1][ets]=t&pg[1][ess]=t&polygon[0]=42.1875%2C-2.40647%2C42.1875%2C-16.46517%2C56.25%2C-16.46517%2C42.1875%2C-2.40647&line[0]=31%2C-15%2C36%2C-17%2C45%2C-19%2C50%2C-21&ff=Customizable&sf=0129248321&sfs[0]=0&sfs[1]=2&lat=2.057288875177182&long=51.25021224930892&zoom=3.515188854418673

2. Run the lambda with `npm run invoke-local cleanupOldShapefiles ./tmp/cleanUpOldShapefilesEvent.json` Note the event is just an empty object, you can put it in whatever directory you want of course, tmp is just already gitignored

### Attachments

# Checklist

### DIFF
--- a/serverless/src/cleanupOldShapefiles/__tests__/handler.test.js
+++ b/serverless/src/cleanupOldShapefiles/__tests__/handler.test.js
@@ -54,9 +54,7 @@ describe('cleanupOldShapefiles', () => {
 
     // Even though both use the same date, Knex creates separate parameterized bindings:
     // one for the main query and one for the subquery
-    expect(queries[0].bindings).toHaveLength(2)
-    expect(queries[0].bindings[0].getTime()).toBe(expectedDate.getTime())
-    expect(queries[0].bindings[1].getTime()).toBe(expectedDate.getTime())
+    expect(queries[0].bindings).toEqual([expectedDate, expectedDate])
 
     expect(result).toEqual({
       body: '{"message":"Successfully deleted 5 shapefile(s)","deletedCount":5}',
@@ -66,6 +64,37 @@ describe('cleanupOldShapefiles', () => {
     expect(consoleMock).toHaveBeenCalledTimes(2)
     expect(consoleMock).toHaveBeenNthCalledWith(1, 'Cleaning up shapefiles older than 2023-01-15T10:00:00.000Z')
     expect(consoleMock).toHaveBeenNthCalledWith(2, 'Successfully deleted 5 shapefile(s) 2023-01-15T10:00:00.000Z')
+  })
+
+  test('correctly handles no shapefiles found and logs a message', async () => {
+    const deletedCount = 0
+    const consoleMock = jest.spyOn(console, 'log').mockImplementation()
+
+    dbTracker.on('query', (query) => {
+      query.response(deletedCount)
+    })
+
+    const result = await cleanupOldShapefiles({}, {})
+
+    const { queries } = dbTracker.queries
+
+    expect(queries[0].sql).toEqual('delete from "shapefiles" where "created_at" < $1 and not exists (select 1 from "shapefiles" as "children" where children.parent_shapefile_id = shapefiles.id and "children"."created_at" >= $2)')
+
+    // Verify bindings are Date objects with the correct date (one year ago)
+    const expectedDate = new Date('2023-01-15T10:00:00.000Z')
+
+    // Even though both use the same date, Knex creates separate parameterized bindings:
+    // one for the main query and one for the subquery
+    expect(queries[0].bindings).toEqual([expectedDate, expectedDate])
+
+    expect(result).toEqual({
+      body: '{"message":"Successfully deleted 0 shapefile(s)","deletedCount":0}',
+      statusCode: 200
+    })
+
+    expect(consoleMock).toHaveBeenCalledTimes(2)
+    expect(consoleMock).toHaveBeenNthCalledWith(1, 'Cleaning up shapefiles older than 2023-01-15T10:00:00.000Z')
+    expect(consoleMock).toHaveBeenNthCalledWith(2, 'No shapefiles older than 2023-01-15T10:00:00.000Z found exiting cleanup process')
   })
 
   test('correctly handles database errors and logs them', async () => {
@@ -90,9 +119,7 @@ describe('cleanupOldShapefiles', () => {
 
     // Even though both use the same date, Knex creates separate parameterized bindings:
     // one for the main query and one for the subquery
-    expect(queries[0].bindings).toHaveLength(2)
-    expect(queries[0].bindings[0].getTime()).toBe(expectedDate.getTime())
-    expect(queries[0].bindings[1].getTime()).toBe(expectedDate.getTime())
+    expect(queries[0].bindings).toEqual([expectedDate, expectedDate])
 
     // Verify error was logged
     expect(consoleMock).toHaveBeenCalledWith(
--- a/serverless/src/cleanupOldShapefiles/handler.js
+++ b/serverless/src/cleanupOldShapefiles/handler.js
@@ -42,7 +42,9 @@ const cleanupOldShapefiles = async (event, context) => {
       })
       .delete()
 
-    console.log(`Successfully deleted ${deletedCount} shapefile(s) ${oneYearAgo.toISOString()}`)
+    console.log(deletedCount > 0
+      ? `Successfully deleted ${deletedCount} shapefile(s) ${oneYearAgo.toISOString()}`
+      : `No shapefiles older than ${oneYearAgo.toISOString()} found exiting cleanup process`)
 
     return {
       statusCode: 200,


serverless/src/cleanupOldShapefiles/handler.js
```diff
@@ -0,0 +1,62 @@
...
+      statusCode: 200,
+      body: JSON.stringify({
+        message: `Successfully deleted ${deletedCount} shapefile(s)`,
+        deletedCount
+      })
+    }
+  } catch (error) {
+    console.log('Error cleaning up old shapefiles:', error)
+
+    return parseError(error)
```
macrouch: You don't need a console log and parseError together. parseError will log the error. But I'm not sure we need parseError at all here. Does this lambda need to return anything? Or is just logging errors ok?

serverless/src/cleanupOldShapefiles/__tests__/handler.test.js
```diff
@@ -0,0 +1,204 @@
...
+
+    expect(result.statusCode).toEqual(200)
+
+    const body = JSON.parse(result.body)
+    expect(body.message).toEqual(`Successfully deleted ${deletedCount} shapefile(s)`)
+    expect(body.deletedCount).toEqual(deletedCount)
+
+    const { queries } = dbTracker.queries
+    expect(queries).toHaveLength(1)
+    expect(queries[0].method).toEqual('del')
```
macrouch:
```
expect(queries[0].sql).toEqual('<full sql here>')
```

serverless/src/cleanupOldShapefiles/__tests__/handler.test.js
```diff
@@ -0,0 +1,204 @@
...
+
+describe('cleanupOldShapefiles', () => {
+  test('successfully deletes non-parent shapefiles older than one year', async () => {
+    const deletedCount = 5
+
+    dbTracker.on('query', (query) => {
+      expect(query.sql).toContain('delete from "shapefiles"')
+      expect(query.sql).toContain('where "created_at" < $1')
+      expect(query.sql).toContain('not exists')
+      expect(query.sql).toContain('children.parent_shapefile_id = shapefiles.id')
```
macrouch: This feels excessive, there is only one sql call, just call query.response()

serverless/src/cleanupOldShapefiles/__tests__/handler.test.js
```diff
@@ -0,0 +1,204 @@
...
+
+      // Verify the date is approximately one year ago (allowing for small time differences)
+      const oneYearAgo = new Date('2023-01-15T10:00:00.000Z')
+      const queryDate = new Date(query.bindings[0])
+      const timeDiff = Math.abs(queryDate.getTime() - oneYearAgo.getTime())
+
+      // Allow up to 1 second difference for test execution time
+      expect(timeDiff).toBeLessThan(1000)
```
macrouch: This shouldn't be necessary, text execution time has nothing to do with the value we pass in to the sql call.

serverless/src/cleanupOldShapefiles/__tests__/handler.test.js
```diff
@@ -0,0 +1,204 @@
...
+  dbTracker.install()
+})
+
+afterEach(() => {
+  dbTracker.uninstall()
+  MockDate.reset()
+})
+
+describe('cleanupOldShapefiles', () => {
+  test('successfully deletes non-parent shapefiles older than one year', async () => {
```
macrouch: Add a test for the console log message

serverless/src/cleanupOldShapefiles/handler.js
```diff
@@ -0,0 +1,62 @@
...
+        // If children exist but all are old, subquery returns empty -> parent can be deleted
+        // If any child is new, subquery returns a row -> parent is NOT deleted
+        this.select(1)
+          .from('shapefiles as children')
+          .whereRaw('children.parent_shapefile_id = shapefiles.id')
+          .where('children.created_at', '>=', oneYearAgo)
+      })
+      .delete()
+
+    console.log(`Successfully deleted ${deletedCount} shapefile(s) older than one year`)
```
macrouch: `older than one year` isn't super useful, put in the variable value used in the query to make the console log tell the whole story

serverless/src/cleanupOldShapefiles/__tests__/handler.test.js
macrouch, line 1: There are 2 branches through the code, it all works, or the database throws an error. But you have 8 different tests, most of which aren't actually testing anything because they are just given a different database response. 

You just need one test that provides a successfull database response and one that shows an error. Each test should ensure the full SQL of the query was correct (see previous comment)

serverless/src/cleanupOldShapefiles/handler.js
eudoroolivares2016: It doesn't need a return value here kind of added that for symmetry to `serverless/src/generateColorMaps/handler.js` we do need to ensure we don't throw an error. The cron is creating an eventBridge rule which has some default behavior for unhandled errors. We're catching it and I'm specifically not throwing an error so eventBridge does not register it as a failure and prevent retrying

serverless/src/cleanupOldShapefiles/__tests__/handler.test.js
eudoroolivares2016: Yea I don't remember why I had this I think I saw it in an example on jest but, its not needed

cdk/earthdata-search/lib/earthdata-search-functions.ts
```diff
@@ -649,5 +649,29 @@ export class Functions extends Construct {
...
+      ...defaultLambdaConfig,
+      description: 'Removes shapefile entries that are older than one year',
+      entry: '../../serverless/src/cleanupOldShapefiles/handler.js',
+      functionName: 'cleanupOldShapefiles',
+      functionNamePrefix,
+      schedules: [{
+        enabled: true,
+        schedule: events.Schedule.cron({
+          day: '1',
+          hour: '8',
```
eudoroolivares2016: find new time avoid overlap with colormps

serverless/src/cleanupOldShapefiles/__tests__/handler.test.js
```diff
@@ -0,0 +1,84 @@
...
+
+    dbTracker.on('query', (query) => {
+      query.response(deletedCount)
+    })
+
+    const result = await cleanupOldShapefiles({}, {})
+
+    const { queries } = dbTracker.queries
+
+    expect(queries[0].sql).toEqual('delete from "shapefiles" where "created_at" < $1 and not exists (select 1 from "shapefiles" as "children" where children.parent_shapefile_id = shapefiles.id and "children"."created_at" >= $2)')
```
macrouch, line 50: Also check the bindings, that will ensure that your time calculation is correct

```
expect(queries[0].bindings).toEqual([...])
```

Add the same assertion in the other test too

serverless/src/cleanupOldShapefiles/handler.js
```diff
@@ -0,0 +1,69 @@
...
+        // If children exist but all are old, subquery returns empty -> parent can be deleted
+        // If any child is new, subquery returns a row -> parent is NOT deleted
+        this.select(1)
+          .from('shapefiles as children')
+          .whereRaw('children.parent_shapefile_id = shapefiles.id')
+          .where('children.created_at', '>=', oneYearAgo)
+      })
+      .delete()
+
+    console.log(`Successfully deleted ${deletedCount} shapefile(s) ${oneYearAgo.toISOString()}`)
```
mandyparson: Should account for situations where no files were deleted. This will say Successfully deleted 0 shapefile(s). Might be helpful to console.log instead "No shapefiles older than ${oneYearAgo.toISOString()} found. Exiting cleanup process. . ."

serverless/src/cleanupOldShapefiles/__tests__/handler.test.js
```diff
@@ -48,6 +48,16 @@ describe('cleanupOldShapefiles', () => {
...
     expect(queries[0].sql).toEqual('delete from "shapefiles" where "created_at" < $1 and not exists (select 1 from "shapefiles" as "children" where children.parent_shapefile_id = shapefiles.id and "children"."created_at" >= $2)')
+
+    // Verify bindings are Date objects with the correct date (one year ago)
+    const expectedDate = new Date('2023-01-15T10:00:00.000Z')
+
+    // Even though both use the same date, Knex creates separate parameterized bindings:
+    // one for the main query and one for the subquery
+    expect(queries[0].bindings).toHaveLength(2)
+    expect(queries[0].bindings[0].getTime()).toBe(expectedDate.getTime())
+    expect(queries[0].bindings[1].getTime()).toBe(expectedDate.getTime())
```
macrouch: This works too 

```
    expect(queries[0].bindings).toEqual([expectedDate, expectedDate])
```

macrouch (approved):

PR #2021, opened 2026-01-13
EDSC-4622: Even more Enzyme tests to Testing Library

macrouch:
# Overview

### What is the feature?

Converts Enzyme tests to Testing Library

Tests

# Testing

Run the tests

# Checklist

### DIFF
--- a/static/src/js/components/EDSCModal/__tests__/EDSCModalOverlay.test.jsx
+++ b/static/src/js/components/EDSCModal/__tests__/EDSCModalOverlay.test.jsx
@@ -1,41 +1,33 @@
 import React from 'react'
-import Enzyme, { shallow } from 'enzyme'
-import Adapter from '@cfaester/enzyme-adapter-react-18'
+import { screen } from '@testing-library/react'
+
+import setupTest from '../../../../../../jestConfigs/setupTest'
 
 import EDSCModalOverlay from '../EDSCModalOverlay'
 
-Enzyme.configure({ adapter: new Adapter() })
-
-function setup(overrideProps) {
-  const props = {
-    ...overrideProps
-  }
-
-  const enzymeWrapper = shallow(<EDSCModalOverlay {...props} />)
-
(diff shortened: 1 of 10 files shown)


static/src/js/components/EDSCModal/__tests__/EDSCModalOverlay.test.jsx
```diff
@@ -1,41 +1,33 @@
...
-        children: <>Test</>
+      setup({
+        overrideProps: {
+          children: <div className="test-child">Test</div>
+        }
       })
 
-      expect(enzymeWrapper.type()).toEqual('div')
-      expect(enzymeWrapper.prop('className')).toEqual('edsc-modal-overlay')
+      expect(screen.getByText('Test')).toBeInTheDocument()
```
eudoroolivares2016, line 30: Should we add that classname assertion back in here?

mandyparson (approved):

static/src/js/components/FilterStack/FilterStackItem.jsx
```diff
@@ -19,8 +19,6 @@ const FilterStackItem = ({
   title,
   variant = 'icon'
 }) => {
-  if (!children) return null
-
```
eudoroolivares2016, line 23: I see this used in 3 places and all of them have child components so that seems fine to remove

macrouch: I removed because the prop is required in the prop types

static/src/js/components/EDSCModal/__tests__/EDSCModalOverlay.test.jsx
macrouch: I don't think so. When the logic in the component changes what classes are applied to an element I try to make sure that logic is tested, but we don't use our jest tests to ensure that every classname is correct on every element, it just isn't designed to do that

PR #2025, opened 2026-01-15
DAS-2457: Add OPeNDAP URL to Harmony output formats.

owenlittlejohns:
# Overview

### What is the feature?

This PR adds a new output format to the mapping of Harmony output formats. This will allow end-users to make requests to Harmony (specifically HOSS) that return an OPeNDAP URL, rather than Harmony itself executing a request against OPeNDAP. (See TRT-460 and DAS-2444 for more details)

### What is the Solution?

A new entry has been added to `harmonyFormatMapping` to map the upcoming UMM-S supported format to a MIME type that Harmony will recognise.

Download customisation form (specifically when a Harmony service is selected).

# Testing

Run the tests (my `grep`-fu didn't find anywhere in the `tests` directory that imported from `outputFormatMaps`, so I've not made any updates)

# Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1779,7 +1779,6 @@
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.28.0.tgz",
       "integrity": "sha512-UlLAnTPrFdNGoFtbSXwcGFQBtQZJCNjaN6hQNP3UPvuNXT1i82N26KL3dZeIpNalWywr9IuQuncaAfUaS1g6sQ==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@ampproject/remapping": "^2.2.0",
         "@babel/code-frame": "^7.27.1",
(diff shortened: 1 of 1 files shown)


owenlittlejohns:
@lyonthefrog - making sure you're on the PR and can confirm the new key and value in the mapping match agreed values.

lyonthefrog (commented):
I'm not familiar enough with earthdata-search to reputably confirm these are the required changed, but I can confirm that this is indeed where we (the Data Services team) expected this change to be made, and the output format mapping string is what we decided on. Thanks!

package-lock.json
macrouch, line 1: These changes shouldn't be included. They might have been made because your node or npm version isn't the same as ours, but when I pulled down your branch and ran `npm install` it undid all of these changes

owenlittlejohns: Got you - it's been a hot minute since I updated `npm`, so I think that's exactly the issue. I've reverted the `package-lock.json` changes in the commit I just pushed. Thanks!

PR #2032, opened 2026-03-04
EDSC-4318:As a User, I want the option to see inactive collections in my search results

mandyparson:
# Overview

### What is the feature?

- Datasets that use the new Collection Progress enumerations of 'PLANNED', 'DEPRECATED', 'PREPRINT', or 'INREVIEW', should not be visible in default search results in Earthdata Search. 
- We need to add an Additional Filter to EDSC to allow users that want to see these results the ability to do so. These will be similar to the other additional filters and should be an update to the URL so those users can share those results.
- During discussion at PI planning we decided that this would be a single toggle for now to encapsulate"Show inactive collections" we should refine that message and provide users a tool-tip detailing what this is doing similar to the one we provide for the "customize feature facet" (THOUGH THIS IS AN ADDITIONAL FILTER)

### What is the Solution?

Added a filter called "Show inactive collections"
It has a tooltip that explains what it shows
Selecting that checkbox adds ic=t (inactiveCollections = true) to the query parameter
The query sets a boolean in the code called showInactiveCollections, this boolean is mapped to a cmr parameter called include_non_operational which brings back all collections regardless of collection progress state. 
This query parameter is present when showInactiveCollections is true and it is undefined (aka not sent as 'false') when showInacitveCollections is false. 

Everywhere that we deal with filters (see onlyEosdisCollections or hasGranulesOrCwic)

# Testing

### Reproduction steps

- **Environment for testing: SIT 
- **Collection to test with: N/A

1. Take note of how many collections show up on initial search query. Open network tab and see what is being sent for collection.json payload
2. Click Show inactive collections and look at the collection.json payload again. There should now be a cmr param of include_non_operational : true. There should now be more collections in the search results and ic=t in the url parameters. 
3. Unclick Show inactive collections. It should go back down in result numbers and include_non_operational should not be in the payload anymore.
4. Also updated url params here: https://wiki.earthdata.nasa.gov/pages/viewpage.action?pageId=236028372&spaceKey=EDSC&title=Earthdata%2BSearch%2BURL%2BParameters

### Attachments

Screenshot 2026-03-04 at 11 01 08 AM

# Checklist

### DIFF
--- a/tests/e2e/tour/tour.spec.js
+++ b/tests/e2e/tour/tour.spec.js
@@ -349,7 +349,7 @@ test.describe('When not logged in', () => {
       left: -10,
       top: 136,
       width: 339,
-      height: 760
+      height: 758
     }, 10, 5)
 
     // Step 6: Refine Search by Category


sharedConstants/nonIndexedCmrKeys.js
```diff
@@ -8,6 +8,7 @@ export const collectionRequestNonIndexedCmrKeys = [
   'granule_data_format_h',
   'granule_data_format',
   'horizontal_data_resolution_range',
+  'include_non_operational', // How can I double check this?
```
mandyparson: can take out. Only need for arrays. Array of values we pass to stingify util that ensures that they dont keep those indexes so its valid for cmr.

static/src/js/routes/Search/Search.jsx
```diff
@@ -206,6 +216,30 @@ export const Search = () => {
...
+                                        superseded, or not provided in results
+                                      </Tooltip>
+                                    )
+                                  }
+                                >
+                                  <div>
+                                    <Form.Check
+                                      checked={isInactiveCollectionsChecked}
+                                      id="input__show-inactive"
+                                      data-testid="input_show-inactive"
```
mandyparson: remove

static/src/js/util/request/__tests__/collectionRequest.test.ts
```diff
@@ -94,6 +95,7 @@ describe('CollectionRequest#nonIndexedKeys', () => {
       'granule_data_format_h',
       'granule_data_format',
       'horizontal_data_resolution_range',
+      'include_non_operational',
```
mandyparson: remove

portals/obdaac/config.json
```diff
@@ -36,7 +36,8 @@
     "secondary": "Ocean Biology Distributed Active Archive Center"
   },
   "ui": {
-    "showNonEosdisCheckbox": true,
-    "showOnlyGranulesCheckbox": true
+    "showNonEosdisCheckbox": false,
+    "showOnlyGranulesCheckbox": false,
+    "showInactiveCollectionsCheckbox": true
```
mandyparson: change back to all true

portals/idn/config.json
```diff
@@ -12,6 +12,7 @@
   },
   "ui": {
     "showNonEosdisCheckbox": false,
-    "showOnlyGranulesCheckbox": false
+    "showOnlyGranulesCheckbox": false,
+    "showInactiveCollectionsCheckbox": false
```
mandyparson: change all of these back to true. Double check with DAACs to see if anyone doesn't want it.

static/src/js/routes/Search/Search.jsx
```diff
@@ -206,6 +216,30 @@ export const Search = () => {
...
+                                      label="Show inactive collections"
```
mandyparson: Change to 'Include inactive collections' look into getting an i icon in here.

static/src/js/routes/Search/__tests__/Search.test.jsx
```diff
@@ -146,6 +147,18 @@ describe('Search component', () => {
...
 
+    test('renders the "Include inactive collections" checkbox under PortalFeatureContainer', async () => {
+      const { user } = setup()
+
+      const checkbox = await screen.findByLabelText('Include inactive collections')
+      expect(checkbox).toBeInTheDocument()
+
+      const icons = screen.getAllByRole('graphics-symbol')
+      const tooltip = icons[1]
+      await user.hover(tooltip)
```
eudoroolivares2016, line 158: I tried to find some other more semantic way to find this thing but, I was unable to which didn't also mean ignoring other things

PR #2036, opened 2026-03-13
EDSC-4636: Close vulnerabilities for EDSC

eudoroolivares2016:
# Overview

### What is the feature?

We need to update dependencies for EDSC on `snyk` findings

### What is the Solution?

Updates dependencies using the `npm audit fix` CLI
When updating the `@aws-sdk`, `client-sqs` tool serialization has become more strict as a result a change needed to be made to update a test. Additionally, for local development we use `elastic-mq` which failed to recieve messages for creating retrievals when in use (e.g. when running start-optionals) this turned out to be because the `orderDelay` was being parsed as undefined. Prior versions where able to accomodate for this but, this newer version resulted in errors propogating from elasticmq. The errors can be resolved simply by assigning a value to `ORDER_DELAY_SECONDS` in `bashrc` but, this adds some robustness. In production this would not occur which was seen in testing, as the value is defaulted and set by the `cdk` deployment

I did also try simply pulling the latest `elasticmq` version `[1.6.16]`(https://hub.docker.com/layers/softwaremill/elasticmq/1.6.16/images/sha256-ef07956a1f734e8ce7d0db5cd63d87b3f0ae3d37ae404c60237bf382f0f4f2cb) that this time but, that didn't make a difference

This change occurred between `
3.913.0 << 3.955.0` these versions. Since the `https://github.com/aws/aws-sdk-js-v3` does rolling updates there are many possible versions where this would have broken.

Application dependencies creating retrievals and their default delay value 

# Testing

### Reproduction steps

- **Environment for testing:*SIT*
- **Collection to test with:*Any (with customizations)*

1. Start EDSC local with elastic-mq turned on (You may need to pull it from docker-hub) using `npm run start:optionals`
2. Regression test behavior involving the SQS queue: make a Harmony request or echo-orders.
3. Use the requeue feature on the admin portal
4. Ensure the `snyk` issues are resolved

### Attachments

Not applicable

# Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "migrate": "node-pg-migrate",
     "start:app": "vite dev --port 8080",
     "start": "dotenvx run node bin/start.js",
-    "start:optionals": "USE_IMAGE_CACHE=true SKIP_SQS=false npm start",
+    "start:optionals": "USE_IMAGE_CACHE=true SKIP_SQS=false ORDER_DELAY_SECONDS=1 npm start",
     "copy-secrets": "cp secret.config.json.example secret.config.json",
     "test": "TZ=US/Eastern vitest",
     "test:ui": "TZ=US/Eastern vitest --ui",


serverless/src/graphQl/resolvers/retrieval.js
```diff
@@ -244,8 +244,8 @@ export default {
               // sqs entry by ORDER_DELAY_SECONDS value for each page. i.e. page 1
               // will not wait, page 2 will wait ORDER_DELAY_SECONDS, page 3 will wait
               // ORDER_DELAY_SECONDS times 2, etc.
-              const delay = (pageNum - 1) * parseInt(process.env.ORDER_DELAY_SECONDS, 10)
```
eudoroolivares2016, line 247: We should identify any env vars which differ from the production env

eudoroolivares2016: I'm gonna put that work into this for next sprint theres some other things we can trim in `[DSC-4633](https://bugs.earthdata.nasa.gov/browse/EDSC-4633)`

PR #2044, opened 2026-05-04
EDSC-4666: Fix `isRecurring` not correctly being encoded and decoded into the url

eudoroolivares2016:
# Overview

### What is the feature?

Resolves and issue with url encoding on the `isRecurring filter` this also solves the issue of the dead code we have in the projects page for isRecurring

This also bumps some dependencies including a minor release of `axios` see notes below

I noticed we had a reference to legacy services which is OBE so this also just removes that from `static.config`

### What is the Solution?

Correctly parse the way that the isRecurring url gets into the URL state and parsed out. Added a comment to explain how that works

This bumps axios by minor release versions we do not use the breaking change url.parse() and v1.14.0 did not identify any breaking issues these are according to their release notes below. I'm not seeing anything on version 13 either

https://github.com/axios/axios/releases/tag/v1.14.0
https://github.com/axios/axios/releases/tag/v1.15.0
https://github.com/axios/axios/releases/tag/v1.13.0

Seems this is already in the documentation https://wiki.earthdata.nasa.gov/spaces/EDSC/pages/236028372/Earthdata+Search+URL+Parameters

url encoding specifically for temporal searching

# Testing

### Reproduction steps

- **Environment for testing:*any*
- **Collection to test with:*C1593392869-LAADS* (Really any that have temporal subsetting)

1. Add a temporal search with isRecurring. Refresh or paste to a new browser that should appear in the search since its being encoded and decoded from the URL
2. Add that collection above with a few granules after doing temporal search using recurring filter. You should see the warning we have about using isrecurring and how its disabled
3. Do some regression testing here for the axios bump in particular

### Attachments

image

# Checklist

- [] New and existing unit tests pass locally with my changes
- [NA] I have made corresponding changes to the documentation

### DIFF
--- a/static/src/js/util/url/temporalEncoders.js
+++ b/static/src/js/util/url/temporalEncoders.js
@@ -15,15 +15,15 @@ export const encodeTemporal = (temporal) => {
   } = temporal
 
   const valuesToEncode = [
-    startDate || '',
-    endDate || ''
+    startDate,
+    endDate
   ]
 
   if (isRecurring) {
-    valuesToEncode.push(...[recurringDayStart || '', recurringDayEnd || ''])
+    valuesToEncode.push(...[recurringDayStart, recurringDayEnd])
   }
 
-  if (valuesToEncode.filter(Boolean).length === 0 && !isRecurring) return undefined
+  if (valuesToEncode.filter(Boolean).length === 0) return undefined
 
   const encodedString = valuesToEncode.join(',')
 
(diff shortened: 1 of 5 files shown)


mandyparson (approved):

static/src/js/util/url/temporalEncoders.js
```diff
@@ -40,14 +40,19 @@ export const decodeTemporal = (string) => {
...
     endDate,
     recurringDayStart = '',
     recurringDayEnd = ''
-  ] = string.split(',')
+  ] = values
 
-  const isRecurring = !!(recurringDayStart && recurringDayEnd)
+  // We check `values.length > 2` because when `encodeTemporal` runs with `isRecurring: true`,
+  // it appends the recurring day start and end values to the array (even if they are empty strings).
+  // This check ensures we preserve the `isRecurring: true` state when decoding the URL, even if the user
```
william-valencia: even if the user... was there supposed to be another line in this comment?

eudoroolivares2016: Fixed

william-valencia (approved):

PR #2060, opened 2026-06-26
EDSC-4692: Cache geocoder results

macrouch:
# Overview

### What is the feature?

Cache geocoder results to avoid duplicate Bedrock calls

NLP Search

# Testing

- In your .env file, ensure `GEOCODE_CACHE_EXPIRE_SECONDS=2592000` is set, and `USE_IMAGE_CACHE` was renamed to `USE_CACHE`
- Run `npm run start:optionals` to run cache
- `curl -N --location 'http://localhost:3001/nlp?query=rainfall%20in%20Maryland'` to fetch results first time
- Run curl again and ensure geocoder lambda is not called, logs should indicate a cache hit and the results should be the same as the first call

# Checklist


bin/start-playwright.sh
```diff
@@ -4,4 +4,5 @@
 SHARD_NUMBER=$1
 
 # Run the Playwright tests with the specified shard number
+# Wait for an extra time after the server is up to ensure that the application is fully loaded before running the tests
```
macrouch: Remove

serverless/src/nlpSearch/handler.js
```diff
@@ -48,8 +51,28 @@ const callPythonLocal = async (query) => {
  * Calls the geocoder lambda to get the spatial area for a given query.
  */
 const getSpatial = async (query) => {
+  const isCacheEnabled = process.env.USE_CACHE === 'true'
+  const cacheKey = `geocoder:${query}`
```
macrouch: Lowercase the query

bin/start-playwright.sh
```diff
@@ -5,4 +5,4 @@ SHARD_NUMBER=$1
 
 # Run the Playwright tests with the specified shard number
 # Wait for an extra time after the server is up to ensure that the application is fully loaded before running the tests
-npx start-server-and-test preview http-get://localhost:8080 'sleep 30; npm run playwright -- --shard='"$SHARD_NUMBER"''
+npx start-server-and-test preview http-get://localhost:8080 'npm run playwright -- --shard='"$SHARD_NUMBER"''
```
stephenkilbourn, line 7: it might be more stable to use playwright's native webserver controls and not the third party start-server-and test.  Playwright pings the baseURL until timeout and only starts the web server after that.  in the playwright config:
```
export default defineConfig({
  // Rest of your config...

  // Run your local dev server before starting the tests
  webServer: {
    command: 'npm run start',
    url: 'http://localhost:8080',
    reuseExistingServer: !process.env.CI,
    timeout: 120 * 1000,
  },
});
```
w
https://playwright.dev/docs/test-webserver#adding-a-server-timeout

macrouch: I ruled out the problem that the server wasn't ready. My current theory is that some cdn served files are not able to be downloaded (no idea why), and that is causing Playwright not to be able to interact with the application, so the test times out.

mandyparson (approved):

serverless/src/util/cache/getItemFromCache.js
```diff
@@ -1,19 +1,19 @@
...
-      if (image) {
+    .then((item) => {
+      if (item) {
         console.log(`Cache HIT '${key}'`)
 
-        return image
+        return item
       }
 
       console.log(`Cache MISS '${key}'`)
```
eudoroolivares2016, line 19:
```suggestion
export const getItemFromCache = async (key) => {
  const client = getCacheConnection()

  const item = await client.get(key)

  if (item) {
    console.log(`Cache HIT '${key}'`)

    return item
  }

  console.log(`Cache MISS '${key}'`)

  return null
}
```

I think this is the same I'll leave it up to you if its worth refactoring

.env.example
```diff
@@ -2,11 +2,12 @@
 # any secret values, so they do not need to be git ignored
 
 AWS_REGION=us-east-1
-CACHE_KEY_EXPIRE_SECONDS=84000
+IMAGE_CACHE_EXPIRE_SECONDS=84000
```
eudoroolivares2016, line 5: I think this might need to be put into bamboo

eudoroolivares2016 (approved):

serverless/src/util/cache/getItemFromCache.js
macrouch: My next ticket will involve minor refactoring around this code, so I'll do it then

.env.example
macrouch: I've updated the necessary deployment variables

PR #2062, opened 2026-06-26
EDSC-4683: Adding e2e playwright tests

mandyparson:
# Overview

### What is the feature?

Since we decided not to move ahead on any design changes, this ticket just zips up remaining playwright tests. 

### What is the Solution?

Most playwright tests were failing because I needed to mock a response from /capabilities since we're hitting that endpoint every time we call getCollections

Playwright tests and few mocks that were setting items: null when they should have been setting items: []

# Testing

### Reproduction steps

1. run playwright tests and make sure they all pass. Regression test as needed.
2. Let me know if additional tests are needed,

### Attachments

# Checklist

### DIFF
--- a/tests/e2e/projects/harmony_with_mbr/harmony_with_mbr.spec.js
+++ b/tests/e2e/projects/harmony_with_mbr/harmony_with_mbr.spec.js
@@ -4,10 +4,11 @@ import { login } from '../../../support/login'
 import { setupTests } from '../../../support/setupTests'
 import { getAuthHeaders } from '../../../support/getAuthHeaders'
 
-import timeline from './__mocks__/timeline.json'
-import granules from './__mocks__/granules.json'
-import shapefile from './__mocks__/shapefile.json'
 import collectionsGraphQlBody from './__mocks__/collections_graphql_body.json'
+import granules from './__mocks__/granules.json'
+import harmonyCapabilitiesDocument from './__mocks__/harmonyCapabilitiesDocument.json'
+import shapefile from './__mocks__/shapefile.json'
+import timeline from './__mocks__/timeline.json'
 
 const screenshotClip = {
   x: 1027,
@@ -16,7 +17,7 @@ const screenshotClip = {
   height: 90
 }
 
-const mbrWarning = 'Only bounding boxes are supported. If this option is enabled, your line will be automatically converted into the bounding box shown above and outlined on the map.'
+const mbrWarning = 'Only bounding boxes are supported. Your line has been automatically converted into the bounding box shown above and outlined on the map.'
 
 test.describe('Harmony with MBR', () => {
   test.beforeEach(async ({ page, context }) => {
@@ -34,6 +35,12 @@ test.describe('Harmony with MBR', () => {
         await page.route(/saved_access_configs/, async (route) => {
           await route.fulfill({
             json: {}
+          })
+        })
+
+        await page.route('**/capabilities**', async (route) => {
+          await route.fulfill({
+            json: harmonyCapabilitiesDocument
           })
         })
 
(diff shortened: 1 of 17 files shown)


tests/e2e/projects/harmony_with_mbr/harmony_with_mbr.spec.js
```diff
@@ -79,11 +86,17 @@ test.describe('Harmony with MBR', () => {
 
           // Wait for the base map to load to avoid bad screenshots
           await tilesPromise
+
+          // Click 'Customize Download' radio button
+          await page.locator('.access-method-radio__radio').first().check()
```
macrouch: The page.goto URL is supposed to be selecting the Harmony access method so you don't have to do additional clicks, change `&pg[1][m]=harmony0` to `&pg[1][m]=harmony` on line 85 and you can remove this action

tests/e2e/projects/harmony_with_mbr/harmony_with_mbr.spec.js
```diff
@@ -99,17 +112,22 @@ test.describe('Harmony with MBR', () => {
 
         test.describe('when deselecting Enable Spatial Subsetting', () => {
           test.beforeEach(async ({ page }) => {
-            await page.getByLabel('Trim output granules to the selected spatial constraint').uncheck()
+            // Click 'Trim output granules' to enable spatial subsetting (ess)
+            await page.getByRole('checkbox', { name: 'Trim output granules to the' }).check()
+
+            // Click 'Trim output granules' to disable spatial subsetting (ess)
+            await page.getByRole('checkbox', { name: 'Trim output granules to the' }).uncheck()
```
macrouch: I think this is a symptom of not loading from the URL correctly, but after I fixed the URL I could remove the `.check()` call and only have to uncheck the box

tests/e2e/projects/harmony_with_mbr/harmony_with_mbr.spec.js
```diff
@@ -126,29 +144,33 @@ test.describe('Harmony with MBR', () => {
 
           // Wait for the base map to load to avoid bad screenshots
           await tilesPromise
+
+          // Click 'Customize Download' radio button
+          await page.locator('.access-method-radio__radio').first().check()
```
macrouch: Same comment as above, fix the page.goto URL and this is unnecessary

tests/e2e/projects/harmony_with_mbr/harmony_with_mbr.spec.js
mandyparson: That was it! Looks like there was only one other file with a remaining harmony0 in it. Fixed.

tests/e2e/projects/harmony_with_mbr/harmony_with_mbr.spec.js
```diff
@@ -75,15 +82,18 @@ test.describe('Harmony with MBR', () => {
...
           await tilesPromise
         })
 
         test('displays a mbr on the map @screenshot', async ({ page }) => {
           // Expect the mbrWarning to be displayed
-          await expect(page.getByRole('alert')).toHaveText(mbrWarning)
+          await expect(page.getByRole('alert').filter({ hasText: 'Only bounding boxes are' })).toHaveText(mbrWarning)
+
+          // Click 'Trim output granules' to enable spatial subsetting (ess)
+          await page.getByRole('checkbox', { name: 'Trim output granules to the' }).check()
```
macrouch: Something is wrong here, the URL specifies `ess=t`, which means the checkbox needs to be checked on page load, but when I take this out the test fails.

I also can't replicate it in my browser, every time I load the page with the `ess=t` in the URL my checkbox is checked like I expect it to be

macrouch: Nothing wrong, I wasn't up to date on your branch. But this does need to be removed

macrouch (approved):

stephenkilbourn (approved):

PR #2067, opened 2026-07-14
EDSC-4676: Integrate Streaming NLP to landing page

stephenkilbourn:
# Overview

### What is the feature?

This PR updates the NLP search flow to stream status updates from the /nlp endpoint and move users to Search only after NLP completion is finalized. It also introduces explicit Home-slice state to scope NLP map auto-centering to NLP-driven navigation only. Status updates are displayed by a dedicated NLP status component, UI lock/cancel behavior is added to the home, and then navigation occurs once a final parsed result is available.

### What is the Solution?

1. Added NLP streaming status component: `NlpSearchStatus.tsx`
2. NLP flow now uses useCompletion for streaming status text from /nlp.
3. Added nlp stream mocking for playwright testing: `tests/support/nlpStreamMock.js` and `tests/support/nlpHandlers.js`
4. Updated the home route to display the NlpSearchStatus during search, Search input is disabled and search button toggles to a cancel button to allow users to interrupt search if erroneous input is entered. 
5. Added explicit Home Zustand state:
- nlpAutoCenterPending
- setNlpAutoCenterPending
6. Added/reset logic for the new flag so it is:
set true only when an NLP result is ready to drive the next map-center action
consumed and reset after use or reset on cancel/failure/reset paths to avoid stale behavior.
7. `drawNlpSpatialData` is (still) currently not wired into this path and remains unused in this PR.

#### Intermediate step handling 

1. Intermediate NLP progress is handled in `NlpSearchStatus.tsx`` using streamed text from `nlp`
2. The component reads the stream text via `useCompletion`
3. A short delay is used before showing intermediate steps so fast responses do not flicker.
4. Progress lines are parsed from stream text prior to the final marker (`Final result:`).

### Why useCompletion (not useObject) for `/nlp`
`useCompletion` was chosen because this interaction is text-stream, not object-first. The `/nlp` stream emits progressive human-readable lines (for status) and only later emits a final JSON payload after a marker (`Final result:`). The UI needs incremental status updates from partial stream text before that final parse.

`useObject` is better when the server streams/returns object-shaped deltas that are safe to merge into state as they arrive. In this case, intermediate chunks are not object updates; they are progress text plus a final, terminal object segment. `useCompletion` maps naturally to this endpoint:
- `completion` for real-time status parsing
- `onFinish` for one-time final parse/commit
- `stop` for cancel of status parsing

#### Why not reuse `getNlpCollections`
`getNlpCollections` handled a non-streaming, single-response flow to CMR and did not model streaming lifecycle concerns.
The new flow required:
- Incremental progress rendering
- Cancel/reset
- Final-only query commit after parse
- Navigation tied to stream completion

# Follow-up / technical debt
There is currently an unused NLP spatial drawing utility that appears to be dead code: `drawNlpSpatialData.ts`
I didn't delete it, but it was unused.  I didn't know if there's a plan to rewire it.

home search route

# Testing
in your `package.json, find the npm script called `start:optionals` and set `USE_GEOCODER=false`
set `useMockNlpStream` in your config to false.  run app as normal and validate that no changes appear in the Home search experience

1. Make a search using the input.  Validate that the input is disabled, a cancel button appears, and once the steps complete you are correctly navigated to the /search results.

2. Interrupt a search with the cancel button and validate that the search state is reset.

3. Cause the NLP search to fail (e.g. use bad AWS keys) and verify that the app gracefully handles a failure from the /nlp endpoint

### Attachments

https://github.com/user-attachments/assets/cb5cc71f-1e9c-4636-b51d-5c5eaf3f2156

# Checklist

- [ x] I have added automated tests that prove my fix is effective or that my feature works
- [x ] New and existing unit tests pass locally with my changes
- [x ] I have performed a self-review of my own code
- [x ] I have commented my code, particularly in hard-to-understand areas
- [ x] I have made corresponding changes to the documentation
- [ x] My changes generate no new warnings


static/src/js/components/NlpSearchStatus/NlpSearchStatus.tsx
```diff
@@ -0,0 +1,369 @@
...
+    .map(toProgressStep)
+    .filter(Boolean)
+}
+
+// Const OrbitLoader = ({ label }: { label: string }) => React.createElement('terra-loader', {
+//     variant: 'orbit',
+//     indeterminate: true,
+//     'aria-label': label,
+//     class: 'nlp-search-chat__step-loader'
+// })
```
stephenkilbourn: remove: terra-ui component wouldn't resize correctly

static/src/js/components/NlpSearchStatus/NlpSearchStatus.tsx
```diff
@@ -0,0 +1,369 @@
...
+
+  return (
+    <section className="nlp-search-chat" aria-live="polite">
+      <div className="nlp-search-chat__panel" role="status" aria-live="polite">
+        <div className="nlp-search-chat__step-row">
+          <Spinner
+            type="dots"
+            inline
+            size="tiny"
+            className="nlp-search-chat__step__loader"
```
stephenkilbourn:
```suggestion
            className="nlp-search-chat__step-loader"
```

static/src/js/components/NlpSearchStatus/NlpSearchStatus.tsx
```diff
@@ -0,0 +1,369 @@
+import React, {
```
mandyparson, line 1: I'm noticing that while the stream is canceled when I click 'cancel', a few seconds later I'm still redirected to /search with the results. This is probably because we're using mocks, but it would still be worth revisiting this to make sure that onNlpSearchComplete doesn't fire when 'cancel' has been clicked.

static/src/js/routes/Home/Home.tsx
```diff
@@ -206,28 +213,51 @@ export const Home: React.FC = () => {
...
-      // If nlp is enabled, use NLP search to fetch collections
-      await getNlpCollections(trimmedKeyword)
-    } else {
-      // Manually update the query in the store
-      changeQuery({
-        collection: {
-          keyword: trimmedKeyword
-        }
-      })
+      if (isNlpStreaming || !trimmedKeyword) return
```
mandyparson: If no keyword is submitted, we should redirect users to /search

static/src/js/routes/Home/Home.tsx
```diff
@@ -206,28 +213,51 @@ export const Home: React.FC = () => {
...
+
     // After collections are fetched, navigate to the Search route
     navigate(`${routes.SEARCH}${window.location.search}`)
   }
 
+  const onNlpSearchComplete = () => (
+    // Use the shared router instance so we can await real navigation.
+    // This prevents query updates from racing navigation and pushing back
+    // to Home.
+    Promise.resolve(routerHelper.router?.navigate(routes.SEARCH, {}))
```
mandyparson: can remove Promise.resolve as router.navigate does this already

static/src/js/routes/Home/__tests__/Home.test.tsx
```diff
@@ -69,6 +94,8 @@ const setup = setupTest({
 const OLD_ENV = process.env
 
 beforeEach(() => {
+  vi.clearAllMocks()
```
mandyparson: Can remove. Our tests on configured to do this automatically.

static/src/js/components/SearchForm/SearchForm.jsx
```diff
@@ -34,11 +41,29 @@ const SearchForm = () => {
...
+          ) : (
+            <SearchAutocomplete />
+          )
+        }
       </div>
       <div className="search-form__secondary">
         <div className="search-form__secondary-actions d-flex justify-content-between flex-row">
           <div className="d-flex gap-1">
+            {
+              isNlpEnabled && (
```
mandyparson: should have a test for this

static/src/js/components/NlpSearchStatus/NlpSearchStatus.tsx
```diff
@@ -0,0 +1,369 @@
...
+
+    if (typeof stop === 'function') stop()
+
+    lastStartedRequestIdRef.current = null
+    queuedStatusRef.current = ''
+    currentStatusStepRef.current = ''
+    setDisplayStatusStep('')
+    setStatusTransitionState('idle')
+    onStreamingChange(false)
+  }, [cancelRequestId, onStreamingChange, setCompletion, stop])
```
mandyparson: setCompletion isn't used in this function, can be removed

static/src/js/components/NlpSearchStatus/__tests__/NlpSearchStatus.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+  })
+
+  test('handles cancel request by stopping stream and resetting state', () => {
+    const { props } = setup({
+      overrideProps: {
+        cancelRequestId: 1
+      }
+    })
+
+    expect(mockStop).toHaveBeenCalledTimes(1)
```
mandyparson: for every CalledTimes(x) we do a CalledWith(xyz)

tests/e2e/paths/home/home.spec.js
```diff
@@ -33,6 +39,7 @@ test.describe('Home Page', () => {
   })
 
   test.describe('when performing a keyword search', () => {
+    test.skip(isNlpEnabled, 'Skipping keyword search when NLP search is enabled.')
```
mandyparson: Great idea! May want to tie this to a TODO for when NLP is fully implemented. Not sure if we have a ticket for that yet...

package.json
```diff
@@ -28,7 +28,7 @@
     "start:app": "vite dev --port 8080",
     "start": "dotenvx run node bin/start.js",
     "start:geocoder": "docker build -f geocoder.Dockerfile -t geocoder-lambda . && docker run -p 4001:8080 -e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY geocoder-lambda",
-    "start:optionals": "USE_CACHE=true SKIP_SQS=false USE_NLP_SEARCH=true USE_GEOCODER=true ORDER_DELAY_SECONDS=1 npm start",
+    "start:optionals": "USE_CACHE=true SKIP_SQS=true USE_NLP_SEARCH=true USE_GEOCODER=false ORDER_DELAY_SECONDS=1 npm start",
```
mandyparson: We'll want to keep these as they were. Just make a note in your PR to change these values locally

sharedConstants/nlpMockStream.js
```diff
@@ -0,0 +1,46 @@
...
+  query: prompt || DEFAULT_NLP_MOCK_PROMPT
+})
+
+export const createNlpMockStreamChunks = ({
+  prompt = DEFAULT_NLP_MOCK_PROMPT,
+  progressLines = DEFAULT_NLP_MOCK_PROGRESS_LINES
+} = {}) => {
+  const chunks = []
+
+  if (Array.isArray(progressLines) && progressLines.length > 0) {
```
eudoroolivares2016: nitpick; since its hardcoded I'd just not use this check

tests/e2e/paths/home/home.spec.js
```diff
@@ -17,6 +20,9 @@ import whatIsThisImageGranules from './__mocks__/what-is-this-image-granules.bod
 import whatIsThisImageGranulesHeaders from './__mocks__/what-is-this-image-granules.headers.json'
 import whatIsThisImageGraphQlBody from './__mocks__/what-is-this-image-collections.graphql.body.json'
 
+const { nlpSearch } = getApplicationConfig()
```
eudoroolivares2016: We discussed this off githhub but, since NLP search is intended to be the default search pattern on EDSC with rollout and we cannot without creating a second GH action pipeline for those tests have it switch at runtime we can just in the playwright test ensure its in NLP mode all other playwright tests are on other pages so those won't be affected at this time

static/src/js/routes/Home/Home.tsx
```diff
@@ -294,16 +330,30 @@ export const Home: React.FC = () => {
...
+                  type={isNlpStreaming ? 'button' : 'submit'}
                   className="home__hero-submit-button flex-shrink-0 btn btn-primary btn-lg focus-light"
                   bootstrapVariant="primary"
                   bootstrapSize="lg"
-                  spinner={isLoading}
+                  spinner={!isNlpStreaming && isLoading}
+                  onClick={isNlpStreaming ? onCancelNlpSearch : undefined}
                 >
-                  Search
+                  {isNlpStreaming ? 'Cancel' : 'Search'}
```
eudoroolivares2016: Cancel does seem to still send to the /search endpoint a UX solution would be to either cancel that page transition/the inflight request to the API or keep "Search" and just disable the button fields

static/src/js/routes/Home/Home.tsx
```diff
@@ -294,16 +330,30 @@ export const Home: React.FC = () => {
...
                 </Button>
               </form>
             </div>
+            {
+              isNlpEnabled && hasSubmittedNlpSearch && !!activeNlpPrompt && (
+                <div className="home__nlp-chat-wrapper">
+                  <NlpSearchStatus
```
eudoroolivares2016: Seems to leak the component on page transition or directly to /search page. Its rendering a checkbox to the left of the temporal button

static/src/js/components/SearchForm/SearchForm.jsx
stephenkilbourn: oh, yikes.  This change and the one up on line 44 were strays from earlier.  The Search form itself shouldn't have been changed.  I missed it in my revert.  🤦

static/src/js/routes/Home/Home.tsx
stephenkilbourn: the checkbox was an accidental edit to the search page that I pushed up.  Thankfully, @mandyparson  caught it as an untested change and it is removed.

sharedConstants/nlpMockStream.js
macrouch, line 1: sharedConstants and sharedUtils is for code that is shared between the static and serverless directories

static/src/js/components/NlpSearchStatus/NlpSearchStatus.tsx
```diff
@@ -0,0 +1,362 @@
...
+
+    lastStartedRequestIdRef.current = null
+    queuedStatusRef.current = ''
+    currentStatusStepRef.current = ''
+    setDisplayStatusStep('')
+    setStatusTransitionState('idle')
+    onStreamingChange(false)
+  }, [cancelRequestId, onStreamingChange, stop])
+
+  const statusStepLabel = displayStatusStep || 'Waiting for NLP status updates'
```
macrouch: Don't show the user "NLP", they don't need to know what that means

macrouch (commented):
I can't get responses to work, it looks like I am sending a request to localhost:8080/nlp which is not the api endpoint

static/src/js/components/NlpSearchStatus/NlpSearchStatus.tsx
```diff
@@ -0,0 +1,362 @@
...
+        prompt = body?.prompt || ''
+      }
+
+      const query = encodeURIComponent(prompt)
+
+      if (shouldUseMockNlpStream) {
+        return createMockNlpStreamResponse(prompt, init?.signal ?? undefined)
+      }
+
+      return fetch(`/nlp?query=${query}`, {
```
macrouch: I think this is why my requests aren't working. we should use the existing `NlpSearchRequest` to perform the query. That class will need to be modified to suite the new endpoint, but we don't use `fetch` directly, we use those request classes

static/src/js/routes/Home/Home.tsx
```diff
@@ -206,33 +217,120 @@ export const Home: React.FC = () => {
...
+  const onNlpSearchFailed = useCallback(() => {
+    resetNlpSearchUi()
+  }, [resetNlpSearchUi])
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
     event.preventDefault()
+    // Guards against any submit event that immediately follows a cancel or
+    // failure reset (e.g. a race condition causing form submission that
+    // that clears hasSubmittedNlpSearch). Without this, the empty-keyword
+    // branch below would incorrectly redirect to /search
```
mandyparson: But if someone legitimately cancels their request, retypes another one, then clicks 'search' again, it fails. They have to press Search twice to get it to go

static/src/js/components/Map/Map.tsx
```diff
@@ -1038,6 +1044,82 @@ const Map: React.FC<MapProps> = ({
...
+    })
+
+    if (previousSpatialSearchKeyRef.current === spatialSearchKey) {
+      setNlpAutoCenterPending(false)
+
+      return
+    }
+
+    const sourceExtent = spatialDrawingSource.getExtent()
+    if (!sourceExtent) return
```
mandyparson: and setNlpAutoCenterPending(false)

static/src/js/util/request/request.ts
```diff
@@ -194,6 +202,25 @@ export default class Request {
...
+  /**
+   * Makes a streamed GET request to the provided URL using fetch.
+   * This is used for APIs that progressively write their response body
+  /**
+   * @param {String} url URL to send the request to
+   * @param {Object} options Fetch request options
+   * @param {String} baseURL Optional baseUrl override
+   * @return {Promise<Response>} The fetch response promise
+   */
+  stream(url: string, options:StreamRequestOptions = {}, baseURL = this.baseUrl) {
```
mandyparson: does this need to take earthdata.env into account?

static/src/js/components/NlpSearchStatus/__tests__/NlpSearchStatus.test.tsx
```diff
@@ -0,0 +1,175 @@
+import { act } from 'react'
```
macrouch: We should import act from @testing-library/react

static/src/js/components/NlpSearchStatus/__tests__/NlpSearchStatus.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+    requestId: 0,
+    onStreamingChange: vi.fn(),
+    onNlpSearchComplete: vi.fn(),
+    onNlpSearchFailed: vi.fn()
+  }
+})
+
+describe('NlpSearchStatus component', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
```
macrouch: You shouldn't have to do a clearAllMocks in a beforeEach, there is a global config that does that for you

static/src/js/components/NlpSearchStatus/__tests__/NlpSearchStatus.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+  test('starts prompt completion when requestId and prompt are provided', async () => {
+    const { props } = setup({
+      overrideProps: {
+        activePrompt: 'glaciers in montana ',
+        requestId: 1
+      }
+    })
+
+    await waitFor(() => {
+      expect(props.onStreamingChange).toHaveBeenCalledWith(true)
```
macrouch: Any time you call `toHaveBeenCalledWith`, also do a `toHaveBeenCalledTimes(1)`, to ensure that the mock is being called the correct number of times with the correct parameters

static/src/js/components/NlpSearchStatus/NlpSearchStatus.tsx
```diff
@@ -0,0 +1,366 @@
...
+import useEdscStore from '../../zustand/useEdscStore'
+import { getEarthdataEnvironment } from '../../zustand/selectors/earthdataEnvironment'
+// @ts-expect-error: Types do not exist for this file
+import NlpSearchRequest from '../../util/request/nlpSearchRequest'
+
+import './NlpSearchStatus.scss'
+
+export const FINAL_RESULT_MARKER = 'Final result:'
+
+type NlpSearchStatusProps = {
```
macrouch, line 23: Add comments to all types
```
/** With this syntax */
```

static/src/js/components/NlpSearchStatus/NlpSearchStatus.tsx
```diff
@@ -0,0 +1,366 @@
...
+
+    return {
+      point: [pointString]
+    }
+  }
+
+  try {
+    const geometry = new WKT().readGeometry(spatialArea)
+    const geoJsonGeometry = new GeoJSON().writeGeometryObject(geometry)
+    const [minLongitude, minLatitude, maxLongitude, maxLatitude] = turfbox(geoJsonGeometry)
```
macrouch:
```suggestion
    const geometry = new WKT().readGeometry(spatialArea)
    const [minLongitude, minLatitude, maxLongitude, maxLatitude] = geometry.getExtent()
```

static/src/js/components/NlpSearchStatus/NlpSearchStatus.tsx
```diff
@@ -0,0 +1,366 @@
...
+
+      const {
+        keyword,
+        temporal,
+        spatialArea
+      } = parsedResult
+
+      const spatial = parseWktSpatial(spatialArea)
+
+      setCollectionId(null)
```
macrouch, line 233: I'm not sure what this is doing, on the landing page why would need to call this?

static/src/js/routes/Home/__tests__/Home.test.tsx
```diff
@@ -1,5 +1,5 @@
-import React from 'react'
-import { screen } from '@testing-library/react'
+import React, { act } from 'react'
```
macrouch: Import from testing library

static/src/js/routes/Home/__tests__/Home.test.tsx
```diff
@@ -14,6 +14,22 @@ import setupTest from '../../../../../../vitestConfigs/setupTest'
 
 vi.mock('../../../components/Spinner/Spinner', () => ({ default: vi.fn(() => <div />) }))
 
+type NlpSearhStatusMockProps = {
```
macrouch, line 24: Comment types

static/src/js/routes/Home/Home.tsx
```diff
@@ -206,33 +217,120 @@ export const Home: React.FC = () => {
     q: keyword
   }
 
-  const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+  const onNlpSearchComplete = React.useCallback(() => {
```
macrouch: Here you use `React.useCallback`, but below you use `useCallback`. Drop the React here to be consistent

static/src/js/util/request/__tests__/nlpSearchRequest.test.js
```diff
@@ -27,6 +28,26 @@ describe('NlpSearchRequest#search', () => {
...
 
+describe('NlpSearchRequest#stream', () => {
+  test('calls fetch with the apiHost NLP endpoint and encoded query', async () => {
+    const request = new NlpSearchRequest('prod')
+    const fetchSpy = vi.spyOn(global, 'fetch').mockResolvedValue({ ok: true })
+    const { signal } = new AbortController()
+
+    await request.stream('ice sheets in greenland', { signal })
+
+    expect(fetchSpy).toHaveBeenCalledWith(
```
macrouch, line 25: Add `toHaveBeenCalledTimes`

static/src/js/util/request/__tests__/request.test.ts
```diff
@@ -97,6 +97,23 @@ describe('Request#search', () => {
...
 
+describe('Request#stream', () => {
+  test('calls fetch with baseUrl and provided options', async () => {
+    const request = new Request(baseUrl, 'prod')
+    const fetchSpy = vi.spyOn(global, 'fetch').mockResolvedValue({ ok: true } as Response)
+    const { signal } = new AbortController()
+
+    await request.stream('/nlp?query=test', { signal })
+
+    expect(fetchSpy).toHaveBeenCalledWith('http://example.com/nlp?query=test', {
```
macrouch: Add `toHaveBeenCalledTimes`

static/src/js/util/request/request.ts
```diff
@@ -10,6 +10,14 @@ import type {
 } from '../../types/sharedTypes'
 import { routes } from '../../constants/routes'
 
+type StreamRequestCredentials = 'include' | 'omit' | 'same-origin'
+
+type StreamRequestOptions = {
+  credentials?: StreamRequestCredentials
+  headers?: Headers | Record<string, string>
+  signal?: AbortSignal
+}
```
macrouch, line 28: Comment types

static/src/js/util/request/request.ts
macrouch: This doesn't need to default to `this.baseUrl`, baseUrl is most likely the cmrHost, not apiHost. And apiHost never needs to worry about the earthdata env, because that is talking to our API, which will always be the deployed env

static/src/js/util/request/request.ts
```diff
@@ -194,6 +202,25 @@ export default class Request {
...
+   * This is used for APIs that progressively write their response body
+  /**
+   * @param {String} url URL to send the request to
+   * @param {Object} options Fetch request options
+   * @param {String} baseURL Optional baseUrl override
+   * @return {Promise<Response>} The fetch response promise
+   */
+  stream(url: string, options:StreamRequestOptions = {}, baseURL = this.baseUrl) {
+    const normalizedBaseUrl = baseURL.endsWith('/') ? baseURL.slice(0, -1) : baseURL
+    const normalisedUrl = url.startsWith('/') ? url : `/${url}`
```
macrouch:
```suggestion
    const normalizedUrl = url.startsWith('/') ? url : `/${url}`
```

tests/e2e/paths/home/home.spec.js
macrouch: Remove this, playwright shouldn't be aware of any application code

tests/e2e/paths/home/home.spec.js
```diff
@@ -202,6 +209,119 @@ test.describe('Home Page', () => {
     })
   })
 
+  test.describe('when performing an NLP search', () => {
```
macrouch, line 35: Add a test that does not type any input, to ensure we correctly load collections on the search route

static/src/js/routes/Home/Home.tsx
stephenkilbourn: this was a work-around for a bug. Deleting the entire guard because now I don't have that bug,

static/src/js/components/Map/Map.tsx
```diff
@@ -1038,6 +1044,86 @@ const Map: React.FC<MapProps> = ({
     })
   }, [spatialSearch])
 
+  // Keep auto-centering limited to NLP spatial updates so existing manual
+  // drawing/type-in flows (circle/line/polygon) retain their prior behavior.
+  // NLP currently writes spatial query state as point or bounding box.
+  useEffect(() => {
```
macrouch, line 1043: We talked about this on Slack, this will be reworked.

Just leaving a comment so we can check it off later.

static/src/js/components/NlpSearchStatus/NlpSearchStatus.tsx
stephenkilbourn: including in bigger PR feedback commit to remove old turfbox dependency

tests/e2e/paths/home/home.spec.js
stephenkilbourn: i can remove this and instead mark the tests for the new paths to always skip for now.  I like having them ready even though they aren't going to run in CI unless that nlpFlag is set.

static/src/js/components/NlpSearchStatus/__tests__/NlpSearchStatus.test.tsx
```diff
@@ -0,0 +1,359 @@
...
+        requestId: 1
+      }
+    })
+
+    await waitFor(() => {
+      expect(props.onStreamingChange).toHaveBeenCalledTimes(2)
+    })
+
+    expect(props.onStreamingChange).toHaveBeenCalledWith(true)
+    expect(props.onStreamingChange).toHaveBeenCalledWith(false)
```
macrouch: When a mock is called multiple times you can use `.toHaveBeenNthCalledWith(1, true)`, to test that the first call was made with the correct parameter

macrouch (commented):
Update static.config.json and switch `nlpSearch` to true

There are a few places where the old NLP code is still hanging out unused. 

- static/src/js/util/request/nlpSearchRequest.js - remove unused functions
- static/src/js/zustand/slices/createCollectionsSlice.ts - remove `getNlpCollections` (and anywhere that function shows up in the repo)
- static/src/js/components/Map/Map.tsx - remove `nlpSpatialSource` and `nlpSpatialLayer`
- static/src/js/util/map/drawNlpSpatialData.ts - remove file
- static/src/js/util/url/url.js - remove `nlpSearch` from `urlDefs`
- static/src/js/components/SpatialDisplay/SpatialDisplay.jsx - remove `nlpGenerated` (we used to treat nlp spatial as shapefiles, so a lot of conditionals can be removed)
- tests/e2e/paths/home/home.spec.js - unskip the nlp tests and remove the tests for the non-nlp version
- tests/e2e/map/map_nlp_spatial.spec.js - should be unskipped and updated

tests/e2e/map/map_nlp_spatial.spec.js
```diff
@@ -1,88 +1,75 @@
...
+    await expect(page).toHaveURL(/qt=2026-04-01T00%3A00%3A00\.000Z%2C2026-04-30T23%3A59%3A59\.999Z/, { timeout: 15000 })
+    await expect(page).toHaveURL(/sb\[0\]=-116\.05%2C44\.35821%2C-109\.64514%2C49\.00139/, { timeout: 15000 })
 
-    // NLP label is shown in the Spatial section
-    await expect(page.getByTestId('spatial-display_shapefile-name')).toContainText('"DC"')
+    const mapTilesPromise = page.waitForResponse(/World_Imagery\/MapServer\/tile\//)
 
-    // Check collection results count
-    await expect(page.getByText('Showing 17 of 17 matching collections')).toBeVisible()
+    await page.getByRole('button', { name: 'Browse all Earth Science Data' }).click()
```
macrouch: This button shouldn't need to be clicked after the user clicks "Search"

tests/e2e/map/map_nlp_spatial.spec.js
```diff
@@ -1,88 +1,75 @@
...
+    await expect(page).toHaveURL(/\/?q=average(?:%20|\+)temp/, { timeout: 30000 })
 
-    // Closes the modal
-    await page.getByLabel('Close').click()
+    await expect(page).toHaveURL(/qt=2026-04-01T00%3A00%3A00\.000Z%2C2026-04-30T23%3A59%3A59\.999Z/, { timeout: 15000 })
+    await expect(page).toHaveURL(/sb\[0\]=-116\.05%2C44\.35821%2C-109\.64514%2C49\.00139/, { timeout: 15000 })
 
-    // NLP label is shown in the Spatial section
-    await expect(page.getByTestId('spatial-display_shapefile-name')).toContainText('"DC"')
+    const mapTilesPromise = page.waitForResponse(/World_Imagery\/MapServer\/tile\//)
```
macrouch: The `mapTilesPromise` is used as an indicator that the map has finished drawing by waiting for a specific level of map tile to be drawn (level 10 here: `MapServer\/tile\/10`). By removing the tile level that check will not wait correctly.

tests/e2e/map/map_nlp_spatial.spec.js
```diff
@@ -1,88 +1,75 @@
...
+        delayMs: 200,
+        chunks: createNlpMockStreamChunks({
+          prompt: 'average temp in western montana last april'
+        })
+      }))
+    )
+
+    await nlpHandlers.nlp.start()
+
+    await interceptUnauthenticatedCollections({
```
macrouch, line 42: This isn't mocking the CMR collections response. `interceptUnauthenticatedCollections` alone only intercepts the base /search query, which doesn't include the results from your NLP response, so the test is performing a real network request to CMR.

tests/e2e/map/map_nlp_spatial.spec.js
macrouch, line 1: This test needs to include a screenshot like it previously did. That is telling us that the map is drawing the shape we want correctly.

We shouldn't need the long timeouts for the various assertions, we avoid doing that by waiting for indicators that the page is done loading, like waiting for the map to be done drawing with the mapTilesPromise

tests/e2e/paths/home/home.spec.js
```diff
@@ -32,173 +31,135 @@ test.describe('Home Page', () => {
...
-        additionalRequests: [{
-          body: keywordCollections,
-          headers: {
-            ...commonHeaders,
-            'cmr-hits': '1564'
-          },
-          paramCheck: (parsedQuery) => parsedQuery?.keyword === 'MODIS*'
-        }],
-        includeDefault: false
+        headers: commonHeaders
```
macrouch: You need the additionalRequests to correctly mock the collections request, this is sending a real request to CMR

tests/e2e/paths/home/home.spec.js
```diff
@@ -32,173 +31,135 @@ test.describe('Home Page', () => {
...
-        await page.getByRole('textbox', { name: 'End Date' }).fill('2020-02-01 23:59:59')
-        await page.getByRole('button', { name: 'Apply' }).click()
+      // Checks that status text progresses while NLP is running.
+      await expect(page.getByText(/Extracted keyword of "average temp"\.?/i)).toBeVisible({ timeout: 10000 })
 
-        await expect(page).toHaveURL('/search?qt=2020-01-01T00%3A00%3A00.000Z%2C2020-02-01T23%3A59%3A59.999Z')
-      })
+      await expect(page).toHaveURL(/q=average(?:%20|\+)temp/, { timeout: 15000 })
+      await expect(page).toHaveURL(/qt=2026-04-01T00%3A00%3A00\.000Z%2C2026-04-30T23%3A59%3A59\.999Z/, { timeout: 15000 })
+      await expect(page).toHaveURL(/sb\[0\]=-116\.05%2C44\.35821%2C-109\.64514%2C49\.00139/, { timeout: 15000 })
```
macrouch: This isn't checking that the page was redirected to /search like the test name says.

This URL also should have no variability. A string like it used before should work without needing regex.

We shouldn't need the long timeouts, were you seeing tests fail without waiting 15s?

tests/e2e/paths/home/home.spec.js
```diff
@@ -32,173 +31,135 @@ test.describe('Home Page', () => {
...
-            },
-            paramCheck: (parsedQuery) => parsedQuery?.keyword === 'MODIS*'
-              && parsedQuery?.temporal?.[0] === '2020-01-01T00:00:00.000Z,2020-02-01T23:59:59.999Z'
-          }],
-          includeDefault: false
-        })
-
-        await page.goto('/')
-      })
+    test('sends the expected prompt to the /nlp endpoint', async ({ page }) => {
```
macrouch: This test feels like it is already covered by the previous test. If this one isn't providing any additional benefit we don't need it

tests/e2e/paths/home/home.spec.js
```diff
@@ -32,173 +31,135 @@ test.describe('Home Page', () => {
...
-        })
+      await expect(searchInput, 'Search input should be disabled while NLP is running').toBeDisabled()
+      await expect(page.getByRole('button', {
+        name: 'Cancel',
+        exact: true
+      }), 'Cancel button should be visible while NLP is running').toBeVisible()
 
-        await page.goto('/')
-      })
+      await expect(page.locator('.nlp-search-chat__step'), 'NLP chat step should be visible while NLP is running').toBeVisible()
```
macrouch: This should be using getByText rather than a locator, we want to try to keep our tests as close to user behavior as possible, and they are unaware of specific class names.

I think that is what the next line is doing, so is this assertion necessary?

macrouch (commented):
Found a new bug, if the nlp response does not include a spatial area (like the query of "rainfall"), then EDSC errors trying to move the map to a spatial extent involving Infinity. If no spatialArea exists in the response we should ignore it

macrouch:
> Found a new bug, if the nlp response does not include a spatial area (like the query of "rainfall"), then EDSC errors trying to move the map to a spatial extent involving Infinity. If no spatialArea exists in the response we should ignore it

I see there is code to ignore it in `parseWktSpatial`, so it is a problem somewhere else. We should make sure a test somewhere covers this case

tests/e2e/paths/home/home.spec.js
stephenkilbourn: i think i was debugging and making sure it was visible before I wrote the text assertions.  removed.

tests/e2e/paths/home/home.spec.js
stephenkilbourn: ah, i see. there was a race condition on the assertions when i had /search and then the query params. i have a better solution now.

stephenkilbourn:
> > Found a new bug, if the nlp response does not include a spatial area (like the query of "rainfall"), then EDSC errors trying to move the map to a spatial extent involving Infinity. If no spatialArea exists in the response we should ignore it
> 
> I see there is code to ignore it in `parseWktSpatial`, so it is a problem somewhere else. We should make sure a test somewhere covers this case

it looks like the map fit was choking with extent values. I added a check in map.tsx and a test for the keyword only search in `NlpSearchStatus.test.tsx`

static/src/js/routes/Home/Home.scss
```diff
@@ -47,18 +116,34 @@
...
   }
 
-
   &__hero-input-wrapper {
-    max-width: 60rem;
-    align-self: center;
+    width: 100%;
+  }
+
+  &__nlp-chat-wrapper {
```
eudoroolivares2016, line 124: Had to do this patch to fix the scollbar appearing 

[fix-scrol-bar-appearing.patch](https://github.com/user-attachments/files/30608261/fix-scrol-bar-appearing.patch)

tests/e2e/paths/home/home.spec.js
macrouch: Still needs a check for `/search` in the URL

static/src/js/components/NlpSearchStatus/__tests__/NlpSearchStatus.test.tsx
```diff
@@ -329,6 +329,41 @@ describe('NlpSearchStatus component', () => {
...
+    await waitFor(() => {
+      expect(props.onStreamingChange).toHaveBeenCalledTimes(1)
+    })
+
+    expect(props.onStreamingChange).toHaveBeenCalledWith(true)
+
+    await act(async () => {
+      capturedUseCompletionOptions.onFinish?.(
+        'rainfall',
+        'Final result:\n{"keyword":"rainfall","query":"rainfall"}'
```
macrouch: This needs to include `spatial`, `spatialArea` and `temporal` all set to null, the lambda will return null for those keys

static/src/js/components/Map/Map.tsx
```diff
@@ -624,8 +630,9 @@ const Map: React.FC<MapProps> = ({
         if (sourceExtent) extent = sourceExtent
       }
 
-      // Fit the map to the extent
-      if (extent) {
+      // Fit the map only when extent values are finite.
+      // empty extent can cause Infinity bounds.
+      if (hasFiniteExtent(extent)) {
```
macrouch, line 635: This does keep the error page from showing up, and it is a good check to leave in place, but we shouldn't try to move the map if we know there was no spatial.

macrouch (approved):

static/src/js/routes/Home/Home.tsx
```diff
@@ -206,33 +216,115 @@ export const Home: React.FC = () => {
...
+
+  const isSearchInputDisabled = isNlpEnabled && isNlpStreaming
+  const shouldShowNlpStatus = isNlpEnabled && hasSubmittedNlpSearch && !!activeNlpPrompt
+
   return (
     <main className="route-wrapper route-wrapper--content-page route-wrapper--home">
       <div className="route-wrapper__content">
         <section
-          className="home__hero position-relative w-100 d-flex px-5 flex-column justify-content-center flex-shrink-0 gap-5"
+          className="home__hero position-relative w-100 d-flex px-5 flex-column flex-shrink-0"
```
eudoroolivares2016, line 327: Only thing I did notice is we should tie the box shadow only to the nlp state if we turn it off we don't want it to look like that  

Image

eudoroolivares2016 (approved):

PR #2079, opened 2026-08-12
EDSC-4678: Adds GrowthBook SDK for A/B testing

macrouch:
# Overview

### What is the feature?

Adds GrowthBook SDK for A/B testing

### What is the Solution?

Adds new Zustand slice to hold GrowthBook details
Creates a random user id and session id for users
Wraps the GrowthBook code around a deployment variable to turn on/off new code
Adds a new call to LoggerRequest for experiment event logging
Adds a new lambda to send experiment events to AWS Data Firehose

Adds new pieces

# Testing

You can set up GrowthBook to run locally with a docker compose file. Create a user, create a feature flag (probably more things you have to create but I can't remember). Name the feature flag `nlpSearch` and have it service a boolean with a default value of true or false, and you can see it enable/disable the nlpSearch on the home page.

Add the `logEvent` util to the code in `changeQuery` to see it submit to the lambda to be logged.

# Checklist

- [x] I have run `npm audit fix` and made note of any changes in this PR


serverless/src/experimentLogger/handler.js
```diff
@@ -0,0 +1,56 @@
...
+ */
+const experimentLogger = async (event) => {
+  const { body } = event
+  const { params = {} } = JSON.parse(body)
+  const { eventData } = params
+
+  const {
+    eventType,
+    userId
+  } = eventData
```
macrouch: these are underscored, fix it

stephenkilbourn (approved):

bin/deploy-bamboo.sh
```diff
@@ -38,8 +38,13 @@ config="`jq '.application.mapPointsSimplifyThreshold = $newValue' --arg newValue
 config="`jq '.application.nlpSearch = $newValue' --arg newValue $bamboo_NLP_SEARCH <<< $config`"
 config="`jq '.application.numberOfGranules = $newValue' --arg newValue "$bamboo_NUMBER_OF_GRANULES" <<< $config`"
 config="`jq '.application.placeLabelsStyleUrl = $newValue' --arg newValue $bamboo_PLACE_LABELS_STYLE_URL <<< $config`"
+config="`jq '.application.growthbookEnabled = $newValue' --arg newValue $bamboo_GROWTHBOOK_ENABLED <<< $config`"
```
mandyparson, line 41: --argjson, --arg evaluates everything to a string

static/src/js/util/metrics/experiments/logEvent.ts
```diff
@@ -0,0 +1,43 @@
...
+    const loggerRequest = new LoggerRequest()
+
+    const params = {
+      eventData: {
+        experiment_id: eventKey,
+        variation_id: eventValue,
+        event_type: eventType,
+        event_data: eventData,
+        session_id: gbSessionId,
+        user_id: gbUserId
```
mandyparson, line 33: we'll probably tweak this quite a bit, but I think it's worth adding the window.location.pathname as well

static/src/js/components/GrowthBookLoader/__tests__/GrowthBookLoader.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+    setup()
+
+    expect(localStorageGetItemSpy).toHaveBeenCalledTimes(2)
+    expect(localStorageGetItemSpy).toHaveBeenNthCalledWith(1, 'gbUserId')
+    expect(localStorageGetItemSpy).toHaveBeenNthCalledWith(2, 'gbSessionId')
+
+    expect(localStorageSetItemSpy).toHaveBeenCalledTimes(2)
+    expect(localStorageSetItemSpy).toHaveBeenNthCalledWith(1, 'gbUserId', expect.any(String))
+    expect(localStorageSetItemSpy).toHaveBeenNthCalledWith(2, 'gbSessionId', expect.any(String))
```
mandyparson, line 76: Can we make these more specific? Hard to be certain userId is actually the userId and not the sessionId

static/src/js/components/GrowthBookLoader/__tests__/GrowthBookLoader.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+      expect(localStorageGetItemSpy).toHaveBeenCalledTimes(2)
+      expect(localStorageGetItemSpy).toHaveBeenNthCalledWith(1, 'gbUserId')
+      expect(localStorageGetItemSpy).toHaveBeenNthCalledWith(2, 'gbSessionId')
+
+      expect(localStorageSetItemSpy).toHaveBeenCalledTimes(0)
+    })
+  })
+
+  test('renders the spinner then the children', async () => {
```
mandyparson: this just tests the spinner

static/src/js/components/GrowthBookLoader/GrowthBookLoader.tsx
```diff
@@ -0,0 +1,44 @@
...
+    setFeatureFlags: state.growthbook.setFeatureFlags
+  }))
+
+  const nlpSearchValue = useFeatureIsOn('nlpSearch')
+
+  // On initial load, set a user id and session id for GrowthBook tracking.
+  useEffect(() => {
+    const gbUserId = window.localStorage.getItem('gbUserId')
+    if (!gbUserId) {
+      const newGbUserId = crypto.randomUUID()
```
mandyparson, line 19: Thinking about the test, could make this `user-${crypto.randomUUID}` to help differentiate it from the sessionId

static/src/js/components/GrowthBookWrapper/GrowthBookWrapper.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  const {
+    growthbookApiHost = 'http://localhost:4100',
+    growthbookClientKey
+  } = getEnvironmentConfig()
+
+  const growthbook = new GrowthBook({
+    apiHost: growthbookApiHost,
+    clientKey: growthbookClientKey,
+    enableDevMode: true,
```
mandyparson: base off of env instead

static/src/js/components/GrowthBookWrapper/GrowthBookWrapper.tsx
```diff
@@ -0,0 +1,52 @@
...
+    // Called every time a user is put into an experiment
+    trackingCallback: (experiment, result) => {
+      // TODO call logEvent?
+      console.log('Experiment Viewed', {
+        experimentId: experiment.key,
+        variationId: result.key
+      })
+    }
+  })
+  growthbook.init({})
```
mandyparson, line 48: Can we check to see if growthbook has already been intialized first?

static/src/js/util/metrics/experiments/logEvent.ts
```diff
@@ -0,0 +1,43 @@
+// @ts-expect-error This file does not have types
+import LoggerRequest from '../../request/loggerRequest'
+// @ts-expect-error This file does not have types
+import { getApplicationConfig } from '../../../../../../sharedUtils/config'
+
+import useEdscStore from '../../../zustand/useEdscStore'
+
+const logEvent = async (eventKey: string, eventType: string, eventData: string) => {
```
mandyparson, line 8: eventData: object?

static/src/js/util/metrics/experiments/logEvent.ts
macrouch: I'm just guessing on most of these, so I'm not going to change it for this PR. On the next ticket when we learn more that would be when we need to change these

static/src/js/components/GrowthBookLoader/__tests__/GrowthBookLoader.test.tsx
macrouch: I'm not sure how to be more specific. This is ensuring that we are setting `gbUserId` and `gbSessionId`. We don't need to test that `crypto.randomUUID()` is giving us a specific string

static/src/js/components/GrowthBookLoader/GrowthBookLoader.tsx
macrouch: We don't like writing code specifically for tests if we can avoid it. Adding a prefix to this value would only add meaningless characters to every row of data we end up querying downstream. We know its a user or session because they are labeled correctly in the places they are stored

static/src/js/components/GrowthBookWrapper/GrowthBookWrapper.tsx
macrouch: There is annoying little documentation on this method, so I don't know if you can check.

But this component will only be rendered once. The only trigger it has to re-render is if the `children` prop changes, which can't happen

static/src/js/util/metrics/experiments/logEvent.ts
macrouch: My guess here is that we just need a simple data type to query downstream, so I went with string thinking that it would be which filter was changed. But in the next ticket when we know more it could change

mandyparson (approved):

bin/deploy-bamboo.sh
macrouch: I like the idea of switching this, but I wouldn't only change this field, I'd want every boolean to behave the same way.

But that also means that environment variables used in a lambda would behave differently than those in React, because they are set within the `dockerRun` function in this file.

We should put a little thought into that separately

sylvesterkaczmarek (commented):
`eventData` is client-controlled and is spread after the server-generated `event_id` and `timestamp`, so a caller can overwrite both fields before the record reaches Firehose. Spread client data first or explicitly strip protected keys, then add a regression proving `event_id`/`timestamp` cannot be supplied by the request.

PR #2083, opened 2026-08-21
EDSC-4613: Build metrics for how long it takes for the map to load

eudoroolivares2016:
# Overview

### What is the feature?

Creates three events for the map to track
Performance of the granule outlines, background, and a performance indicator for the sluggishness of the map.

This also fixes the main issue by cleaning up accumulating listener objects from the interface and some minor improvements by just bailing out before extra processing is done

### What is the Solution?

1. For the granule performance metrics we can just listen for the postrender events that openlayers producers collect them and then send that to the dataLayer GTM event middleware. Within these events we are also bucketing users based on granule count. This is so that we can get an idea of how many granules users are observing at a time as we may decide to cap that amount.

2. For the "FPS" timing: we create a timer for a configurable length proposing 3 minutes and use that in conjunction to a listener on MoveEnd events. This is so that we don't collect stats for when people aren't moving the map. The idea is to understand how fast/slow the map moves for users. When a move-end occurs and the time has passed we divide up that metric into percentiles so its easier for us to use this metrics on the analytics side.

3. The main performance improvement in this PR is the useEffect at the end this is just cleaning up the granule outline listeners which were accumulating and causing the slowdown. There are a few other improvements based on console.counts appearing e.g. calls that don't need to happen

Map.jsx and data layer metrics

# Testing

### Reproduction steps

- **Environment for testing:*Prod*
- **Collection to test with:*
C1701805625-GES_DISC - granule with many data layers*
C2259345545-GES_DISC - main issue one; complex granule vertices
C3104974973-SEDAC - simple use-case; control 

1. Navigate to the granules menu
4. Use https://tagmanager.google.com/#/home preview mode or add console logs to see the data layer events firing
5. The two should fire when granule drawing events render. The latter requires moving around/zooming the map for a few minutes to see the event then fire

### Attachments

C2259345545-GES_DISC - without the use-effect cleanup
image

C2259345545-GES_DISC - after the use-effect cleanup
image

# Checklist

- [ ] I have run `npm audit fix` and made note of any changes in this PR


static/src/js/components/Map/Map.tsx
```diff
@@ -1076,16 +1274,26 @@ const Map: React.FC<MapProps> = ({
     projectionCode
   ])
 
-  // Draw the granule outlines
-  granuleOutlinesLayer.on(RenderEventType.POSTRENDER as LayerRenderEventTypes, (event) => {
-    const ctx = event.context as CanvasRenderingContext2D
+  // Clear listeners for the granule outlines layer so they do not accumulate
+  useEffect(() => {
```
eudoroolivares2016: remove this fix will put it into the next PR

static/src/js/util/metrics/metricsMap.ts
```diff
@@ -1,15 +1,76 @@
...
+  render: MapRenderStats,
+  windowDurationMs: number
+}
+
+export const metricsMapFramePerformance = (event: MapPerformanceEvent) => {
+  const x = {
+    event: 'MapFramePerformance',
+    ...event
+  }
+  console.log('the map MapFramePerformance event', x)
```
eudoroolivares2016: needs remove

static/src/js/util/metrics/metricsMap.ts
```diff
@@ -1,15 +1,66 @@
...
+*/
+
+export const metricsMapRenderPerformance = (
+  granuleCount: number,
+  totalRenderMs: number,
+  mapEventAction: string,
+  collectionId: string
+) => {
+  dataLayer.push({
+    event: 'map',
```
eudoroolivares2016, line 53: We want:
zoomLevel, center

static/src/js/util/metrics/metricsMap.ts
```diff
@@ -1,15 +1,66 @@
...
+  verySlowFrames: number
+}
+
+export interface MapPerformanceEvent {
+  collectionId: string,
+  render: MapRenderStats,
+  windowDurationMs: number
+}
+
+export const metricsMapFramePerformance = (event: MapPerformanceEvent) => {
```
eudoroolivares2016, line 94: number of granules

static/src/js/util/metrics/metricsMap.ts
```diff
@@ -1,15 +1,66 @@
...
+  dataLayer.push({
+    event: 'MapFramePerformance',
```
eudoroolivares2016: make the casing consistent

static/src/js/util/metrics/metricsMap.ts
eudoroolivares2016: call imagery rather than "granule backgrounds"

static/src/js/util/metrics/helpers.ts
```diff
@@ -160,3 +160,48 @@ export const computeFacets = () => {
...
+ * cardinality low while still conveying rough scale.
+ *
+ * @param count - The number of granules to bucket.
+ * @returns A string representing the bucket range, e.g. '0', '1-20', '21-40',
+ *          '41-60', '61-100', or '100+'.
+ */
+export const computeBucketGranuleCount = (count: number): string => {
+  if (count === 0) return '0'
+  if (count <= 20) return '1-20'
+  // Do the first few pages as bucket
```
eudoroolivares2016: remove

static/src/js/components/Map/Map.tsx
```diff
@@ -99,9 +109,28 @@ import {
 } from '../../types/sharedTypes'
 import { MapView, ShapefileSlice } from '../../zustand/types'
 
+interface MapPerformanceWindow {
+  windowStart: number
+  frames: number
```
eudoroolivares2016, line 116: Put a comment on some of these

static/src/js/components/Map/Map.tsx
```diff
@@ -99,9 +109,28 @@ import {
...
+  slowFrames: 0,
+  verySlowFrames: 0,
+  renderTimes: [],
+  maxRenderTimeMs: 0
+})
+
 let previousGranulesKey: string
 let previousProjectionCode: ProjectionCode
 let layersAdded = false
+const PERFORMANCE_WINDOW_MS = 3000
```
eudoroolivares2016: This should be a bamboo var

static/src/js/components/Map/Map.tsx
```diff
@@ -177,6 +215,63 @@ const overlayLayers: Record<string, TileLayer | VectorTileLayer | null> = {
...
+    latitude: number
+    longitude: number
+  }
+  /** Map zoom level at the time this render was triggered */
+  zoomLevel: number
+}
+
+// Times a single prerender → postrender pass for a layer using .once(),
+// so it only fires for the very next render triggered after this is called.
+export const timeLayerRenderOnce = ({
```
macrouch: This function feels pretty self contained, except for maybe the `performance` calls. If so we should move it to a util function to keep the file cleaner

static/src/js/components/Map/Map.tsx
```diff
@@ -257,6 +352,47 @@ const hasFiniteExtent = (extent: import('ol/extent').Extent | null | undefined):
   return extent.every((coordinate) => Number.isFinite(coordinate))
 }
 
+const flushMapPerformanceMetrics = (
```
macrouch: Same comment as above, could this be a util function?

static/src/js/components/Map/Map.tsx
```diff
@@ -415,6 +555,33 @@ const Map: React.FC<MapProps> = ({
...
   const mapElRef = useRef<HTMLDivElement>(null)
+  const lastFrameRef = useRef<number>(performance.now())
+  const frameTimesRef = useRef<number[]>([])
+  const isTrackingFrameRef = useRef(false)
+  const performanceWindowRef = useRef<MapPerformanceWindow>(createEmptyPerformanceWindow())
+
+  // Mirror the latest focused collection(s) into a ref so the long-lived
+  // map event listeners always read current values without needing to be
+  // re-registered. On the project page there can be multiple focused
+  // collections; everywhere else there's exactly one.
```
macrouch: there isn't exactly one, there could be zero

static/src/js/components/Map/Map.tsx
```diff
@@ -984,51 +1218,93 @@ const Map: React.FC<MapProps> = ({
 
   // When the granules change, draw the granule backgrounds
   useEffect(() => {
+    if (granules && granules.length > 0) {
     // If the granules haven't changed and the projection hasn't changed, don't redraw the granule backgrounds
     // Redraw the granule backgrounds if the product layer from the gibs tag has changed
-    if (granulesKey === previousGranulesKey && projectionCode === previousProjectionCode) return
+      if (granulesKey === previousGranulesKey
+        && projectionCode === previousProjectionCode) return undefined
```
macrouch, line 1156: Keep an empty line after return statements like this one

static/src/js/components/Map/Map.tsx
```diff
@@ -984,51 +1218,93 @@ const Map: React.FC<MapProps> = ({
...
+      // Clear the existing granule backgrounds
+      granuleBackgroundsSource.clear()
+      // Clear any existing granule highlights
+      unhighlightGranule(granuleHighlightsSource)
+
+      // Clear any existing focused granules
+      clearFocusedGranuleSource(mapRef.current as OlMap)
+
+      // Clear the granule imagery layers
+      granuleImageryLayerGroup.getLayers().clear()
```
macrouch: These four calls are duplicated after this if statement below. Move them above the if, or make a util function, something so you don't duplicate the code

static/src/js/components/Map/Map.tsx
```diff
@@ -984,51 +1218,93 @@ const Map: React.FC<MapProps> = ({
...
+          onExcludeGranule,
+          setGranuleId,
+          shouldMoveMap: false,
+          timesIconSvg
+        })
+      }
+
+      const activeCollectionIds = isProjectPage ? projectCollections.allIds : [focusedCollectionId]
+
+      const timingKeys = [
```
macrouch, line 1194: Could use a comment

static/src/js/util/map/drawGranuleBackgroundsAndImagery.ts
```diff
@@ -350,6 +351,7 @@ const drawGranuleBackgroundsAndImagery = ({
     })
   })
 
+  if (granuleImageryLayers.length === 0) return
```
macrouch, line 354: Add empty line below return statement

static/src/js/components/Map/Map.tsx
```diff
@@ -291,6 +427,9 @@ interface MapProps {
   isFocusedCollectionPage: boolean
   /** Flag to show if this is a project page */
   isProjectPage: boolean
+  projectCollections: {
```
macrouch, line 319: Missing comments

sylvesterkaczmarek (commented):
The 3-second frame window has no timer: `handleMoveEndPerf()` only flushes when that particular interaction ends after `PERFORMANCE_WINDOW_MS`. A normal sub-3s pan/zoom therefore emits no frame metric unless the user interacts again later, at which point old samples and window duration are merged with the later context. Schedule a deadline flush (or flush per interaction) and add a single short-interaction regression.

sylvesterkaczmarek (changes requested):
Rechecked current `1de2b36`. The short-interaction metric-loss issue is still present: `handleMoveEndPerf()` only checks elapsed time when a later `moveend` occurs, so a normal interaction that ends before `MAP_PERFORMANCE_WINDOW_MS` can sit indefinitely and be merged into a later interaction. Please schedule a deadline flush (or flush each completed interaction) and cover the single short-interaction case.

static/src/js/util/map/mapPerformanceMetrics.ts
```diff
@@ -0,0 +1,145 @@
...
+        center,
+        zoomLevel
+      })
+    })
+  })
+
+  return [preKey, postKey]
+}
+
+export const flushMapPerformanceMetrics = (
```
mandyparson, line 106: comment

static/src/js/routes/Search/Search.jsx
```diff
@@ -235,7 +235,7 @@ export const Search = () => {
                                           placement="top"
                                           overlay={
                                             (tooltipProps) => renderTooltip({
-                                              children: 'Include collections labeled as planned, deprecated, preprint, in review, superseded, or not provided in results',
+                                              children: 'Include collections labeled as planned, deprecated, preprint, in review, or not provided in results',
```
mandyparson, line 238: why did you remove superseded again?

mandyparson (approved):

static/src/js/routes/Search/Search.jsx
eudoroolivares2016: Its not actually what the CMR query filters on

sylvesterkaczmarek (changes requested):
Re-reviewed current be1b54c. My earlier blocker is still present: handleMoveEndPerf only flushes when a later moveend occurs after MAP_PERFORMANCE_WINDOW_MS, with no deadline timer. A single short interaction can therefore remain unreported indefinitely and be merged into later context. Please add a deadline/per-interaction flush and a single-short-interaction regression.

eudoroolivares2016:
@sylvesterkaczmarek 
Thanks for the re-check. I don't think the metric drops data the way this describes — the window isn't anchored to the interaction, it's anchored to map mount and only re-anchored on a flush.

`performanceWindowRef` is initialized with `windowStart = performance.now()` when the map is created, and the only place that gets reset is `createEmptyPerformanceWindow()` after a successful flush. Nothing resets it on `movestart`. So the elapsed check on the first `moveend` is measuring time since the map loaded, not time since the gesture began. Any user who takes more than `MAP_PERFORMANCE_WINDOW_MS` (3000, from static.config.json) after map load to touch the map — essentially all of them — gets that first interaction flushed at its own `moveend`.

For a short interaction to sit unflushed, a user would have to pan within the first 3 seconds of the map mounting and then never produce another `moveend` for the rest of the session. `moveend` also fires on programmatic view changes (spatial selection, zoom-to-collection, the MOVEMAP emitter), not just gestures, so that's a thin slice — and it isn't the behavior this metric is meant to capture.

On flushing per interaction: accumulating across gestures is deliberate. This is a rolling-window sluggishness aggregate, not a per-interaction measurement. A 300ms pan is roughly 20 frames, and p95/p99 over 20 samples is noise. Per-interaction flushing would also multiply our GTM dataLayer volume by the number of gestures in a session. The comment on `MapPerformanceWindow.frames` already calls out the cumulative-across-gestures behavior as intended.

stephenkilbourn (approved):

sylvesterkaczmarek (changes requested):
I agree with your clarification that the frame metric is intentionally a rolling window rather than a per-interaction metric, so I am dropping my earlier deadline-flush objection. I did spot a separate cleanup bug on the current head: the effect cleanup calls `map.on` for `postrender`, `movestart`, and `moveend` instead of `map.un`, so teardown adds the three performance listeners again rather than removing them. Please switch those cleanup calls to `map.un` and, if practical, cover listener cleanup/remount behavior.

eudoroolivares2016:
@sylvesterkaczmarek Not sure I'm tracking there the useEffect cleanup, line, 797, the performance listeners are already removed with map.un, not map.on

sylvesterkaczmarek (approved):
Rechecked current ef6ef88c. You're right: the cleanup now uses map.un for postrender, movestart and moveend, so the performance listeners are removed rather than re-added on teardown. My cleanup blocker is resolved. No blocker from me.

static/src/js/util/metrics/metricsMap.ts
eudoroolivares2016: So since the rest of the code is calling this backgrounds and the layer is also called that I don't want to arbitrarily make this different

PR #2086, opened 2026-09-08
EDSC-4679: Add preferred home search method to user settings

stephenkilbourn:
# Overview

### What is the feature?

Adds a user preference for the default home search mode with **AI Enhanced** and **Traditional Search** options to allow authenticated users to set their default search method from the Home route.

### What is the Solution?

Adds `homeSearchMode` to site preferences with `AI Enhanced `and `Traditional Search` options.
Hides the preference from the user preference page/schema when NLP search feature toggle is disabled.
Uses the saved preference when initializing the home search mode if local storage has no preference.
If the user has not recently changed the toggle to update local storage, then we check if the authenticated user has a saved preference. We then fall to the GrowthBook feature flag when no saved preference or local storage is available.

### What areas of the application does this impact?

home search

# Testing

### Reproduction steps

1. visit the /home route while unauthenticated.  
2. Validate that the home search method uses the method stored in local storage based on your last selection of the AI Enhanced/Traditional Search toggle.
3. Log In and select a preferred "Home Search Mode".
4. Revisit the /home route and validate that the home search mode matches the saved "Home Search Mode" in your user preferences.

### Attachments
Screenshot 2026-09-08 at 11 52 21 AM

Please include relevant screenshots or files that would be helpful in reviewing and verifying this change.
# Checklist

- [x] I have added automated tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have run `npm audit fix` and made note of any changes in this PR
- [x] My changes generate no new warnings

PR #2092, opened 2026-09-16
ESDC-4679: support default state for logged in user homeSearchMode

stephenkilbourn:
# Overview

### What is the feature?

This PR updates the Home Search Mode preference to support a new `default` value. `homeSearchMode` now allows `default`, `nlp`, and `traditional`, with `default` as the schema default.

This adds a session storage variable: `dontShowNlpPopup` which will help determine if a user has previously toggled the Home Search Mode toggle on the home page.  If the user has not previously seen the toast, then they will be shown a message that they can set their preference on the User Preferences page if they change the search mode.

The Home page resolves its search mode using the Zustand user preference, local storage, and the GrowthBook NLP feature flag. Explicit user preferences from the zustand state take precedence. When the preference is `default`, Home checks local storage before falling back to GrowthBook. The Home toggle updates local storage and the current UI state only; saved user preferences are updated through the User Preferences page.  

### What is the solution?

The changes from #2086  to `static/src/js/components/Preferences/PreferencesRadioField.jsx` were reverted, and the schemas were updated to allow `default` as a valid value.

The user flow now supports a `default` state for both unauthenticated users and users who have not explicitly selected a search mode.

#### User Scenarios

Search mode resolution is the same for authenticated and unauthenticated users:

1. User user preference is `nlp` or `traditional`
   * The Zustand user preference is always used.
   * Local storage and GrowthBook do not override it.

2. User preference is `default`
   * If local storage contains `nlp` or `traditional`, Home uses that value.
   * If no valid local-storage value exists:
     * GrowthBook `nlpSearch: true` uses AI Enhanced Search.
     * GrowthBook `nlpSearch: false` uses Traditional Search.
     * An unavailable GrowthBook value defaults to AI Enhanced Search.

3. NLP is disabled in the application configuration
   * Home uses Traditional Search.
   * The Home search-mode toggle is hidden.
   * The `homeSearchMode` preference is hidden from User Preferences.

4. The Home toggle is changed
   * The selected mode updates the current Home UI immediately.
   * The selected mode is written to local storage.
   * The user’s saved profile preference is not changed.
   * Saved profile preferences are updated only through the User Preferences page.

/home search

# Testing

### Reproduction steps

1. Step 1
2. Step 2...

### Attachments
Screenshot 2026-09-18 at 2 38 09 PM

# Checklist

- [ ] I have run `npm audit fix` and made note of any changes in this PR


static/src/js/routes/Home/Home.tsx
```diff
@@ -83,23 +83,32 @@ type PreferredHomeSearchMode = HomeSearchMode | null
...
   if (!isNlpEnabled) return traditionalSearchMode
 
+  if (hasSavedPreferences) {
+    if (savedSearchMode === defaultSearchMode) {
+      return isNlpFeatureFlagEnabled === false ? traditionalSearchMode : nlpSearchMode
+    }
+
+    if (savedSearchMode) return savedSearchMode
+  }
+
```
macrouch: createUserSlice sets the preferences to `default` by default, so every user regardless of being logged in or not has preferences you can use. This file doesn't need to do additional checks to determine if the user is logged in or if they have saved preferences

static/src/js/zustand/slices/createUserSlice.ts
```diff
@@ -25,6 +25,7 @@ export const initialSitePreferences: PreferencesData = {
   granuleListView: 'default',
   collectionSort: 'default',
   granuleSort: 'default',
+  homeSearchMode: 'default',
```
macrouch, line 28: Take a look at `setSitePreferences`, currently it is removing default preferences if they don't exist in the user's download preferences. If you make sure to keep the defaults set here it will make it easier throughout the code to use preferences

static/src/js/routes/Home/Home.tsx
```diff
@@ -204,26 +216,26 @@ export const Home: React.FC = () => {
...
   useEffect(() => {
-    // Local storage take priority over saved user preference, so don't
-    // let homeSearchMode in preferences override existing local storage value
-    const storedSearchMode = localStorage.getItem(localStorageKeys.homeSearchMode)
-    if (storedSearchMode === nlpSearchMode || storedSearchMode === traditionalSearchMode) return
-
-    if (isNlpEnabled && homeSearchMode) setPreferredHomeSearchMode(homeSearchMode)
-  }, [homeSearchMode, isNlpEnabled])
+    if (isNlpEnabled) {
+      setPreferredHomeSearchMode(getPreferredHomeSearchMode(
```
macrouch: Can you explain this useEffect?

macrouch (changes requested):

static/src/js/routes/Home/Home.tsx
stephenkilbourn: I'm guarding running it on the environment feature flag, but that runs the `setPreferredHomeSearchMode` based on changes to having saved preferences and having local storage settings using getPreferredHomeSearchMode.
It then stores that result in preferredHomeSearchMode, which controls whether the UI shows the NLP or traditional search.  It may need changed based on above comments.

macrouch: Can you add a comment to the new version explain what it is doing

macrouch:
The PR description says local storage is being updated but that is not happening. Is the goal now to not remember the user's selection at all?

stephenkilbourn:
> The PR description says local storage is being updated but that is not happening. Is the goal now to not remember the user's selection at all?

@macrouch -  I took that out in the most recent commit, but accidentally only updated the scenarios and not the 2nd paragraph. I realized that I was setting local storage, but not actually using it. Would it make more sense to return to setting local sotrage, but only use that if it zustand state is "default"?

if zustand === nlp or traditional, always use that
if zustand === default, check for local state
--- if local exist, use that
--- if local does not exist, use growthbook

macrouch (approved):

eudoroolivares2016 (approved):

